PR #4, opened 2018-01-29
Add CMake FortranCInterface to detect Fortran name mangling

junkudo:
This commit modifies the CMake build scripts to use the FortranCInterface module to figure out the Fortran name mangling pattern during configuration.  This commit modifies the corresponding source code to use the CMake generated macros.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,11 @@
-cmake_minimum_required (VERSION 2.8.11)
+cmake_minimum_required (VERSION 3.0.2)
 project (hippopde)
+# enable Fortran for Fortran name mangling
+enable_language(Fortran)
+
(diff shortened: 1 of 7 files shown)


cnpetra:
thanks! looks good, minimal changes. Will test and merge it once I get back from DC.

cnpetra:
@junkudo : is there any reason to require CMake 3.0 ? It is not present on quartz for example. I've changed it to 2.8. Other than that, everything worked fine and it is merged in the master now. Thanks again!

junkudo:
I was under the assumption that FortranCInterface was only available on cmake v 3.0.2: https://cmake.org/cmake/help/v3.0/module/FortranCInterface.html.  Apparently they back-ported it to 2.8?  No idea.  Glad it worked.

Issue #16, opened 2020-02-28
HiOp GPU-enabled build fails


pelesh:
My HiOp build from `dev/NewtonMDS` branch fails at compile stage with message:
```
make[2]: *** No rule to make target `/usr/lib64/libopenblas.so -lmagma -L/share/apps/cuda/9.2/lib64 -lculibos -lcublas -lcublasLt -lnvblas -lcusparse -lcudart -lcudadevrt', needed by `src/LinAlg/test_hiopLinAlgComplex.exe'.  Stop.
```
It seems that there is some mess-up with CMake paths. I used following configuration for build:
```
CC=mpicc CXX=mpicxx FC=mpif90 cmake              \
-DHIOP_USE_MPI=1                                 \
-DHIOP_USE_GPU=1                                 \
-DHIOP_MAGMA_DIR="/.../exasgd/newell/magma" \
-DCMAKE_INSTALL_PREFIX=$HIOP_DIR                 \
../hiop
```
I used cmake 3.13.4, gcc 7.4.0, openmpi 3.1.3, cuda 9.2, OpenBLAS 0.3.3, and LAPACK 3.4.2 for the build. Configure part works fine, but something goes wrong with complex linear algebra compilation.

Verbose output gives me this:
```
make -f src/LinAlg/CMakeFiles/test_hiopLinAlgComplex.exe.dir/build.make src/LinAlg/CMakeFiles/test_hiopLinAlgComplex.exe.dir/depend
make[2]: Entering directory `/.../exasgd/src/hiop/build_newell'
cd /.../exasgd/src/hiop/build_newell && /.../cmake/3.13.4/bin/cmake -E cmake_depends "Unix Makefiles" /.../exasgd/src/hiop/hiop /.../exasgd/src/hiop/hiop/src/LinAlg /.../exasgd/src/hiop/build_newell /.../exasgd/src/hiop/build_newell/src/LinAlg /.../exasgd/src/hiop/build_newell/src/LinAlg/CMakeFiles/test_hiopLinAlgComplex.exe.dir/DependInfo.cmake --color=
make[2]: Leaving directory `/.../exasgd/src/hiop/build_newell'
make -f src/LinAlg/CMakeFiles/test_hiopLinAlgComplex.exe.dir/build.make src/LinAlg/CMakeFiles/test_hiopLinAlgComplex.exe.dir/build
make[2]: Entering directory `/.../exasgd/src/hiop/build_newell'
[...]

PR #17, opened 2020-04-03
Quick fix in dev/NewtonMDS cmake

pelesh:
- Removed redundant `"` in library list that caused linking drivers to HiOp library to fail. This issue may be dependent on CMake version.
- Set minimum CUDA version to 10.1, since HiOp depends on cublasLt, which is available only as of that version.

Fixes issue #16


cnpetra:
not sure what's happening: while I see two commits, there are no files changed ("Files changed 0")

pelesh:
It seems I pushed changes to a wrong remote, sorry. It is in there now.

pelesh:
@cnpetra, can you confirm if this fix works for you or if it breaks things apart? 

By the way, this is a patchwork solution, just to make things work.

cnpetra:
works fine, merged

PR #21, opened 2020-04-16
Mixed dense-sparse GPU linear algebra and Kron reduction

cnpetra:
This is a (super)major PR containing
 - new mixed dense-sparse interface (MDS), MDS linear algebra layer, and partial support for GPU computations (via Magma and CUDA)
 - support for Kron reduction of ACOPF problems
 - support for complex matrices and factorizations (via SuiteSparse)
- implementation of the Newton IPM filter linear search algorithm (inertia correction not supported at this point)
 - new functionality in the (double) sparse matrices
 - various bug fixing


cnpetra:
@pelesh

pelesh:
This is impressive work @cnpetra!

I think CMake fix from #17 applies here, as well. It might be easier to implement it by hand than to try to merge it from `dev/NewtonMDS`. It is just removing `"` at two places and setting minimum CUDA version to 10.1.

PR #26, opened 2020-05-07
Build shared and static library together + added C interface

michel2323:
### Build shared and static library

For calling HiOp I need the shared library in Julia.

This creates both the shared and static library by compiling the object files once. The object files are compiled position independently as that is needed for the shared library. This is in theory a bit slower, but I doubt we see any difference in practice as most of the computation is in the underlying libraries.

Both shared and static libraries have the same name (hiop[.so/.a]). This would only be a problem under Windows.

### C interface

Since I can only run one with other I also added the C interface here. I tested it by using it from Julia. It is still very bare bone with only one set of default options that cannot be changed at the moement.


michel2323:
@cnpetr This is ready to be merged.

michel2323:
@ashermancinelli @cnpetra I tried to pull in the cmake merge in master. Now I get
```
-- Found LAPACK libraries: /usr/lib/x86_64-linux-gnu/liblapack.so;/usr/lib/x86_64-linux-gnu/libf77blas.so;/usr/lib/x86_64-linux-gnu/libatlas.so
CMake Error at src/Optimization/CMakeLists.txt:2 (target_link_libraries):
  Object library target "hiopOptimization" may not link to anything.
```

Other the C interface test is running.

michel2323:
A more recent version of cmake works. Ubuntu 18.04 comes with 3.10. 3,17 seems to work. It might to good to reflect that in the README.md.

ashermancinelli:
Did you try wth 3.8? We bumped the minimum version to 3.8 and I believe that version of cmake should work.

michel2323:
I have 3.10,12,14,15,17 available with 3.10 being the default Ubuntu one. 3.12 works too. Maybe there is also just something wrong with default Ubuntu one.

ashermancinelli:
Do you know the lowest version of cmake that does not break the build? We could just increase the required cmake version to that version if it's not too unreasonable. Feel free to look through the recent PR, as nothing should really have changed except for caching some paths and more consistent library discovery and I would be happy to try to fix whatever it is that broke your build.

CMakeLists.txt
```diff
@@ -177,17 +179,34 @@ if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
   set(CMAKE_INSTALL_PREFIX "${CMAKE_SOURCE_DIR}/_dist-${CMAKE_BUILD_TYPE}")
 endif(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
 
+# Shared library should be position independent
+set_property(TARGET hiopOptimization PROPERTY POSITION_INDEPENDENT_CODE 1)
+set_property(TARGET hiopLinAlg PROPERTY POSITION_INDEPENDENT_CODE 1)
+set_property(TARGET hiopUtils PROPERTY POSITION_INDEPENDENT_CODE 1)
+set_property(TARGET chiopInterface PROPERTY POSITION_INDEPENDENT_CODE 1)
```
ashermancinelli, line 186: You may `set(CMAKE_POSITION_INDEPENDENT_CODE ON)` to enable -fPIC project-wide

pelesh:
The second error will be triggered by CMake < 3.12, but that can be fixed by putting proper guards around the policy setting.

I am still puzzled what is causing the first error related to `hiopOptimization` linking.

CMakeLists.txt
michel2323: Fixed 9684cfcf569 .

michel2323:
@cnpetra Can we merge this for now and open a CMake issue independent of the changes here?

cnpetra:
will take a look and test these days. 

I am thinking to introduce an option for shared library in the build system.  By default only the static build. So that we avoid -fPIC'ing the whole thing.

ashermancinelli:
@cnpetra Would you rather merge this and deal with options for shared/static in another PR or would you like to have the build options in this branch before merging?

cnpetra:
ok, will keep it in #31

ashermancinelli:
If @michel2323 would like, he can rebase off of my PR and then you @cnpetra can merge mine first and then his, which leaving you with no merge conflicts between the two and each PR is responsible for one significant change.

cnpetra:
@michel2323 when cmake-ing with `-DHIOP_USE_MPI=ON`, it does not compile/make, I get MPI-related compilation errors. Funny thing is that I get different errors on Mac and on Ubuntu.

cnpetra:
with `-DHIOP_USE_MPI=OFF`, compiles and works. But valgrind reports issues of all sorts on `nlpMDS_cex4.exe`.

cnpetra:
btw, this is what I use when I merge in the `master` : [README_developers.md](https://github.com/LLNL/hiop/blob/master/README_developers.md)

michel2323:
@cnpetra I had also an error now with MPI switched to on. After I deleted the file `hiop_defs.hpp` it worked and I cannot even reproduce it anymore. I can switch to MPI on and off.

Anyhow, I'm on the valgrind warnings. Can't find one uninitialized variable for now.

ashermancinelli:
yes, the cmake refactor now configures `hiop_defs.hpp` as `hiop_defs.hpp.in` and any leftover hiop_defs may interfere.

michel2323:
@cnpetra Valgrind is purring with no errors anymore. It runs with MPI on and off. Can you take a look again?

cnpetra:
```
valgrind --leak-check=full ./src/Drivers/nlpMDS_cex4.exe 
```

==4974== LEAK SUMMARY:
==4974==    definitely lost: 428,864 bytes in 9 blocks

src/Interface/chiopInterface.cpp
```diff
@@ -0,0 +1,36 @@
+#include "chiopInterface.hpp"
+extern "C" {
+
+using namespace hiop;
+
+// These are default options for the C interface for now. Setting options from C will be added in the future.
+int hiop_createProblem(cHiopProblem *prob) {
+  cppUserProblem * cppproblem = new cppUserProblem(prob);
```
cnpetra, line 8: please also delete this; it leaks.

I did by adding a `void *` member in the `C` struct to hold this, and then `static_cast` and `delete` in `hiop_destroyProblem` (I couldn't push because I don't have rights to your fork, maybe next time you should branch of this repo?)

michel2323:
==43908== 
==43908== HEAP SUMMARY:
==43908==     in use at exit: 0 bytes in 0 blocks
==43908==   total heap usage: 13,833 allocs, 13,833 frees, 25,723,974 bytes allocated
==43908== 
==43908== All heap blocks were freed -- no leaks are possible
==43908== 
==43908== For counts of detected and suppressed errors, rerun with: -v
==43908== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)

@cnpetra I was working on it. Next time I branch off here in this repository. I didn't have the rights back then.

cnpetra:
thanks.

Issue #25, opened 2020-05-06
`copyFrom` and similar methods are not part of the abstract hiopMatrix interface


ashermancinelli:
Heavy use is made of `copyFrom` and similar methods, yet the abstract interface does not mandate that a hiopMatrix implementation has this method. In contrast, hiopVector does have `copyTo` and `copyFrom` methods in the abstract interface. I believe this could lead to implementation-specific code which will cause problems when we attempt to migrate to other implementations for hiopMatrix.

cnpetra:
I agree. will revisit hiopMatrix abstract class

another related todo would be to reconcile overloads of `addSubDiagonal`. in particular
```
void addSubDiagonal(const double& alpha, long long start_on_dest_diag, const hiopVector& d);
```
is redundant

ashermancinelli:
If you don't get to this by next week I might take a stab at it, I know you've got a lot on your plate at the moment. Thanks!

cnpetra:
on a closer inspection, it appears that `copyFrom` is specific to hiopDenseMatrix. I prefer to not push it up in the inheritance hierarchy to avoid confusion later on and since its functionality can be achieved via `new_copy`

Issue #23, opened 2020-04-29
axdzpy_w_pattern is implemented, but not part of hiopVector abstract interface


pelesh:
Method `axdzpy_w_pattern` is implemented in class `hiopVectorPar`, but there is no abstract interface to it in (almost) pure virtual class `hiopVector`. Is this method needed?

cnpetra:
apparently it is used: https://github.com/LLNL/hiop/search?q=axdzpy_w_pattern&unscoped_q=axdzpy_w_pattern

I would add it  as pure vritual in the base class. I forgot to.

PR #29, opened 2020-05-11
Add copy methods to hiopMatrix interface

ashermancinelli:
As per #25 adds methods to interface. What other methods may we need to add to the interface? How can we reconcile the `addSubDiagonal` implementations?

### DIFF
--- a/src/LinAlg/hiopMatrix.hpp
+++ b/src/LinAlg/hiopMatrix.hpp
@@ -84,6 +84,16 @@ public:
   virtual hiopMatrix* alloc_clone() const=0;
   virtual hiopMatrix* new_copy() const=0;
 
+  virtual void copyFrom(const hiopMatrix& matrix) = 0;
+  virtual void copyFrom(const double* buffer) = 0;
+
+  /* copies 'src' into this as a block starting at (i_block_start,j_block_start) */
+  void copyBlockFromMatrix(const long i_block_start, const long j_block_start, const hiopMatrix& src) = 0;
+
+  /* overwrites 'this' with 'src''s block that starts at (i_src_block_start,j_src_block_start) 
+   * and has dimensions of 'this' */
+  void copyFromMatrixBlock(const hiopMatrix& src, const int i_src_block_start, const int j_src_block_start) = 0;
+
   virtual void setToZero()=0;
   virtual void setToConstant(double c)=0;
 


cnpetra:
I am working on addSubDiagonal in a different branch, let's leave it out for now.

src/LinAlg/hiopMatrix.hpp
```diff
@@ -84,6 +84,16 @@ class hiopMatrix
   virtual hiopMatrix* alloc_clone() const=0;
   virtual hiopMatrix* new_copy() const=0;
 
+  virtual void copyFrom(const hiopMatrix& matrix) = 0;
+  virtual void copyFrom(const double* buffer) = 0;
+
+  /* copies 'src' into this as a block starting at (i_block_start,j_block_start) */
+  void copyBlockFromMatrix(const long i_block_start, const long j_block_start, const hiopMatrix& src) = 0;
```
cnpetra, line 91: this method makes sense only to particular matrices (dense and maybe mds), and I would not make it part of the interface.

src/LinAlg/hiopMatrix.hpp
```diff
@@ -84,6 +84,16 @@ class hiopMatrix
   virtual hiopMatrix* alloc_clone() const=0;
   virtual hiopMatrix* new_copy() const=0;
 
+  virtual void copyFrom(const hiopMatrix& matrix) = 0;
+  virtual void copyFrom(const double* buffer) = 0;
```
cnpetra, line 88: this method does not make sense for sparse of mds matrices (specific to dense matrices in other words). I would not make it part of the interface.

src/LinAlg/hiopMatrix.hpp
```diff
@@ -84,6 +84,16 @@ class hiopMatrix
...
+
+  /* copies 'src' into this as a block starting at (i_block_start,j_block_start) */
+  void copyBlockFromMatrix(const long i_block_start, const long j_block_start, const hiopMatrix& src) = 0;
+
+  /* overwrites 'this' with 'src''s block that starts at (i_src_block_start,j_src_block_start) 
+   * and has dimensions of 'this' */
+  void copyFromMatrixBlock(const hiopMatrix& src, const int i_src_block_start, const int j_src_block_start) = 0;
```
cnpetra, line 95: this method is needed only for dense and can be efficiently implemented only dense matrices. I would not make it part of the interface.

cnpetra:
hiopVector may be in need of a lifting a couple of methods, see #23.

ashermancinelli:
You may PR my branch for the sake of cleanliness, or you can send me a diff that I can apply manually. However you would like to proceed is well with me.

cnpetra:
better to not lift the `copyXXX` methods as previously discussed

`addSubDiagonal` is addressed in PR#42 and will move the discussion there

will be closing this (without merge), but feel free to reopen if I missed anything.

PR #31, opened 2020-05-14
Options for building shared and static library

ashermancinelli:
Builds shared lib as in #26 with options for shared and static.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,15 @@ option(HIOP_WITH_KRON_REDUCTION "Build Kron Reduction code (requires MA86)" OFF)
 option(HIOP_DEVELOPER_MODE "Build with extended warnings and options" OFF)
 #with testing drivers capable of 'selfchecking' (-selfcheck)
 option(HIOP_WITH_MAKETEST "Enable 'make test'" ON)
+option(HIOP_BUILD_SHARED "Build shared library" OFF)
+option(HIOP_BUILD_STATIC "Build static library" ON)
+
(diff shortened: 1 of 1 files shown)


cnpetra:
Everything under `plan\` is supposed to be private-ish ;)

cnpetra:
since this PR addresses build system issues, on my mac this line in `CMakeLists.txt`
```
target_link_libraries(hiop_math INTERFACE OpenMP::OpenMP_CXX)
```
causes cmake to crap out. If I comment the above line, everything works. The underlying issue is that I don't have C++ OpenMP libraries. But I have Fortran OpenMP, which seems to be enough since everything works. Do we really need to specifically use OpenMP_CXX?

ashermancinelli:
I cleaned my history and cherry-picked the commits so that `plan/` is removed from the history. Sorry about that!

cnpetra:
my bad, I should have move it from the gitlab hiop

ashermancinelli:
I should never have used the openmpi-specific package, the latest commit should fix this.

cnpetra:
be aware, OpenMP not the same OpenMPI

ashermancinelli:
ahh, I see. I thought `OpenMP_CXX` only referred to the cpp bindings, but I guess I was wrong.

ashermancinelli:
I think [this link](https://github.com/Kitware/CMake/blob/master/Modules/FindOpenMP.cmake#L56) indicates that the 'cxx' in `OpenMP_CXX` only refers to the language openmp will be used from. Very odd that it would cause problems in your build.

ashermancinelli:
@cnpetra Does the most recent build still fail on your machine?

PR #33, opened 2020-05-15
build script to run all builds recommended in the readme

ashermancinelli:
This would prevent future (and present) contributors from being confused about the standards for contribution. All PR's should have to pass this build script before submission. This runs all the builds and tests recommended in `README_developers.md`, along with clang sanitize if on osx. This is similar to our internal CI pipeline, however that only runs one build. If you do intend to have every PR pass every build, then a script like this seems useful.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -3,6 +3,7 @@
 [[ -f /etc/prfile.d/modules.sh ]] && \
   source /etc/profile.d/modules.sh
 
+declare -a cmake_args
 builddir=build
 [ -d $builddir ] && rm -rf $builddir; mkdir $builddir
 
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -0,0 +1,101 @@
...
+    magma=2.5.2_cuda10.2
+    metis=5.1.0
+  else
+    gcc=7.3.0
+    cmake=3.15.3
+    openmpi=3.1.3
+    cuda=10.1.243
+    magma=2.5.2_cuda10.2
+    metis=5.1.0
+  fi
```
pelesh, line 30:
```suggestion
  else
    echo "Generic build ..."
  fi
```
The default option should not try to load any modules.

build.sh
```diff
@@ -0,0 +1,101 @@
...
+
+if type module
+then
+  proj_dir=/qfs/projects/exasgd
+  mod_dir=/share/apps
+  cluster=$(uname -n | cut -d'.' -f1 | sed -E 's/[[:digit:]]//g')
+  module purge
+  if [ "$cluster" == "newell" ]
+  then
+    gcc=7.4.0
```
pelesh, line 17: Perhaps user has loaded modules on a machine that is not available as an option in this script, so she does not want to purge modules. Try something like:
```suggestion
  cluster=$(uname -n | cut -d'.' -f1 | sed -E 's/[[:digit:]]//g')
  if [ "$cluster" == "newell" ]
  then
    proj_dir=/qfs/projects/exasgd
    mod_dir=/share/apps
    module purge
    gcc=7.4.0
```

build.sh
```diff
@@ -0,0 +1,101 @@
...
+    -DHIOP_UMFPACK_DIR=$proj_dir/$cluster/suitesparse \
+    -DHIOP_METIS_DIR=$mod_dir/metis/5.1.0"
+  "-DHIOP_USE_MPI=ON -DHIOP_DEEPCHECKS=OFF -DCMAKE_BUILD_TYPE=RELEASE"
+  "-DHIOP_USE_MPI=OFF -DHIOP_DEEPCHECKS=OFF -DCMAKE_BUILD_TYPE=RELEASE"
+  "-DHIOP_USE_MPI=ON -DHIOP_DEEPCHECKS=ON -DCMAKE_BUILD_TYPE=RELEASE"
+  "-DHIOP_USE_MPI=OFF -DHIOP_DEEPCHECKS=ON -DCMAKE_BUILD_TYPE=RELEASE"
+  "-DHIOP_USE_MPI=ON -DHIOP_DEEPCHECKS=OFF -DCMAKE_BUILD_TYPE=DEBUG"
+  "-DHIOP_USE_MPI=OFF -DHIOP_DEEPCHECKS=OFF -DCMAKE_BUILD_TYPE=DEBUG"
+  "-DHIOP_USE_MPI=ON -DHIOP_DEEPCHECKS=ON -DCMAKE_BUILD_TYPE=DEBUG"
+)
```
pelesh, line 54: Flags should probably be specific to the cluster where the test is run.

build.sh
ashermancinelli: The modules will only be purged within the environment of the script, not in the user's environment. If the user's build needs a new module loaded, that should be part of the PR as well I think. Not purging modules could shadow a new dependency that would not show up in the build script but would fail on other dev's devices.

build.sh
ashermancinelli: Good call

build.sh
ashermancinelli: Good point

cnpetra:
note: can we automatically enforce style guidelines on the PR's (git) diff?

ashermancinelli:
Yes, I would run [this script](https://clang.llvm.org/extra/doxygen/clang-tidy-diff_8py_source.html) in the CI to only enforce style guides on the diff of the PR.

PR #36, opened 2020-05-27
Unit Testing Framework

ashermancinelli:
At PNNL we have been developing a unit testing framework based off of the work of the [Sundials](https://github.com/LLNL/sundials) developers.

The tests are intended to mirror the layout of the source directory. Currently, we only have tests for classes defined in `LinAlg`, so that is the only subdirectory in the `tests` directory. Tests for all classes within `src/LinAlg` will live in `tests/LinAlg`, tests for classes within `src/Optimization` will live in `tests/Optimization`, etc.

At the highest level within the `tests` directory live the drivers, currently only for testing vector and matrix classes.

```console
hiop
│
├── src
│   ├── Drivers
│   ├── Interface
│   ├── _Internals
│   ├── LinAlg
│   ├── Optimization
│   └── Utils
└── tests
    ├── CMakeLists.txt
    ├── LinAlg
    ├── testMatrix.cpp
    └── testVector.cpp
```

The drivers are ran within CTest, both sequentially and with MPI.

At the lowest level, the test classes inherit from testBase, which contains utility methods to be used in all tests. From there, each test class adds methods specific to that implementation/interface.
```console
hiop/tests
├── CMakeLists.txt
├── LinAlg
│   ├── matrixTestsDense.cpp
│   ├── matrixTestsDense.hpp
│   ├── matrixTests.hpp
│   ├── testBase.hpp
│   ├── vectorTests.hpp
│   ├── vectorTestsPar.cpp
│   └── vectorTestsPar.hpp
├── testMatrix.cpp
└── testVector.cpp
```
Methods defined only in the abstract interface for `hiopMatrix` live in `matrixTests.hpp`, while methods specific to `hiopMatrixDense` will live in `matrixTestsDense.hpp`. Moving forward, we will have tests for sparse matrices in `matrixTestsSparse.hpp` and so on.

`matrixTests.hpp` defines helper methods specific to each implementation as well, such that the tests are implementation-agnostic. For example, each test class must provide `getLocalElement` and `setLocalElement` methods specific to the data layout of that implementation, among others. In this way, each new implementation need only implement the implementation-specific helper methods for the test class to have access to the entire test suite.

For example, when we added a RAJA implementation of `hiopVector`, we needed only to add a test class `vectorTestsRaja.hpp` which implemented `getLocalElement` (and other helper methods) and we were able to run the full test suite on our RAJA implementation.

In this manner, we hope that our test suite will facilitate HiOp's transition to multiple implementations.


ashermancinelli:
Closing for now until we clean up a few issues and add a few features on our end.

Issue #37, opened 2020-05-28
hiopMatrixDense shiftRows segfault


ashermancinelli:
When testing the `shiftRows` method of `hiopMatrixDense`, I am running into a segfault on our Power9 systems:

```console
233	    A.shiftRows(shift);
(gdb) 

Program received signal SIGSEGV, Segmentation fault.
0x0000000010017474 in hiop::hiopMatrixDense::shiftRows (this=0x7fffffffe410, shift=4) at /people/manc568/projects/hiop/src/LinAlg/hiopMatrix.cpp:256
256	    assert(test1==M[shift<0?0:m_local][0] && "a different copy technique than memcpy is needed on this system");
```
Should we add a fallback method in case memcpy will not work? How do you suggest we test this method? @cnpetra

cnpetra:
test assert test is incorrect for shift > 0

fixed by 06185762879d3e71df7a14de5c71f06f7c71fbc3

ashermancinelli:
After applying your patch I still ran into the a segfault. I implemented the method using `std::rotate` in my branch, would you mind taking a look?

cnpetra:
should be 
`
| 2 2 2 |
| 1 1 1  |
| 2 2 2 |
`
row that are shifted out of the matrix do not enter back at the top of the matrix.

ashermancinelli:
No wrapping, good to know.

cnpetra:
yes, no need to `rotate`

cnpetra:
please confirm whether the master `memcpy` works or shiftrows-forloop-fix `for` copy loop works (aa77c056d28b159f296b93db3d0291aeaf9a8029)?

ashermancinelli:
Your original `memcpy` method is working.

PR #38, opened 2020-05-28
switching memcpy for std::rotate

ashermancinelli:
Please find #37

### DIFF
--- a/src/LinAlg/hiopMatrix.cpp
+++ b/src/LinAlg/hiopMatrix.cpp
@@ -224,39 +224,22 @@ void hiopMatrixDense::copyFromMatrixBlock(const hiopMatrixDense& src, const int 
 
 void hiopMatrixDense::shiftRows(long long shift)
 {
+  shift=-shift; // To preserve shift>0 -> up and shift<0 -> down
   if(shift==0) return;
   if(-shift==m_local) return; //nothing to shift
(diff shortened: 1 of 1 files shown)


cnpetra:
I am pretty sure your update is not going to work (you're rotating pointers to columns). `std::rotate` is a good idea, however, it is slightly more expensive than what I had in mind. Will take another stab at the function, without memcpy (btw, the function works perfectly on all three OS I have).

cnpetra:
is this working ?

https://github.com/LLNL/hiop/commit/aa77c056d28b159f296b93db3d0291aeaf9a8029

cnpetra:
works on mac and ubuntu

ashermancinelli:
I see, you're right about my std::rotate method. My confusion about the current method however. Given a matrix A:
```c++
| 2 2 2 |
| 1 1 1 |
| 1 1 1 |
```
calling `A.shiftRows(2);` results in:
```c++
| 2 2 2 |
| 1 1 1 |
| 2 2 2 |
```
Where I expected this:
```c++
| 1 1 1 |
| 1 1 1 |
| 2 2 2 |
```
Is this behavior intended?

ashermancinelli:
Moving discussion to the issue now that we established the infeasibility of the PR.

Issue #46, opened 2020-06-13
Build issue


cnpetra:
Two reports of build failure came in the last week via email. Coincidentally, I've just encountered the same exact problem on summit (using cmake 3.9.2)

```cmake ..
CMake Error at CMakeLists.txt:7 (cmake_policy):
  Policy "CMP0074" is not known to this version of CMake.


-- The C compiler identification is GNU 4.8.5
-- The CXX compiler identification is GNU 4.8.5
(...)
-- Found LAPACK libraries: /autofs/nccs-svm1_sw/summit/.swci/1-compute/opt/spack/20180914/linux-rhel7-ppc64le/gcc-9.1.0/openblas-0.3.9-aymovpat33osbzgh5gsmhyvstsol4sfp/lib/libopenblas.so;/autofs/nccs-svm1_sw/summit/.swci/1-compute/opt/spack/20180914/linux-rhel7-ppc64le/gcc-9.1.0/openblas-0.3.9-aymovpat33osbzgh5gsmhyvstsol4sfp/lib/libopenblas.so
CMake Error at src/Optimization/CMakeLists.txt:2 (target_link_libraries):
  Object library target "hiopOptimization" may not link to anything.


-- Configuring incomplete, errors occurred!
See also "/ccs/home/cpetra/work/projects/hiop/build/CMakeFiles/CMakeOutput.log".
See also "/ccs/home/cpetra/work/projects/hiop/build/CMakeFiles/CMakeError.log".
```

@ashermancinelli

cnpetra:
but 

```
module load cmake/3.15.2
rm -f *; cmake ..
```

works

ashermancinelli:
I see. We manually enable a policy to suppress warnings, but we should have checked that the minimum required cmake supported that policy. I'll have a PR out soon.

pelesh:
I think this will do it:
```cmake
if (POLICY CMP0074)
  cmake_policy(SET CMP0074 NEW) # CMake 3.12
endif ()
```

ashermancinelli:
It seems we have other issues building with older cmake versions as well. Is it too much to ask to make cmake/12 our lowest supported version?

cnpetra:
cmake min version 3.12 is a bit too constraining at this time...

cnpetra:
after weighing the pros and cons with @pelesh let go with min ver 3.12

ashermancinelli:
I believe this is resolved with #57

cnpetra:
correct

PR #49, opened 2020-06-13
checking for cmake >= 12 before setting policy introduced in that release

ashermancinelli:
Please see #46

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required (VERSION 3.8)
+cmake_minimum_required (VERSION 3.12)
 
 set(CMAKE_CXX_STANDARD 11)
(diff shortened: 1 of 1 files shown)


cnpetra:
looks like it does not fix the issue

cmake's GREATER_EQUAL does not work as you think it does. Apparently, it performs string comparison, meaning (cmake version) 3.9.2 > 3.11.2

pelesh:
@ashermancinelli the correct fix is to check if the policy is defined, not to check for CMake version. See my comment in #46.

ashermancinelli:
I've built with every version of gcc and cmake we have on newell. I'll have to build some older versions of cmake manually to test below cmake 12.1.

ashermancinelli:
closing as per offline conversation.

ashermancinelli:
Closing in favor of more comprehensive PR.

PR #52, opened 2020-06-29
Install rpath fix

ashermancinelli:
We had some internal issues when using kron reduction, since the current build does not configure the rpath correctly, and this patch fixed it for us.


ashermancinelli:
Closing in favor of more comprehensive PR.

PR #48, opened 2020-06-13
Removed incorrectly included header

ashermancinelli:
See #47

PR #57, opened 2020-07-13
Cmake Improvements

ashermancinelli:
We've had three PRs open for a while now, and we had some additional smaller internal PRs that we wrapped into one. This PR should address #48 #49 #52.

CC @pelesh


cnpetra:
same issue with mpi wrapping as reported in PR #59. Only on some systems

PR #60, opened 2020-07-20
updated startingAtCopyFromStartingAt

cnpetra:
Update of the confusing documentation and implementation of vector's startingAtCopyFromStartingAt

### DIFF
--- a/src/LinAlg/hiopVectorPar.cpp
+++ b/src/LinAlg/hiopVectorPar.cpp
@@ -153,25 +153,26 @@ void hiopVectorPar::copyFromStarting(int start_index/*_in_src*/,const hiopVector
   memcpy(data_+start_index, v.data_, v.n_local_*sizeof(double));
 }
 
-void hiopVectorPar::startingAtCopyFromStartingAt(int start_idx_src, 
-						 const hiopVector& v_, 
-						 int start_idx_dest)
+void hiopVectorPar::startingAtCopyFromStartingAt(int start_idx_dest, 
+						 const hiopVector& v_in, 
+						 int start_idx_src)
 {
 #ifdef HIOP_DEEPCHECKS
   assert(n_local_==n_ && "only for local/non-distributed vectors");
 #endif
-  assert((start_idx_src>=0 && start_idx_src<this->n_local_) || this->n_local_==0);
-  const hiopVectorPar& v = dynamic_cast<const hiopVectorPar&>(v_);
-  assert((start_idx_dest>=0 && start_idx_dest<v.n_local_) || v.n_local_==0);
+  assert((start_idx_dest>=0 && start_idx_dest<this->n_local_) || this->n_local_==0);
+  const hiopVectorPar& v = dynamic_cast<const hiopVectorPar&>(v_in);
+  assert((start_idx_src>=0 && start_idx_src<v.n_local_) || v.n_local_==0);
 
-  int howManyToCopy = this->n_local_ - start_idx_src;
-  const int howManyToCopyDest = v.n_local_-start_idx_dest;
-  assert(howManyToCopy <= howManyToCopyDest);
-  //howManyToCopy = howManyToCopy <= v.n_local_-start_idx_dest ? howManyToCopy : v.n_local_-start_idx_dest;
-  if(howManyToCopy > howManyToCopyDest) howManyToCopy = howManyToCopyDest;
+  int howManyToCopy = this->n_local_ - start_idx_dest;
+  const int howManyToCopySrc = v.n_local_-start_idx_dest;
+  assert(howManyToCopy <= howManyToCopySrc);
+
+  //just to be safe when not NDEBUG
+  if(howManyToCopy > howManyToCopySrc) howManyToCopy = howManyToCopySrc;
 
   assert(howManyToCopy>=0);
-  memcpy(data_+start_idx_src, v.data_+start_idx_dest, howManyToCopy*sizeof(double));
+  memcpy(data_+start_idx_dest, v.data_+start_idx_src, howManyToCopy*sizeof(double));
 }
 
 void hiopVectorPar::copyToStarting(int start_index, hiopVector& v_)
--- a/src/LinAlg/hiopVectorPar.cpp
+++ b/src/LinAlg/hiopVectorPar.cpp
@@ -165,7 +165,7 @@ void hiopVectorPar::startingAtCopyFromStartingAt(int start_idx_dest,
   assert((start_idx_src>=0 && start_idx_src<v.n_local_) || v.n_local_==0);
 
   int howManyToCopy = this->n_local_ - start_idx_dest;
-  const int howManyToCopySrc = v.n_local_-start_idx_dest;
+  const int howManyToCopySrc = v.n_local_-start_idx_src;
   assert(howManyToCopy <= howManyToCopySrc);
 
   //just to be safe when not NDEBUG
--- a/src/LinAlg/hiopVector.hpp
+++ b/src/LinAlg/hiopVector.hpp
@@ -74,8 +74,17 @@ public:
   /// @brief Copy v in 'this' starting at start_index_in_src in  'this'. */
   virtual void copyFromStarting(int start_index_in_src, const hiopVector& v) = 0;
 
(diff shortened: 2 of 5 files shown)


src/LinAlg/hiopVectorPar.cpp
```diff
@@ -153,25 +153,26 @@ void hiopVectorPar::copyFromStarting(int start_index/*_in_src*/,const hiopVector
...
-  assert((start_idx_src>=0 && start_idx_src<this->n_local_) || this->n_local_==0);
-  const hiopVectorPar& v = dynamic_cast<const hiopVectorPar&>(v_);
-  assert((start_idx_dest>=0 && start_idx_dest<v.n_local_) || v.n_local_==0);
+  assert((start_idx_dest>=0 && start_idx_dest<this->n_local_) || this->n_local_==0);
+  const hiopVectorPar& v = dynamic_cast<const hiopVectorPar&>(v_in);
+  assert((start_idx_src>=0 && start_idx_src<v.n_local_) || v.n_local_==0);
+
+  int howManyToCopy = this->n_local_ - start_idx_dest;
+  const int howManyToCopySrc = v.n_local_-start_idx_dest;
+  assert(howManyToCopy <= howManyToCopySrc);
```
pelesh, line 169: Do you think the assertion should be the other way around?
```suggestion
  int howManyToCopy = this->n_local_ - start_idx_dest;
  const int howManyToCopySrc = v.n_local_-start_idx_dest;
  assert(howManyToCopySrc <= howManyToCopy);
```
I would think that the number of source elements should not exceed the number of destination elements. Otherwise, one may end up writing passed the end of the array, which can lead to undefined behavior. I don't see anything in the code what would prevent this from happening.

src/LinAlg/hiopVectorPar.cpp
```diff
@@ -153,25 +153,26 @@ void hiopVectorPar::copyFromStarting(int start_index/*_in_src*/,const hiopVector
...
-  const int howManyToCopyDest = v.n_local_-start_idx_dest;
-  assert(howManyToCopy <= howManyToCopyDest);
-  //howManyToCopy = howManyToCopy <= v.n_local_-start_idx_dest ? howManyToCopy : v.n_local_-start_idx_dest;
-  if(howManyToCopy > howManyToCopyDest) howManyToCopy = howManyToCopyDest;
+  //just to be safe when not NDEBUG
+  if(howManyToCopy > howManyToCopySrc) howManyToCopy = howManyToCopySrc;
 
   assert(howManyToCopy>=0);
-  memcpy(data_+start_idx_src, v.data_+start_idx_dest, howManyToCopy*sizeof(double));
+  memcpy(data_+start_idx_dest, v.data_+start_idx_src, howManyToCopy*sizeof(double));
```
pelesh, line 175: If assertion `howManyToCopySrc <= howManyToCopy`, then following will ensure that no elements are read past the end of source array and no elements are overwritten past the end of the destination array:
```suggestion
  memcpy(data_+start_idx_dest, v.data_+start_idx_src, howManyToCopySrc*sizeof(double));
```

src/LinAlg/hiopVectorPar.cpp
```diff
@@ -153,25 +153,26 @@ void hiopVectorPar::copyFromStarting(int start_index/*_in_src*/,const hiopVector
...
 #endif
-  assert((start_idx_src>=0 && start_idx_src<this->n_local_) || this->n_local_==0);
-  const hiopVectorPar& v = dynamic_cast<const hiopVectorPar&>(v_);
-  assert((start_idx_dest>=0 && start_idx_dest<v.n_local_) || v.n_local_==0);
+  assert((start_idx_dest>=0 && start_idx_dest<this->n_local_) || this->n_local_==0);
+  const hiopVectorPar& v = dynamic_cast<const hiopVectorPar&>(v_in);
+  assert((start_idx_src>=0 && start_idx_src<v.n_local_) || v.n_local_==0);
+
+  int howManyToCopy = this->n_local_ - start_idx_dest;
+  const int howManyToCopySrc = v.n_local_-start_idx_dest;
```
pelesh, line 168: This seems to be a bug:
```suggestion
  const int howManyToCopySrc = v.n_local_-start_idx_src;
```

pelesh:
Consider following use case:
```c++
hiopVectorPar dest(10);
hiopVectorPar src(10);
dest.startingAtCopyFromStartingAt(0, src, 9);
```
This is supposed to copy last element in `src` to the first position in `dest` vector. However, the assertion on line 169 will fail, since there are 10 elements where to copy in `dest`, while there is only one element to be copied from `src`. It is not clear to me why this assertion is in place and why is this use case prohibited?

Assuming the assertion on 169 is needed, then it seems to me it must hold `howManyToCopy == howManyToCopySrc`.

cnpetra:
> Consider following use case:
> 
> ```c++
> hiopVectorPar dest(10);
> hiopVectorPar src(10);
> dest.startingAtCopyFromStartingAt(0, src, 9);
> ```
> 
> This is supposed to copy last element in `src` to the first position in `dest` vector. However, the assertion on line 169 will fail, since there are 10 elements where to copy in `dest`, while there is only one element to be copied from `src`. It is not clear to me why this assertion is in place and why is this use case prohibited?
> 
> Assuming the assertion on 169 is needed, then it seems to me it must hold `howManyToCopy == howManyToCopySrc`.

the precondition of the method is not satisfied. 
```
/*
* ...
* Precondition: The method expects that in 'v' there are at least as many elements starting 
   * 'start_idx_src' as 'this' has starting at 'start_idx_dest', or in other words,
   * length(this) - start_idx_dest <= length(v) - start_idx_src
  */
```

cnpetra:
intended use
```
hiopVectorPar dest(10);
hiopVectorPar src(100000);
dest.startingAtCopyFromStartingAt(0, src, 9); //will copy 10 elements of src starting elem 9 into dest
```

cnpetra:
your change
```
const int howManyToCopySrc = v.n_local_-start_idx_src;
```
is a good catch though

pelesh:
> intended use
> 
> ```
> hiopVectorPar dest(10);
> hiopVectorPar src(100000);
> dest.startingAtCopyFromStartingAt(0, src, 9); //will copy into dest 10 elements starting at index 0 from src starting elem 9 
> ```
> 
> or
> 
> ```
> hiopVectorPar dest(10);
> hiopVectorPar src(100000);
> dest.startingAtCopyFromStartingAt(6, src, 9); //will copy into dest 4 elements starting at index 6 from src starting at elem 9
> ```

I see. So the post condition is that all destination elements in the selected range must be overwritten by source elements, correct? Therefore `length(this) - start_idx_dest <= length(v) - start_idx_src` must hold.

src/LinAlg/hiopVectorPar.cpp
```diff
@@ -153,25 +153,26 @@ void hiopVectorPar::copyFromStarting(int start_index/*_in_src*/,const hiopVector
...
+  const int howManyToCopySrc = v.n_local_-start_idx_src;
+  assert(howManyToCopy <= howManyToCopySrc);
 
-  int howManyToCopy = this->n_local_ - start_idx_src;
-  const int howManyToCopyDest = v.n_local_-start_idx_dest;
-  assert(howManyToCopy <= howManyToCopyDest);
-  //howManyToCopy = howManyToCopy <= v.n_local_-start_idx_dest ? howManyToCopy : v.n_local_-start_idx_dest;
-  if(howManyToCopy > howManyToCopyDest) howManyToCopy = howManyToCopyDest;
+  //just to be safe when not NDEBUG
+  if(howManyToCopy > howManyToCopySrc) howManyToCopy = howManyToCopySrc;
```
pelesh, line 172: Does this contradict the assertion on line 169? The assertion requires that there are sufficiently many elements in source to overwrite all elements at the destination starting from `start_idx_dst`. It seems as if this is not adequate safety when the code is built with `NDEBUG` flag, because it will simply allow for some destination elements not to be overwritten. Perhaps it is better to have here something that will crash the code instead of allowing it quietly to run after post-condition is violated? For example:
 ```suggestion
  if(howManyToCopy > howManyToCopySrc) { delete data_; data_ = nullptr};
```

cnpetra: intended to quietly run in release/ndebug (instead of undefined behavior)

PR #62, opened 2020-07-27
Kron reduction interface update

cnpetra:
Light update of the interface to support the application (mat-vec) of the mapping from nonauxiliary to auxiliary buses


pelesh:
Perhaps it would be good to resync with `master` before merging. Git will merge without conflicts, but the code won't build. It should be an easy fix, though.

PR #69, opened 2020-08-13
Fix build issues.

pelesh:
I think I fixed build issues identified in #51. This also simplifies a little the build system.


pelesh:
Just to add, I tested this with GPU support turned on and off. It seems that HiOp's C-interface does not work when GPU is enabled. Perhaps @michel2323 should take a look, as well?

CC @ashermancinelli

cnpetra:
thanks. works fine on all systems tested, including summit (with gpu support on).

Issue #71, opened 2020-08-19
nlpMDS_ex5.exe fails on Broadwell/Volta100


pelesh:
The `nlpMDS_ex5.exe` test fails to converge on Intel platform with Volta GPU. HiOp is built with GPU support and Kron reduction enabled. When running with 1 MPI rank and on one GPU device, following error message is obtained:
```bash
...

[Warning] hiopLinSolverMagmaBuka error: 191 entry in the factorization's diagonal
is exactly zero. Division by zero will occur if it a solve is attempted.
[Warning] hiopLinSolverMagmaBuka error: 191 entry in the factorization's diagonal
is exactly zero. Division by zero will occur if it a solve is attempted.
[Warning] KKT_MDS_XYcYd linsys: Detected negative eigenvalues in (1,1) sparse block.
  83 -1.3559463e+03 1.260e-02  5.259e-03  -2.55  1.000e+00  1.000e+00  1(f)
[Warning] hiopLinSolverMagmaBuka error: 191 entry in the factorization's diagonal
is exactly zero. Division by zero will occur if it a solve is attempted.
[Warning] hiopLinSolverMagmaBuka error: 191 entry in the factorization's diagonal
is exactly zero. Division by zero will occur if it a solve is attempted.
[Warning] KKT_MDS_XYcYd linsys: Detected negative eigenvalues in (1,1) sparse block.
  84 -1.3560463e+03 4.662e-02  2.022e-03  -3.82  1.000e+00  1.000e+00  1(h)
[Warning] hiopLinSolverMagmaBuka error: 191 entry in the factorization's diagonal
is exactly zero. Division by zero will occur if it a solve is attempted.
[Warning] hiopLinSolverMagmaBuka error: 191 entry in the factorization's diagonal
is exactly zero. Division by zero will occur if it a solve is attempted.
[Warning] KKT_MDS_XYcYd linsys: Detected negative eigenvalues in (1,1) sparse block.
Panic: minimum step size reached. The problem may be infeasible or the gradient inaccurate. Will exit here.
  85 -1.3560463e+03 4.662e-02  2.028e-03  -3.82  1.000e+00  5.551e-17  54(?)
Couldn't solve the problem.
Linesearch returned unsuccessfully (small step). Probable cause: inaccurate gradients/Jacobians or infeasible problem.
Total time 6.014 sec 
Hiop internal time:     total 6.002 sec     avg iter 0.071 sec 
    internal total std dev across ranks 0.000 percent
[...]

PR #73, opened 2020-08-22
Small changes to conform to style guide in hiopMatrixSparse*

ashermancinelli:
Title is self-explanatory. Passes all tests, except for the case described in #71. Mostly just adding trailing underscore to member variables for sparse matrix classes.


cnpetra:
looks great, thanks!

PR #76, opened 2020-08-27
sync with Magma

cnpetra:
(work in progress)
(do not try this at home nor at work :))

### DIFF
--- a/src/LinAlg/hiopLinSolverIndefDenseMagma.cpp
+++ b/src/LinAlg/hiopLinSolverIndefDenseMagma.cpp
@@ -3,9 +3,9 @@
 namespace hiop
 {
   hiopLinSolverIndefDenseMagmaBuKa::hiopLinSolverIndefDenseMagmaBuKa(int n, hiopNlpFormulation* nlp_)
-    : hiopLinSolverIndefDense(n, nlp_)
+    : hiopLinSolverIndefDense(n, nlp_), work_(NULL)
   {
-    ipiv = new int[n];
+    ipiv_ = new int[n];
 
     //
     // commented code - this class is using CPU MAGMA interface for now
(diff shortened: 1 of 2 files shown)


pelesh:
What is the minimum Magma version for this code to work?

cnpetra:
this works with magma master only at this point until the new features we requested make it into the next release.

src/LinAlg/hiopLinSolverIndefDenseMagma.cpp
```diff
@@ -119,9 +136,40 @@ namespace hiop
...
     //printf("(pos,null,neg)=(%d,%d,%d)\n", posEigVal, nullEigVal, negEigVal);
     nlp_->runStats.linsolv.tmInertiaComp.stop();
-    
-    if(nullEigVal>0) return -1;
-    return negEigVal;
+
+    return true;
+  }
+
+  bool hiopLinSolverIndefDenseMagmaBuKaDev::compute_inertia(double **MM,
```
pelesh, line 143:
```suggestion
  bool hiopLinSolverIndefDenseMagmaBuKaDev::compute_inertia(double * MM, 
```
Perhaps it would be better if we pass the pointer to matrix data block here? That would avoid dereferencing pointer a few lines below.

src/LinAlg/hiopLinSolverIndefDenseMagma.cpp
```diff
@@ -119,9 +136,40 @@ namespace hiop
...
+
+    nlp_->runStats.linsolv.tmInertiaComp.start();
+
+    if(NULL==work_) {
+      work_ = new double[N];
+    }
+
+
+    //determinant if only for logging/output/curiosity purposes
+    magma_dsidi(MM[0], N, N, ipiv, det, inert, work_, job, &info);
```
pelesh, line 163:
```suggestion
    magma_dsidi(MM, N, N, ipiv, det, inert, work_, job, &info);
```
If `MM` is `double*` type we do not need to dereference here.

cnpetra: @ashermancinelli and I are pm'ed about this recently as I am looking to solve issue #81.

pelesh:
@cnpetra, is the plan to merge this directly to `master` or `develop`?

cnpetra:
reopened as PR#156

Issue #79, opened 2020-09-03
Counting negatives in Hxs should be implemented as a vector kernel?


pelesh:
Consider implementing loop that is counting negative eigenvalues in [`hiopKKTLinCompressedMDSXYcYd` class](https://github.com/LLNL/hiop/blob/master/src/Optimization/hiopKKTLinSysMDS.cpp#L274) as a vector kernel and adding it to the abstract vector interface. Otherwise, method `hiopKKTLinSysCompressedMDSXYcYd::update` depends on implementation detail of the `Hxs_` vector.

This loop will not run on GPU, for example.

cnpetra:
indeed, thanks for reporting this

cnpetra:
addressed by eca78f9f8db9c9cbefa6b1a4b44568b6a10f1f39

PR #85, opened 2020-09-23
add support to sparse linear algebra. 

nychiang:
Adds an NLP general sparse problem formulation and interface, new sparse linear algebra, and backend linear solvers (MA57 and StrumPack). The PR focuses on software and some math aspects and does not fully support device computations. These will be addressed by a future PR.


src/LinAlg/hiopMatrixSparseTripletStorage.hpp
```diff
@@ -134,11 +134,11 @@ namespace hiop {
...
   protected:
     friend class hiopMatrixComplexSparseTriplet;
-  
+
     Tidx nrows, ncols, nnz;
-  
+
     Tidx *irow, *jcol;
-    Tval *values; 
+    Tval *values;
```
pelesh, line 141:
```suggestion
    Tval* values_;
```
Perhaps it would be good to follow HiOp style guidelines.

src/LinAlg/hiopMatrixSparseTriplet.cpp
```diff
@@ -435,17 +466,74 @@ void hiopMatrixSparseTriplet::copyRowsFrom(const hiopMatrix& src_gen,
...
-				    int rank/*=-1*/) const 
+
+/**
+ * @brief Copy 'n_rows' rows started from 'rows_src_idx_st' (array of size 'n_rows') from 'src' to the destination,
+ * which starts from the 'rows_dest_idx_st'th row in 'this'
+ *
+ * @pre 'this' must have exactly, or more than 'n_rows' rows
+ * @pre 'this' must have exactly, or more cols than 'src'
+ */
+void hiopMatrixSparseTriplet::copyRowsFromSrcToDest(const hiopMatrix& src_gen,
```
pelesh, line 483: Please add a unit test for this kernel.

src/LinAlg/hiopMatrixSparseTriplet.cpp
```diff
@@ -141,14 +141,45 @@ void hiopMatrixSparseTriplet::addSubDiagonal(const double& alpha, long long star
   assert(false && "not needed");
 }
 
+void hiopMatrixSparseTriplet::copySubDiagonalEleFromVec(const long long& start_on_dest_diag, const long long& num_elems,
```
pelesh, line 144: I think this kernel needs a unit test.

pelesh:
It would be good to take a look at RAJA snapshot branch to get a heads up on possible merge conflicts. RAJA PR will be submitted soon, so it would be good to coordinate.

cnpetra:
@pelesh RAJA PR will be merged first in master

src/Drivers/nlpMDS_ex5.hpp
```diff
@@ -201,22 +201,22 @@ class Ex5 : public hiop::hiopInterfaceMDS
...
     //return false so that HiOp will rely on the on-call constraint evaluator defined below
     return false;
   }
-  bool eval_cons(const long long& n, const long long& m, 
+  bool eval_cons(const long long& n, const long long& m,
 		 const double* x, bool new_x, double* cons)
   {
     const double* s = x+ns_;
     const double* y = x+2*ns_;
-
```
cnpetra, line 216: @nychiang : end of line is changed in very many places. This will make merging extremely difficult and laborious. I know why it happened ;) I suggest you somehow fix this.

src/Drivers/nlpMDS_ex5.hpp
```diff
@@ -272,16 +272,16 @@ class Ex5 : public hiop::hiopInterfaceMDS
...
+
     return true;
   }
-  
+
   //sum 0.5 {x_i*(x_{i}-1) : i=1,...,ns} + 0.5 y'*Qd*y + 0.5 s^T s
   bool eval_grad_f(const long long& n, const double* x, bool new_x, double* gradf)
   {
     //! assert(ns>=4); assert(Q->n()==ns/4); assert(Q->m()==ns/4);
-    //x_i - 0.5
```
cnpetra, line 283: @nychiang : also here, I guess \n changed with \r\n

src/Drivers/nlpMDS_ex5.hpp
nychiang: I tried to run command dos2unit, but it seems that I need to fix this issue manually. :-)

cnpetra: ouch... maybe core.autocrlf settings of git will help?

nychiang:
Just push another commit, in which I fix the bug and white-space issue.
Ex6 now is working fine. Its usage is:
nlpSparse_ex6.exe n
where n>=3 is a the number of variables provided by the user

nychiang:
NOTE: need to fix the hard-coded path for Metis/Ma57 in CMake file before merging into the master branch

src/LinAlg/hiopMatrixSparseTriplet.cpp
nychiang: add a unit test into test/testMatrixSparse.cpp.
The corresponding code is added into test/LinAlg/matrixTestsSparseTriplet.cpp
However, the corresponding function verifyAnser(...) hasn't been implemented. Will set a follow-up meeting with  Asher @ashermancinelli to discuss this issue.

ashermancinelli: Update: @nychiang and I discussed our unit test conventions, so he should now be able to update accordingly.

tests/LinAlg/matrixTestsSparse.hpp
```diff
@@ -432,10 +500,12 @@ class MatrixTestsSparse : public TestBase
   virtual real_type getLocalElement(const hiop::hiopMatrix* a, local_ordinal_type i, local_ordinal_type j) = 0;
   virtual real_type getLocalElement(const hiop::hiopVector* x, local_ordinal_type i) = 0;
   virtual real_type* getMatrixData(hiop::hiopMatrixSparse* a) = 0;
+  virtual real_type getMatrixData(hiop::hiopMatrixSparse* a, local_ordinal_type i, local_ordinal_type j) = 0;
   virtual const local_ordinal_type* getRowIndices(const hiop::hiopMatrixSparse* a) = 0;
   virtual const local_ordinal_type* getColumnIndices(const hiop::hiopMatrixSparse* a) = 0;
   virtual local_ordinal_type getLocalSize(const hiop::hiopVector* x) = 0;
   virtual int verifyAnswer(hiop::hiopMatrix* A, real_type answer) = 0;
+  virtual int verifyAnswer(hiop::hiopMatrix* A, local_ordinal_type nnz_st, local_ordinal_type nnz_ed, const double answer) = 0;
```
nychiang, line 508: @ashermancinelli Please verify. I add one function to verify the answers for a sparse matrix, since the other "verifyAnswer" requires a dense input matrix "A". See file matrixTestsSparseTriplet.cpp for examples. NOTE that since this branch is developed based on 'master', it doesn't has the RAJA implementation.

ashermancinelli: I think if you're just checking a single element, probably better to use `getLocalElement` and check result

tests/LinAlg/matrixTestsSparse.hpp
```diff
@@ -422,6 +422,74 @@ class MatrixTestsSparse : public TestBase
...
+
+    auto itnz_src{0};
+    auto itnz_dest=B_nnz_st;
+    int fail{0};
+
+    B.copyRowsFromSrcToDest(A, A_rows_st, n_rows, B_rows_st, B_nnz_st);
+
+    auto val = getMatrixData(&B);
+
+    fail += verifyAnswer(&B,0,B_nnz_st,B_val);
```
ashermancinelli, line 484: To be more thorough, we should have another verifyAnswer that takes a lambda for sparse mats, eg
```cpp
fail += verifyAnswer(&B,
  [=] (local_ordinal_type i, local_ordinal_type j) -> real_type
  {
    if(i==0 && j==B_nnz_st) return B_val;
    else if // other two conditions...
    else if // other two conditions...
    else return 0.;
  });
```

nychiang:
I thought about it. The problem is that in this routine, there is no way to get the ‘correct’ number of entries in each row.
See line 117 in the main function testMatrixSparse.cpp:
                initializeSparseTriplet(mxn_sparse, entries_per_row);
which sets the number of entries in each row of a given matrix.

If we can pass this information to the LinAlg layer, I don’t need to do this loop to know how many entries will be copied from A to B.
Instead, I can simply set
                auto nnz_A_need_to_copy = n_rows *  entries_per_row.

NY

tests/LinAlg/matrixTestsSparse.hpp
```diff
@@ -432,10 +500,12 @@ class MatrixTestsSparse : public TestBase
   virtual real_type getLocalElement(const hiop::hiopMatrix* a, local_ordinal_type i, local_ordinal_type j) = 0;
   virtual real_type getLocalElement(const hiop::hiopVector* x, local_ordinal_type i) = 0;
   virtual real_type* getMatrixData(hiop::hiopMatrixSparse* a) = 0;
+  virtual real_type getMatrixData(hiop::hiopMatrixSparse* a, local_ordinal_type i, local_ordinal_type j) = 0;
```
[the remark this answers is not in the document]
nychiang: getLocalElement in class matrixTestsSparseTriplet assumes that the destination matrix "hiop::hiopMatrix* a" is a dense matrix.
See https://github.com/LLNL/hiop/blob/fcd29a04d9b28f57f3ab85f95f969924843a5546/tests/LinAlg/matrixTestsSparseTriplet.cpp#L86

ashermancinelli:
Could you respond via github in the future? Email responses create a lot of garbage. It would be appropriate to pass num entries per row to the test from the test driver.

tests/LinAlg/matrixTestsSparse.hpp
nychiang: I am trying to check multiple values in a sparse matrix, not a single value.
The nonzero indices for these nonzero starts from nnz_st to nnz_ed.
I can change it to a lambda function like you mentioned, but I need to assign it to a different name, or use different declaration.
This is because the current implementation assumes the input matrix is dense 
https://github.com/LLNL/hiop/blob/fcd29a04d9b28f57f3ab85f95f969924843a5546/tests/LinAlg/matrixTestsSparseTriplet.cpp#L211-L224

nychiang:
I noticed it. :-) 
just deleted all the email garbage in my reply.

tests/LinAlg/matrixTestsSparseTriplet.cpp
```diff
@@ -158,6 +178,31 @@ int MatrixTestsSparseTriplet::verifyAnswer(hiop::hiopMatrix* A, const double ans
...
 }
 
+/**
+ * @brief Verifies values of the sparse matrix *only at indices already defined by the sparsity pattern*
+ * This may seem misleading, but verify answer does not check *every* value of the matrix,
+ * but only `nnz` elements with index from nnz_st to nnz_ed
+ *
+ */
+[[nodiscard]]
+int MatrixTestsSparseTriplet::verifyAnswer(hiop::hiopMatrix* A, local_ordinal_type nnz_st, local_ordinal_type nnz_ed, const double answer)
```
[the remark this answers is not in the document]
nychiang: sure!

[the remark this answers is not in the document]
nychiang: fixed! will be added into the next commit.

tests/LinAlg/matrixTestsSparse.hpp
ashermancinelli: I'll submit a pr against your branch to fix this

nychiang: sure. I tried to not touch your existing code and only add my code into the repository.
I believer it will be really helpful if we can fix this issue.
About verifying the answer for a sparse matrix, I do have some different idea. Currently we only verify the nonzero values given by the nonzero index, but I think we may also want to verify the nonzero pattern, For example, verify the kth nonzero is located on some particular row and col.  This may be not urgent, since now we assume the sparsity pattern won't change. If we use sparse matrices correctly, we can skip check the pattern. However, if there is an error in the sparsity pattern when copy to/from another matrix/vector, it will be hard to detect. Please let me know your opinion about this. Cheers,

pelesh: The way I see it there is little use for a `getElement` method for a sparse matrix simply because such method is cumbersome to implement and use. Method `verifyAnswer` is always implementation specific, so it should access matrix data directly, no need to use `getElement` or similar.

pelesh: Verifying sparsity pattern should not be difficult either. Again, I would implement that by accessing matrix data directly and not try to implement `getElement` method for that. Method `getElement` is design to retrieve and check only one single element in a vector or matrix. If you are checking multiple elements, you should be accessing matrix data directly. This is also true for dense matrices.

pelesh:
I believe this PR should target `develop` rather than `master` branch.

cnpetra:
Does this also address issue #79 ?

nychiang:
NO. The patch is ready. Do you want to address issue #79 in this PR?

cnpetra:
> NO. The patch is ready. Do you want to address issue #79 in this PR?

let's address it after the merge as this PR blocks other efforts as well.

Issue #80, opened 2020-09-03
GPU compute mode option


pelesh:
It would be good to add "gpu" as a "compute_mode" option in HiOp. However, this option is not independent of "mem_space" selection. Here is the summary of compute mode and memory space option compatibilities:
1. "cpu" option works only with "default", "host" and likely "um" memory space options.
2. "hybrid" works with all but "device" memory space option.
3. "gpu", if added, would work with "um", "pinned" or "device" options.

We can simply document this and expect user to select a meaningful combination or we can add some logic to HiOp options class that would warn user or when incompatible combination was selected and fall back to the next best thing.

If HiOp is built without RAJA support, only "default" memory space is available.

Options "um", "pinned" or "device" are available only if HiOp is built with GPU support (for now it's CUDA only) turned on.

CC @ashermancinelli @cnpetra

cnpetra:
>  warn user or when incompatible combination was selected and fall back to the next best thing

this can/should be done in 
https://github.com/LLNL/hiop/blob/2252b90bc40f86b8d842da6fec30105feddbf647/src/Utils/hiopOptions.cpp#L292

PR #88, opened 2020-09-25
Portable implementation using Umpire and RAJA libraries

### LINKED ISSUE #72: Review RAJA kernels
Review RAJA kernels (currently in `raja-dev` branch) and flag potential bottlenecks. The purpose of this review is pre-screening of potential performance issues to give us heads up what to pay attention to when profiling the performance.

The current implementation of RAJA kernels was done with objective to ensure accurate computations. Some kernels, such as `hiopVectorRajaPar::projectIntoBounds`, are implemented in a way that is not quite "GPU friendly". Help from RAJA developers in identifying other potential bottlenecks and suggestions how to implement these kernels better is very much appreciated.

RAJA kernels are implemented in following HiOp classes:
- `hiopVectorRajaPar`
- `hiopMatrixRajaDense`
- `hiopMatrixRajaSparseTriplet`
- `hiopMatrixRajaSymSparseTriplet`

Currently, RAJA kernels run only within unit tests. See tests in:
- `testVector.cpp`
- `testMatrix.cpp`
- `testMatrixSparse.cpp`

CC @davidbeckingsale @rhornung67

pelesh:
Enables running linear algebra kernels on GPU or CPU based on compile time selection of execution policies and run-time selection of memory space.
- For now supports OpenMP and CUDA execution policies.
- CUDA execution policies work with unified virtual memory space only for now.
- Example 4 demonstrates use of portability libraries within HiOp

Addresses/closes #72, #80 and #98.

Contributed to this PR: @ashermancinelli @JuanDiegoMontoya @CameronRutherford @davidbeckingsale


BUILD.sh
```diff
@@ -1,69 +1,226 @@
...
+    module load gcc-ext/7.4.0
+    module load spectrum-mpi-ext
+    module load openblas
+    module load magma/2.5.3-cuda11
+    module load metis
+    module load mpfr
+    module load suitesparse
+    module load cmake/3.18.2
+    module load raja
+    module load umpire
```
nychiang, line 108: Tried to run these commands in login1.ascent, but many modules are missing. For example, metis.
Should they be provided by `/gpfs/wolf/proj-shared/csc359/ascent/Modulefiles/Core`?

ashermancinelli: Metis is found above on line 103. What other modules seem to be missing?

src/Drivers/nlpMDSForm_raja_ex4.hpp
```diff
@@ -0,0 +1,218 @@
...
+
+  /**
+   * @todo fill out param descriptions below to determine whether or not
+   * they will reside on device and will have to be accessed/assigned to 
+   * in a RAJA kernel
+   *
+   * @param[out] m ?
+   * @param[out] clow ?
+   * @param[out] cupp ?
+   * @param[out] type ?
```
nychiang, line 72: I think these parameters live on host.

src/LinAlg/hiopMatrixMDS.hpp
```diff
@@ -332,29 +336,36 @@ class hiopMatrixSymBlockDiagMDS : public hiopMatrix
...
-  virtual void addMatrix(double alpha, const hiopMatrix& X)
+  virtual void addMatrix(double alpha, const hiopMatrix& X_in)
   {
-    const hiopMatrixSymBlockDiagMDS* pX=dynamic_cast<const hiopMatrixSymBlockDiagMDS*>(&X);
-    if(pX==NULL) {
-      assert(false && "operation only supported for hiopMatrixMDS left operand");
-    } 
-    mSp->addMatrix(alpha, *pX->mSp);
-    mDe->addMatrix(alpha, *pX->mDe);
+    const hiopMatrixSymBlockDiagMDS& X = dynamic_cast<const hiopMatrixSymBlockDiagMDS&>(X_in);
```
nychiang, line 341: @ashermancinelli A helper function or dynamic_cast?

src/LinAlg/hiopMatrixMDS.hpp
```diff
@@ -435,8 +445,8 @@ class hiopMatrixSymBlockDiagMDS : public hiopMatrix
   }
 #endif
 private:
-  hiopMatrixSymSparseTriplet* mSp;
-  hiopMatrixDense* mDe;
+  hiopMatrixSparse* mSp; ///< Symmetric sparse matrix
+  hiopMatrixDense*  mDe; ///< Row-major dense matrix
 private:
   hiopMatrixSymBlockDiagMDS() : mSp(NULL), mDe(NULL) {};
```
nychiang, line 451: Use "nullptr" (C++11) for consistence.

src/LinAlg/hiopVector.hpp
```diff
@@ -198,6 +198,13 @@ class hiopVector
...
+  virtual double* local_data_host() = 0;
+  virtual const double* local_data_host_const() const = 0;
+
+  virtual void copyToDev() = 0;
+  virtual void copyFromDev() = 0;
+  virtual void copyToDev() const = 0;
+  virtual void copyFromDev() const = 0;
 
 protected:
   long long n_; //we assume sequential data
```
nychiang, line 210: what's the difference between class hiopVector and class hiopVectorIntSeq. Seems both of them assume sequential data, and length of vector is a given value from initialization. The only difference seems to me is that hiopVector asks for more pure functions. Can we link hiopVector with hiopVectorInt or hiopVectorIntSeq?

src/LinAlg/hiopMatrixSparseTriplet.cpp
```diff
@@ -539,8 +550,13 @@ hiopMatrix* hiopMatrixSymSparseTriplet::new_copy() const
...
-/* block of W += alpha*this 
- * Note W; contains only the upper triangular entries */
+/** 
+ * @brief block of W += alpha*this 
+ * @note W contains only the upper triangular entries
+ * 
+ * @warning This method should not be called directly.
+ * Use addUpperTriangleToSymDenseMatrixUpperTriangle instead.
+ */
 void hiopMatrixSymSparseTriplet::addToSymDenseMatrixUpperTriangle(int row_start, int col_start,
```
nychiang, line 560: function `addToSymDenseMatrixUpperTriangle` has never been used in the code. 
This kind of functions, i.e, `addToSymDenseMatrixUpperTriangle` and `transAddToSymDenseMatrixUpperTriangle` are used to build KKT matrix, and hence only `transAddToSymDenseMatrixUpperTriangle` is required in the current code.
Maybe we should remove it from the base class [see here](https://github.com/LLNL/hiop/blob/2252b90bc40f86b8d842da6fec30105feddbf647/src/LinAlg/hiopMatrix.hpp#L140) 
Tested with Ex1,2,4,5, and all of them can run without this function.

src/LinAlg/hiopMatrixSparseTriplet.cpp
```diff
@@ -558,22 +574,19 @@ void hiopMatrixSymSparseTriplet::addToSymDenseMatrixUpperTriangle(int row_start,
...
 }
+
+/** 
+ * @brief block of W += alpha*(this)^T 
+ * @note W contains only the upper triangular entries
+ * 
+ * @warning This method should not be called directly.
+ * Use addUpperTriangleToSymDenseMatrixUpperTriangle instead.
+ */
 void hiopMatrixSymSparseTriplet::transAddToSymDenseMatrixUpperTriangle(int row_start, int col_start,
```
nychiang, line 585: This function has never been called in the current examples.
There may be a bug in this code. It calls `addToSymDenseMatrixUpperTriangle`, which doesn't perform transportation to the matrix.

src/LinAlg/hiopVector.hpp
pelesh: `hiopVectorInt` is a base class for integer vectors. They are supposed to be used for mapping equality and inequality constraints. For now they are not used in the code, but will be needed to enable computations in the device memory instead of unified virtual memory.

src/LinAlg/hiopMatrixSparseTriplet.cpp
pelesh: I think currently `addUpperTriangleToSymDenseMatrixUpperTriangle` is implemented to call `addToSymDenseMatrixUpperTriangle`. Perhaps it would be a good idea to create a separate GitHub issue to discuss how this is best handled? 

When porting to Umpire/RAJA we tried to make minimal changes to HiOp code.

src/LinAlg/hiopMatrixSparseTriplet.cpp
pelesh: Same as before, I would suggest to create a separate issue to discuss this. Only changes to HiOp API we made in this PR were the ones that were absolutely necessary for porting to GPU.

src/Drivers/nlpMDSForm_raja_ex4.hpp
pelesh: I think they do, just as you described. We do need to document this correctly. Thanks for pointing this out.

src/Optimization/hiopNlpFormulation.hpp
```diff
@@ -94,21 +94,21 @@ class hiopNlpFormulation
...
 protected:
   //calls specific hiopInterfaceXXX::eval_Jac_cons and deals with specializations of hiopMatrix arguments
-  virtual bool eval_Jac_c_d_interface_impl(double* x, bool new_x, hiopMatrix& Jac_c, hiopMatrix& Jac_d) = 0;
+  virtual bool eval_Jac_c_d_interface_impl(hiopVector& x, bool new_x, hiopMatrix& Jac_c, hiopMatrix& Jac_d) = 0;
 public:
-  virtual bool eval_Hess_Lagr(const double* x, bool new_x, 
+  virtual bool eval_Hess_Lagr(const hiopVector& x, bool new_x, 
 			      const double& obj_factor,  
 			      const double* lambda_eq, 
 			      const double* lambda_ineq,
```
nychiang, line 114: should we use generic hiopVector kernels for duals? From other discussions, eval_Hess_Lagr should be executed on device.

tests/LinAlg/matrixTests.hpp
```diff
@@ -71,32 +73,60 @@ class MatrixTests : public TestBase
   MatrixTests() {}
   virtual ~MatrixTests(){}
 
-  int matrixSetToZero(hiop::hiopMatrix& A, const int rank=0)
+  int matrixSetToZero(hiop::hiopMatrixDense& A, const int rank)
```
nychiang, line 76: It seems that matrixTests are only used for dense matrix now, while its original design may want to support both dense or sparse input.
I can see we have 2 branches of classes:
1. TestBase->MatrixTests->MatrixTestsDense or MatrixTestsRAJADense 
2. TestBase->MatrixSparseTests->MatrixTestsSparseTriplet or MatrixTestsRAJASparseTriplet 
It is a little bit confusing and hard to tell from the name that MatrixTestsDense inherits from MatrixTests, while MatrixSparseTests doesn't.
To clarify and to keep consistency, how about change the 1st branch to something like:
TestBase->MatrixTestsDense->MatrixTestsDenseRowMajor or MatrixTestsRAJADense

src/Optimization/hiopNlpFormulation.hpp
cnpetra: `double*` are portable to device from what I understand

src/LinAlg/hiopMatrixSparseTriplet.cpp
cnpetra: yes, @nychiang please put that on the TODO list

nychiang (approved):

src/Optimization/hiopNlpFormulation.hpp
pelesh: Actually, @nychiang makes a good point here. If we encapsulate `lambda_eq` and `lambda_ineq` within a HiOp vector then we can eliminate `copyFromStarting` methods that copy from raw arrays and use only the overload that has a HiOp vector as the input.

But, again, this would be beyond the scope of this PR, because `double*` can point to data on device or host, so this is not needed for porting to GPU.

src/LinAlg/hiopMatrixMDS.hpp
pelesh: @nychiang, I am not sure what is your question here. Could you clarify, please?

src/Optimization/hiopNlpFormulation.hpp
cnpetra: I see, that sounds reasonable especially since this is internal / non-interface code.

src/LinAlg/hiopMatrixMDS.hpp
nychiang: This is minor and not relevant to this PR. It is about whether we should use "NULL" or "nullptr" for an empty pointer.
I can tell the original code uses 'NULL', but the new commits show that PNNL folks use "nullptr".
I also use 'nullptr' in my implementation of sparse linear algebra, since 'NULL' has some drawbacks. 
Somehow I think we should keep consistence in this usage.

pelesh: @nychiang, I think you are right. A wholesale transition `NULL` ~> `nullptr` is probably in order, since HiOp adopted C++11 as of 0.3 release.

tests/LinAlg/matrixTests.hpp
pelesh: This is awkward naming scheme. `MatrixTests` refer to dense matrix tests. There is no base class for both dense and sparse matrix tests because the two sets of tests are so different that it does not make sense to have a base class. 

I think now is probably the best time to fix the confusing naming scheme.

BUILD.sh
```diff
@@ -1,69 +1,226 @@
...
+      ;;
+  esac
+done
+
+set -x
+x="unset"
+
+if [[ ! -v MY_CLUSTER ]]
+then
+  export MY_CLUSTER=`uname -n | sed -e 's/[0-9]//g' -e 's/\..*//'`
```
cnpetra, line 56: `uname -n` returns "login1" on ascent :(

src/LinAlg/hiopMatrixMDS.hpp
```diff
@@ -272,16 +281,11 @@ class hiopMatrixSymBlockDiagMDS : public hiopMatrix
   virtual void timesVec(double beta,  hiopVector& y,
 			double alpha, const hiopVector& x) const
   {
-    hiopVectorPar* yp = dynamic_cast<hiopVectorPar*>(&y);
-    const hiopVectorPar* xp = dynamic_cast<const hiopVectorPar*>(&x);
-    assert(yp);
```
cnpetra, line 277: any reason(s) these asserts were removed?

BUILD.sh
ashermancinelli: I usually pass `MY_CLUSTER=ascent ./BUILD.sh` since that sysname discovery can be quite misleading (this is how we run in CI as well). I question whether we should use that at all or require users to specify which cluster they are running on.

tests/LinAlg/matrixTests.hpp
ashermancinelli: Would you like our team to resolve this internally concurrent to this PR or should we incorporate into this PR?

BUILD.sh
ashermancinelli: FYI there seems to be an issue with metis installation on Ascent. We have a separate issue to track this on ORNL gitlab if you have an account there.

src/LinAlg/hiopMatrixMDS.hpp
ashermancinelli: It seems like the dynamic casts were removed to favor using the interface, in which case they're reference types and I don't see the point of keeping those assertions.

src/LinAlg/hiopMatrixMDS.hpp
```diff
@@ -332,29 +336,36 @@ class hiopMatrixSymBlockDiagMDS : public hiopMatrix
...
+   * 
+   * @warning This method is implemented correctly, but should never be called actually.
+   * Use addUpperTriangleToSymDenseMatrixUpperTriangle instead.
+   */
   virtual void transAddToSymDenseMatrixUpperTriangle(int row_start, int col_start, double alpha, hiopMatrixDense& W) const
   {
-    assert(mSp->m() == mSp->n());
-    mSp->transAddToSymDenseMatrixUpperTriangle(row_start,          col_start,          alpha, W);
-    mDe->transAddToSymDenseMatrixUpperTriangle(row_start+mSp->n(), col_start+mSp->n(), alpha, W);
+    addToSymDenseMatrixUpperTriangle(row_start, col_start, alpha, W);
```
ashermancinelli, line 367: I'm questioning the decision to place this function call above the assertion. If some user were to call this function (which we don't want them to do) and an error is raised deeper into the call to `addToSymDenseMatrixUpperTriangle`, the error would be very misleading and potentially hard to trace down. @pelesh why did we leave function calls like this one in the code when we will just end up asserting afterwards?

src/Optimization/hiopIterate.cpp
```diff
@@ -434,12 +434,12 @@ void hiopIterate::addLinearDampingTermToGrad_x(const double& mu, const double& k
   /*sxl->addLinearDampingTermToGrad(nlp->get_ixl(), nlp->get_ixu(), mu, kappa_d, grad_x);
     sxu->addLinearDampingTermToGrad(nlp->get_ixu(), nlp->get_ixl(), mu, kappa_d, grad_x); */
   //I'll do it in place, in one for loop, to be faster
```
cnpetra, line 436: I guess the code in lines 444-484, needs to be "ported" and use generic `hiopVector` interface ? @pelesh

tests/LinAlg/matrixTests.hpp
cnpetra: concurrently, as I would like to merge this PR asap

BUILD.sh
cnpetra: I'll put a note at the top of the script file

src/LinAlg/hiopMatrixMDS.hpp
cnpetra: oh, I see, I misread the code.

BUILD.sh
cnpetra: follow-up: I get this error when I ran BUILD.sh on  ascent

```
jsrun -n 1 -g 1 /ccsopen/home/petra1/hiop/build/src/Drivers/nlpMDS_ex4_raja.exe 400 100 1 -selfcheck
Error: It is only possible to use js commands within a job allocation unless CSM is running
10-16-2020 17:05:21:448 95719 main: Error initializing RM connection. Exiting.
```

src/Optimization/hiopIterate.cpp
pelesh: You are absolutely right @cnpetra. I didn't dare messing up with that part of the code. This is probably the primary reason why we need to use unified virtual memory when running on GPU. We also take performance hit there, because all vectors need to be copied to host memory for this computation to take place.

src/LinAlg/hiopMatrixMDS.hpp
cnpetra: see https://github.com/LLNL/hiop/issues/98

`addToSymDenseMatrixUpperTriangle` and `transAddToSymDenseMatrixUpperTriangle` were replaced with `addUpperTriangleToSymDenseMatrixUpperTriangle`

src/Optimization/hiopIterate.cpp
pelesh: This should be easy to fix, though.

src/LinAlg/hiopMatrixMDS.hpp
cnpetra: see a4022c8 and comments in #98

src/LinAlg/hiopMatrixSparseTriplet.cpp
cnpetra: see a4022c8 and comments in #98

src/LinAlg/hiopMatrixMDS.hpp
pelesh: The unit test for `addToSymDenseMatrixUpperTriangle` in symmetric sparse matrix class  should be kept to test the symmetric sparse matrix method `addUpperTriangleToSymDenseMatrixUpperTriangle`. It unit test was already doing that, because `addUpperTriangleToSymDenseMatrixUpperTriangle` was implemented as a wrapper to `addToSymDenseMatrixUpperTriangle` with row offset set to be equal to the column offset. We implemented test for the wrapped function instead of the wrapper, but simply replacing the names in the unit test would make it work again.

src/LinAlg/hiopMatrixMDS.hpp
pelesh: I suggest we make this change to `develop` branch after this PR is merged.

CC @cnpetra

tests/LinAlg/matrixTestsSymSparse.hpp
```diff
@@ -118,64 +118,64 @@ class MatrixTestsSymSparse : public TestBase
...
-    hiop::hiopMatrixDense& W,
-    hiop::hiopMatrixSparse& A, // sym sparse matrix
-    const int rank=0)
-  {
-    const local_ordinal_type N_loc = W.get_local_size_n();
-    const local_ordinal_type A_M = A.m();
-    const local_ordinal_type A_N_loc = A.n();
-    assert(W.m() == W.n());
-    assert(W.m() >= A.m());
-    assert(W.n() >= A.n());
```
pelesh: This test should be kept and modified to test `addUpperTriangleToSymDenseMatrixUpperTriangle`.

src/LinAlg/hiopMatrixMDS.hpp
cnpetra: will reactivate test for `addToSymDenseMatrixUpperTriangle` as test for `addUpperTriangleToSymDenseMatrixUpperTriangle`

pelesh: > will reactivate test for addToSymDenseMatrixUpperTriangle as test for addUpperTriangleToSymDenseMatrixUpperTriangle

I've got it. Will push shortly.

pelesh: Fixed in 3d15a97.

PR #91, opened 2020-10-03
redirect the classes. to use same code for IC

nychiang:
the structure changes a lot.
@cnpetra please feel free to comment and set a meeting with me for further discussion.  :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,14 @@ $> make
 $> make test
 $> make install
 ```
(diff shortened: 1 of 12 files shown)


cnpetra:
does not look that bad to me, at least it's the way I expected it. will pm with comments

pelesh:
Perhaps this should be merged to `develop` rather than `master`?

PR #104, opened 2020-11-17
fix #79: add new functions to vector kernel

### LINKED ISSUE #79: Counting negatives in Hxs should be implemented as a vector kernel?

nychiang:
attempt to fix #79, by adding new functions into vector kernel.

- [x] #128 must be merged before this can be merged


nychiang:
@ashermancinelli 
I know the I can pass some basic types of variables to device, e.g., 'int' and 'double'.
However, I am wondering if we should have a copy of 'tol' on device in [hiopVectorRajaPar.cpp#L1521](https://github.com/LLNL/hiop/blob/82a3dca1f995f191b29a416bea3a856606eb56b7/src/LinAlg/hiopVectorRajaPar.cpp#L1521)
and  [hiopVectorRajaPar.cpp#L1533](https://github.com/LLNL/hiop/blob/82a3dca1f995f191b29a416bea3a856606eb56b7/src/LinAlg/hiopVectorRajaPar.cpp#L1533), in order to have a better performance.

ashermancinelli:
The scalar `tol` is trivially copyable, and will be copied into RAJA lambda anyways. Unless I am mistaken, I don't think this should be an issue. @pelesh is this correct?

pelesh:
I think @ashermancinelli is correct -- lambda function will capture `tol` (i.e. copy it to the device) correctly. I don't think there is anything you need to do beyond that.

src/LinAlg/hiopVector.hpp
```diff
@@ -209,6 +209,10 @@ class hiopVector
   virtual void copyFromDev() = 0;
   virtual void copyToDev() const = 0;
   virtual void copyFromDev() const = 0;
+  
+  /// @brief get number of negative/zero values according to the given tolerance 'tol'
+  virtual long long numOfNegVal_w_tol(const double &tol) const = 0;
```
ashermancinelli, line 214: I think these should conform to the naming conventions we established in the developer readme - perhaps one of the following?

- `numOfNegVal`
- `numOfNegValWithTolerance`
- `countNegativeValues` (I think this is the clearest)

Although this does match the convention used in a couple other internal HiOp functions like `addConstant_w_patternSelect` - @pelesh @cnpetra how should we apply the style guide to these functions?

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -1514,5 +1514,30 @@ void hiopVectorRajaPar::copyFromDev() const
   resmgr.copy(data_host, data_dev_);
 }
 
+long long hiopVectorRajaPar::numOfNegVal_w_tol(const double &tol) const
+{  
+  double* data = data_dev_;
+  RAJA::ReduceSum<hiop_raja_reduce, int> sum(0);
```
ashermancinelli, line 1520: The function returns a `long long`, but the reduction variable uses an `int` - I think these should be consistent.

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -1514,5 +1514,30 @@ void hiopVectorRajaPar::copyFromDev() const
...
+  RAJA::forall<hiop_raja_exec>( RAJA::RangeSegment(0, n_local_),
+    RAJA_LAMBDA(RAJA::Index_type i)
+    {
+      sum += (data_[i]<-tol);
```
ashermancinelli, line 1524: It looks like the kernel is using the member variable for host data (`data_`) instead of using the local copy of pointer to device data (`data`, a local copy of `data_dev_`). I think the underscore suffix should be removed (`data_` -> `data`).

ashermancinelli (commented):
I think we should get some feedback on naming conventions and I think the pointers used in RAJA kernels need to be changed as I commented above, unless I misunderstand.

src/LinAlg/hiopVectorRajaPar.cpp
nychiang: Sorry, my fault. It's a typo. I did create the point in line [1519](https://github.com/LLNL/hiop/blob/82a3dca1f995f191b29a416bea3a856606eb56b7/src/LinAlg/hiopVectorRajaPar.cpp#L1519) but forget to use it.

nychiang:
@ashermancinelli @pelesh 
I see. The lambda function uses pass-by-value and hence it will copy the argument into device.
How about if we pass it by reference in the lambda function? 
Does RAJA need to convert the value to device in each iteration (note that this lambda function is applied in a for loop)?

pelesh:
> The lambda function uses pass-by-value and hence it will copy the argument into device.
> How about if we pass it by reference in the lambda function?
> Does RAJA need to convert the value to device in each iteration (note that this lambda function is applied in a for loop)?

I think RAJA requires all data to be passed by value to the lambda functions in `forall` method. Also, I am not sure how much it would buy you to pass a single `double` by reference. Instead of a 64-bit floating point value you would be copying 64-bit integer address.

src/LinAlg/hiopVector.hpp
cnpetra: how about 

`numElemsAbsLessThan` `numElemsLessThen`

cnpetra:
please let us know if it passes tests with GPU_ON on lassen or summit

nychiang:
> I think RAJA requires all data to be passed by value to the lambda functions in `forall` method. Also, I am not sure how much it would buy you to pass a single `double` by reference. Instead of a 64-bit floating point value you would be copying 64-bit integer address.
Oh, my question is not for this case. Just curious about the general behavior on the hybrid system. My guess is that using a reference to host memory may make things slower.

nychiang:
> please let us know if it passes tests with GPU_ON on lassen or summit
sure, will quickly correct these according to the feedbacks and test them on Lassen or Summit.

src/LinAlg/hiopVector.hpp
nychiang: I vote for `numOfElemsAbsLessThan` and `numOfElemsLessThan`.

src/LinAlg/hiopVectorPar.cpp
```diff
@@ -804,4 +804,19 @@ void hiopVectorPar::print(FILE* file, const char* msg/*=NULL*/, int max_elems/*=
   }
 }
 
+
+long long hiopVectorPar::numOfElemsLessThan(const double &val) const
```
cnpetra, line 808: hiopVectorPar is supposed to support distributed (MPI) vectors. We need to either implement full functionality for the MPI case (only need to do MPI_AllReduce) or assert(false) if the method(s) is called for a distributed vector (assert(n_local_==n_global_);

src/Optimization/hiopKKTLinSysMDS.cpp
```diff
@@ -266,20 +266,16 @@ namespace hiop
...
-	    break;
-	  }
-	}
+        // 'n_neg_eig' is the number of negative eigenvalues of the "dense" (reduced) KKT
+        //
+        // One can compute the number of negative eigenvalues of the whole MDS or XYcYd
+        // linear system using Haynsworth inertia additivity formula, namely,
+        // count the negative eigenvalues of the sparse Hessian block.
+        int n_neg_eig_Hxs  = Hxs_->numOfElemsLessThan(-1e-14);
+        int n_zero_eig_Hxs = Hxs_->numOfElemsLessThan(1e-14);
```
cnpetra, line 275: Shouldn't be 
`int n_zero_eig_Hxs = Hxs_->numOfElemsAbsLessThan(1e-14);`

nychiang: yes, you're right.

nychiang:
Address the following issues:
1. set RAJA and Umpire to OFF
2. issue #79
3. issue #136 
Not the new functions in vector kernels support distributed (MPI) vectors.

nychiang:
@cnpetra MPI support is added
@ashermancinelli  please check the corresponding RAJA implementations
Cheers,

ashermancinelli:
@nychiang Should this target/be rebased on #128? Some of these changes look the same as in interface-dev3 branch.

cnpetra:
> @nychiang Should this target/be rebased on #128? Some of these changes look the same as in interface-dev3 branch.

I discussed this with nai-yuan, it should be rebased on #128, but PR-ed against develop. Maybe this is why you see the same changes as in interface-dev3?

ashermancinelli:
> > @nychiang Should this target/be rebased on #128? Some of these changes look the same as in interface-dev3 branch.
> 
> I discussed this with nai-yuan, it should be rebased on #128, but PR-ed against develop (will be merged after #128). Maybe this is why you see the same changes as in interface-dev3?

Yes this seems right to me. If you don't mind I'll add this to the description. Thanks!

ashermancinelli (approved):
LGTM as long as this is merged after (1) #128 is merged and then (2) this is rebased on develop.

PR #105, opened 2020-11-17
Continuous Integration Fixes

ashermancinelli:
This diff resolves several issues we've encountered with continuous integration at ORNL and makes our naming conventions more consistent.


pelesh:
The commit fixes test failures on Ascent.

PR #106, opened 2020-11-18
HiOp Interface changes

cnpetra:
- interface changed to work with single double pointers


cnpetra:
@ashermancinelli

ashermancinelli:
Great thank you!

PR #108, opened 2020-11-18
update documentation for sparse linear algebra

nychiang:
@cnpetra Please read the sections I added. Especially the part about symmetric sparse matrix.
In fact, now Hessian in the interface requires upper triangular part, and both Jac/Hess for the test problems are sorted by row.
However, when I assemble the sparse KKT matrix in triplet form, I copy the entries from Hessian and jac to the **lower** triangular part. This allow me to keep the elements sorted by row too.
Not sure if I should mention this in the document, since it really doesn't matter how the internal code build KKT and pass it to a direct solver.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,6 +117,15 @@ or, if you are using C-shell, as
 $ setenv NVBLAS_CONFIG_FILE <hiop install dir>/lib/nvblas.conf
 ```
 
+## Existing issues
+Users are highly encouraged to report any issues they found from using Hiop.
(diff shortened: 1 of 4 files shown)


cnpetra:
wondering how messy would be to rebase on develop or interface-dev3 branches?

nychiang:
rebase this one to develop is simple. Most of the 22 commits come from #120, and only 1 commit is used for document.
I will push another commit soon to fix some typos in the documents.

nychiang:
only 4 text files are changed after I rebase to `interface-dev3`

PR #122, opened 2020-11-23
Adjusting path for remote mirror

### LINKED ISSUE #119: Repo mirror switch ssh to https
Because of firewall and policy constraints on port 22, gitlab.pnnl.gov is using 2222 for ssh git commands. The current Actions Market place tool used to sync is forcing to port 22.

Plan:
Fork the current market place addon and change the port to 2222 and try again.

I'll take this on, just making note.

poke @ashermancinelli @pelesh

GitBytes:
This should fix #119

### DIFF
--- a/.github/workflows/push_mirror.yml
+++ b/.github/workflows/push_mirror.yml
@@ -6,8 +6,9 @@ jobs:
   to_gitlab:
     runs-on: ubuntu-18.04
     steps:
-    - uses: actions/checkout@v1
-    - uses: pixta-dev/repository-mirroring-action@v1
-      with:
-        target_repo_url: ssh://git@gitlab.pnnl.gov:2222/exasgd/solvers/hiop-mirror.git
-        ssh_private_key: ${{ secrets.GITLAB_SSH_PRIVATE_KEY }}
+      - uses: actions/checkout@v1
+      - uses: spyoungtech/mirror-action@master
+        with:
+          REMOTE: 'https://gitlab.pnnl.gov/exasgd/solvers/hiop-mirror.git'
+          GIT_USERNAME: githubactions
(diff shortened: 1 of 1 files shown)


GitBytes:
@ashermancinelli I think the issue is I was just missing the prefix. I tested locally and it should do the right thing. We will still need to add an ssh key (gitlab token) to the secret values of the pipeline to fully work. Should we get together on a call to put that in?

GitBytes:
@ashermancinelli I am going to make a change. I am looking at https://github.com/spyoungtech/mirror-action which is actually a bit closer to what we want I think. This way we don't have to wrangle ssh keys like I was thinking originally. If that is ok, I'll reform this to use the HTTPS variant instead of ssh. This will pay off in the long term as well as the Lab is apprehensive about using ssh for git related tasks.

ashermancinelli:
> @ashermancinelli I am going to make a change. I am looking at https://github.com/spyoungtech/mirror-action which is actually a bit closer to what we want I think. This way we don't have to wrangle ssh keys like I was thinking originally. If that is ok, I'll reform this to use the HTTPS variant instead of ssh. This will pay off in the long term as well as the Lab is apprehensive about using ssh for git related tasks.

This sounds great to me!

GitBytes:
Example passing push stage: https://github.com/GitBytes/hiop/runs/1445441686?check_suite_focus=true
This is going to my remote fork on gitlab.pnnl.gov.
I added a couple more parameters so it can be used in other cases if desired as a drop in job step.

GitBytes:
@pelesh FYI I think my part here is complete. Let me know if there are any issues.

ashermancinelli:
Secrets should be in place, I believe this is ready for merge @cnpetra 
CC @pelesh @nychiang

Issue #121, opened 2020-11-23
Flawed logic in cmake for combinations of GPU, RAJA, UMPIRE, and MAGMA flags


nychiang:
Tried to use different options to compile the code and noticed some error when combining GPU, RAGA, UMPIRE and MAGMA.
For example:
1.  Switch off GPU and RAJA --- fail
    `cmake -DHIOP_USE_GPU=OFF -DHIOP_USE_RAJA=OFF -DHIOP_SPARSE=OFF ..`
error message:
/ccs/home/chiang7/project/hiop/src/LinAlg/hiopVectorRajaPar.cpp:73:10: fatal error: RAJA/RAJA.hpp: No such file or directory
   73 | #include <RAJA/RAJA.hpp>

2.	Switch off GPU --- fail 
    `cmake -DHIOP_SPARSE=OFF -DHIOP_USE_GPU=OFF ..`
cmake configuration error:
CMake Error: Error required internal CMake variable not set, cmake may not be built correctly.
Missing variable is:
CMAKE_CUDA_COMPILE_WHOLE_COMPILATION

If we ignore this error and continue to build hiop, then I will get errors showing MAGMA is not referenced.

Therefore, if I want to switch off MAGMA, I need to switch off RAJA, UMPIRE and GPU:
That is, using command "cmake -DHIOP_USE_GPU=OFF -DHIOP_USE_RAJA=OFF -DHIOP_USE_UMPIRE=OFF -DHIOP_SPARSE=OFF -DHIOP_USE_MAGMA=OFF .."

nychiang:
You can also reproduce the similar error in branch `develop`.

pelesh:
@nychiang thank you for reporting this! I think these should not be difficult to fix. Do you want to first complete #120 merge before we tackle this?

CC @cnpetra @ashermancinelli

ashermancinelli:
> @nychiang thank you for reporting this! I think these should not be difficult to fix. Do you want to first complete #120 merge before we tackle this?
> 
> CC @cnpetra @ashermancinelli

Must this be addressed for #120 to pass CI? If so, I don't mind spending some time to get this working before merging.

pelesh:
@ashermancinelli, I suggest we keep fix for this issue separate from #120. These errors occur only when RAJA is on, but GPU support is off. I think CMake will need another overhaul beyond fix for this issue. @cnpetra, what do you think?

cnpetra:
[...]

PR #124, opened 2020-11-24
Fix in linear algebra module CMake config

pelesh:
This is a quick fix to the logic in the CMake file for linear algebra module. I added some comments explaining what each CMake stanza does. The modifications passes CI tests on newell and marianas.

Note that at this stage we assume GPU == CUDA and Magma is always enabled when GPU is enabled. This needs to be improved eventually, but it is outside the scope of this PR.

### DIFF
--- a/src/LinAlg/CMakeLists.txt
+++ b/src/LinAlg/CMakeLists.txt
@@ -1,3 +1,5 @@
+
+# Set linear algebra common source files
 set(hiopLinAlg_SRC
   hiopVectorPar.cpp
   hiopVectorIntSeq.cpp
@@ -10,6 +12,7 @@ set(hiopLinAlg_SRC
   hiopMatrixComplexSparseTriplet.cpp
 )
 
+# Add interfaces for sparse linear solvers when enabled
 if(HIOP_SPARSE)
     if(HIOP_USE_COINHSL)
       set(hiopLinAlg_SRC ${hiopLinAlg_SRC} hiopLinSolverIndefSparseMA57.cpp)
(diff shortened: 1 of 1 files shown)


pelesh:
Passed CI tests on ascent, as well.

ashermancinelli (approved):
lgtm

src/LinAlg/CMakeLists.txt
```diff
@@ -19,16 +22,7 @@ if(HIOP_SPARSE)
...
-            hiopMatrixRajaDense.cpp
-            hiopMatrixRajaSparseTriplet.cpp)
-    set_source_files_properties(${f} PROPERTIES LANGUAGE CUDA)
-  endforeach()
-endif()
-
+# Add RAJA/Umpire sources when enabled
 if(HIOP_USE_UMPIRE)
   set(hiopLinAlg_SRC ${hiopLinAlg_SRC}
     hiopVectorRajaPar.cpp
```
nychiang, line 28: Since now Raja is required whenever HIOP_USE_UMPIRE=ON, should we use change the if condition to
if(HIOP_USE_UMPIRE AND HIOP_USE_RAJA). I think this will also address the issue in #121

pelesh: @nychiang, RAJA has been implicitly required to be on whenever Umpire was on and we failed to make proper enforcements in the CMake code. I think the simpler solution would be to use only one of `HIOP_USE_UMPIRE` or `HIOP_USE_RAJA` and once it is set to `ON` enable both. I can't think of a use case where you would use one without the other in HiOp. I would suggest refactoring CMake and addressing #121 shortly after this and #120 PRs are merged.

nychiang (approved):

PR #125, opened 2020-11-25
Removing RAJA from github actions build

ashermancinelli:
RAJA Cmake option defaults to ON which will fail in github actions build/tests. This sets RAJA/Umpire to OFF for actions build. [Here is a link to passing build in my fork.](https://github.com/ashermancinelli/hiop/runs/1455141500?check_suite_focus=true)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
           cc: gcc
           cxx: g++
           build-type: Debug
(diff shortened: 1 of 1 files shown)


cnpetra:
@ashermancinelli : please do not assume RAJA and UMPIRE are ON by default. They will most likely be switched back to default of OFF in the release (since this will not disrupt existing users)

ashermancinelli:
> please do not assume RAJA and UMPIRE are ON by default

In this PR I'm manually setting them to OFF, which will work if the defaults change in the future. No assumptions here :smile:

cnpetra (approved):

PR #135, opened 2020-11-30
Updating MPI configurations for testing

CameronRutherford:
Previously HiOp was not using mpi commands to run tests by default. In this PR, I have updated this in the CMakeLists.txt, along with adding some options that are relevant to CI on Newell. 

Removed `bash -c` as a part of some of the testing commands as they are only necessary for some situations. When necessary, this is placed after the mpi command.

cc @ashermancinelli


ashermancinelli (approved):
LGTM

ashermancinelli (changes requested):
See comment below. Sorry for weird formatting, I messed up in the GUI.

CMakeLists.txt
```diff
@@ -345,60 +345,62 @@ set(STRIP_TABLE_CMD "awk '/Problem Summary/ { while (!match($0, /termination/)){
...
   set(MPICMD "jsrun")
   if(HIOP_USE_GPU)
-    set(MPICMD "${MPICMD} -g 1")
-    set(RUNCMD "${RUNCMD} -g 1")
+    set(MPICMD ${MPICMD} "-g" "1")
+    set(RUNCMD ${RUNCMD} "-g" "1")
   endif()
 else()
-  set(RUNCMD "") # No special command is needed to run this program
+  set(RUNCMD "mpirun" "-n" "1")
```
ashermancinelli, line 355: I think you'll have to check if `HIOP_USE_MPI` is on before using mpirun.

ashermancinelli (changes requested):

CMakeLists.txt
CameronRutherford: Just added a commit to address this. Should be fixed now

ashermancinelli (approved):
lgtm

CMakeLists.txt
```diff
@@ -345,60 +345,66 @@ set(STRIP_TABLE_CMD "awk '/Problem Summary/ { while (!match($0, /termination/)){
...
+    set(MPICMD ${MPICMD} "-g" "1")
+    set(RUNCMD ${RUNCMD} "-g" "1")
   endif()
 else()
-  set(RUNCMD "") # No special command is needed to run this program
   set(MPICMD "mpirun")
+  if(HIOP_USE_MPI)
+    set(RUNCMD "mpirun" "-n" "1")
+    set(RUNCMD ${RUNCMD} "-mca" "oob" "tcp")
+    set(MPICMD ${MPICMD} "-mca" "oob" "tcp")
```
ashermancinelli, line 359: Are these specific to newell? Can these cause problems elsewhere? I don't think the line
```
set(RUNCMD ${RUNCMD} "-mca" "oob" "tcp")
```
makes much sense either.

Perhaps we should allow the user to pass an argument or set an environment variable to add arguments to runcmd/mpicmd.

ashermancinelli (changes requested):

CMakeLists.txt
CameronRutherford: These are probably only specific to PNNL clusters, and so I agree that we should probably remove these in favour of the optional input you suggested. Since the options are just suppressing warnings, would it suffice to just remove the options? Is there a particular use case that would require optional options being passed to mpi/jsrun?

ashermancinelli: I'd add a cmake line like:
```cmake
set(HIOP_EXTRA_MPI_FLAGS "" CACHE STRING "Extra arguments to mpiexec when running tests")
```
and then in the build script under the newell-specific section, we can add
```shell
extra_cmake_args="$extra_cmake_args -DHIOP_EXTRA_MPI_FLAGS=\"-mca oob tcp\""
```

CameronRutherford: I have implemented this suggestion in my most recent commit. The only thing that I had to change was passing the `extra_cmake_args` in a slightly different way:
```shell
extra_cmake_args="$extra_cmake_args -DHIOP_EXTRA_MPI_FLAGS=-mca;oob;tcp;"
```

ashermancinelli (approved):
Passing CI everywhere and lgtm

cnpetra:
I assume this waits for @pelesh

pelesh (approved):
Tests pass on all pipelines, fix removes explicit dependence on bash shell.

PR #140, opened 2020-12-03
use NDEBUG instead of DEBUG in preprocessor directives

nychiang:
use `#ifndef NDEBUG` instead of `#ifdef DEBUG`

### DIFF
--- a/src/Drivers/IpoptAdapter.hpp
+++ b/src/Drivers/IpoptAdapter.hpp
@@ -572,7 +572,7 @@ public:
 	  nnzit++;
 	}
       }
-#ifdef DEBUG
+#ifndef NDEBUG
       //nnzit += nx_dense*nx_dense;
       assert(nnzit==nele_hess);
 #endif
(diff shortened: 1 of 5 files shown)


ashermancinelli:
Looks like github ci found an error:
```
src/LinAlg/hiopVectorPar.cpp:221:2: error: invalid preprocessing directive #infdef; did you mean #ifdef?
```

nychiang:
@ashermancinelli Yes, I noticed the CI result and I have already fixed it in my next commit. In the meantime I found another problem in CMake, which I will file a new issue soon.

ashermancinelli:
Just to clarify, when are we using `NDEBUG` vs `HIOP_DEEPCHECKS`?

cnpetra:
> Just to clarify, when are we using `NDEBUG` vs `HIOP_DEEPCHECKS`?

it' about a fixing an inconsistency in hiop re: assert and DEBUG. `assert` is based on NDEBUG, while hiop in many cases uses `assert` inside `ifdef DEBUG`. This inconsistency can result in compilation errors or asserts being disabled in debug mode. The latter happens for example when NDEBUG not defined and also DEBUG not defined, which is a valid "debug" configuration as far as I remember.

cnpetra:
@nychiang is this ready to review / merge ?

nychiang:
@cnpetra  yes, it is ready. just need your quick review and approval

Issue #133, opened 2020-11-27
Unit test method reuse


ashermancinelli:
Many of the methods in the testing classes were simply copy/pasted into others as a shortcut, for example [RAJA verify answer](https://github.com/LLNL/hiop/blob/develop/tests/LinAlg/matrixTestsRajaSparseTriplet.cpp#L157) and [CPU-bound verify answer](https://github.com/LLNL/hiop/blob/develop/tests/LinAlg/matrixTestsSparseTriplet.cpp#L253). This shortcut will hold us back as we expand our testing framework.

ashermancinelli:
@CameronRutherford is reviewing UT framework and will post his evaluation soon. From there we will act on the design changes.

CameronRutherford:
When the unit test methods are currently re-used/duplicated, there are a few ways in which this happens:

1. The function is blatantly duplicated, and can easily be moved to a single implementation location.
2. The function is almost identical, but static casts the object first to a particular linear algebra object.
3. (2) occurs, but there is a call to `copyFromDev` before data is accessed.
4. (2) occurs, along with a call to `getLocalElement` instead of `getLocalData`.
5. The function is unique to the test, and should remain as such.

For cases 1-4 (which comprises as nearly __all__ of our existing unit tests), the duplicated functions can be removed, and instead, we can implement a singular test function, that takes advantage of the abstract underlying base class of our linear algebra objects. This singular implementation can be defined in the associated base class, allowing child classes to call the same function. We can also take advantage of non-GPU objects having no-ops when the `copyFromDev` functions are called, allowing us to re-use the same tests for the RAJA and non-RAJA implementations. 

There are only a handful of functions that fall under 5, and so these will most likely remain as is and cannot be consolidated.

It should be noted that sparse and dense matrices should still be treated differently, but keeping this in mind should still allow us to remove many of the duplicate methods. Given this, we should be able to move nearly all duplicated methods into the associated base testing classes by taking full advantage of the underlying base classes of our linear algebra objects.

In addition to these changes, @pelesh also mentioned that we should look an cleaning up our test drivers themselves. Per his suggestion, we could implement something along the lines of the following:

```c++
template <class TestT>
static int run_tests(std::string mem_space)
{
  int fail;
  options.SetStringValue("mem_space", mem_space);
  hiop::LinearAlgebraFactory::set_mem_space(options.GetString("mem_space"));
  std::string mem_space = hiop::LinearAlgebraFactory::get_mem_space();
  TestT test;
[...]

PR #141, opened 2020-12-03
Vector Tests Consolidation

CameronRutherford:
This is the first pass attempt at consolidating the unit testing framework per #133.

@ashermancinelli @pelesh please review, as I may have made some non-kosher changes.

### DIFF
--- a/tests/LinAlg/vectorTests.hpp
+++ b/tests/LinAlg/vectorTests.hpp
@@ -1542,14 +1542,56 @@ public:
     return x->local_data_host_const()[i];
   }
 
-protected:
-  // Interface to methods specific to vector implementation
-  virtual local_ordinal_type getLocalSize(const hiop::hiopVector* x) = 0;
-  virtual real_type* getLocalData(hiop::hiopVector* x) = 0;
-  virtual int verifyAnswer(hiop::hiopVector* x, real_type answer) = 0;
+  /// Returns pointer to local vector data
+  local_ordinal_type getLocalSize(const hiop::hiopVector* x)
+  {
+    return static_cast<local_ordinal_type>(x->get_local_size());
+  }
+
+  /// Checks if _local_ vector elements are set to `answer`.
+  int verifyAnswer(hiop::hiopVector* x, real_type answer)
+  {
+    x->copyFromDev();
+    const local_ordinal_type N = getLocalSize(x);
+    auto* xdata = static_cast<const real_type*>(getLocalData(x));
+    
+    int local_fail = 0;
+
+    for(local_ordinal_type i = 0; i < N; ++i)
+    {
+      if(!isEqual(xdata[i], answer))
+      {
+        ++local_fail;
+      }
+    }
+
+    return local_fail;
+  }
+
   virtual int verifyAnswer(
       hiop::hiopVector* x,
-      std::function<real_type(local_ordinal_type)> expect) = 0;
+      std::function<real_type(local_ordinal_type)> expect)
+  {
+    x->copyFromDev();
+    const local_ordinal_type N = getLocalSize(x);
+    auto* xdata = static_cast<const real_type*>(getLocalData(x));
+    
+    int local_fail = 0;
+
+    for(local_ordinal_type i = 0; i < N; ++i)
+    {
+      if(!isEqual(xdata[i], expect(i)))
+      {
+        ++local_fail;
+      }
+    }
+
+    return local_fail;
+  }
+
+protected:
+  // Interface to methods specific to vector implementation
+  virtual real_type* getLocalData(hiop::hiopVector* x) = 0;
   virtual bool reduceReturn(int failures, hiop::hiopVector* x) = 0;
   virtual real_type* createLocalBuffer(local_ordinal_type N, real_type val) = 0;
   virtual void deleteLocalBuffer(real_type* buffer) = 0;
--- a/tests/LinAlg/vectorTests.hpp
+++ b/tests/LinAlg/vectorTests.hpp
@@ -1542,6 +1542,20 @@ public:
     return static_cast<local_ordinal_type>(x->get_local_size());
   }
 
+  /// If test fails on any rank set fail flag on all ranks
+  bool reduceReturn(int failures, hiop::hiopVector* x)
+  {
+    int fail = 0;
+
+#ifdef HIOP_USE_MPI
+    MPI_Allreduce(&failures, &fail, 1, MPI_INT, MPI_SUM, getMPIComm(x));
+#else
+    fail = failures;
+#endif
+
+    return (fail != 0);
+  }
+
   /// Checks if _local_ vector elements are set to `answer`.
   int verifyAnswer(hiop::hiopVector* x, real_type answer)
   {
@@ -1561,7 +1575,7 @@ public:
     return local_fail;
   }
 
-  virtual int verifyAnswer(
+  int verifyAnswer(
       hiop::hiopVector* x,
       std::function<real_type(local_ordinal_type)> expect)
   {
@@ -1585,9 +1599,11 @@ protected:
   // Interface to methods specific to vector implementation
   virtual real_type* getLocalData(hiop::hiopVector* x) = 0;
   virtual void setLocalElement(hiop::hiopVector* x, local_ordinal_type i, real_type val) = 0;
-  virtual bool reduceReturn(int failures, hiop::hiopVector* x) = 0;
   virtual real_type* createLocalBuffer(local_ordinal_type N, real_type val) = 0;
   virtual void deleteLocalBuffer(real_type* buffer) = 0;
+#ifdef HIOP_USE_MPI
+  virtual MPI_Comm getMPIComm(hiop::hiopVector* x) = 0;
+#endif
 };
 
 }} // namespace hiop::tests
--- a/tests/LinAlg/vectorTests.hpp
+++ b/tests/LinAlg/vectorTests.hpp
@@ -1531,9 +1531,9 @@ public:
   }
 
   /// Returns element _i_ of vector _x_.
-  real_type getLocalElement(hiop::hiopVector* x, local_ordinal_type i)
+  real_type getLocalElement(const hiop::hiopVector* x, local_ordinal_type i)
   {
-    return getLocalData(x)[i];
+    return getLocalDataConst(x)[i];
   }
 
   /// Returns pointer to local vector data
@@ -1598,6 +1598,7 @@ public:
 protected:
   // Interface to methods specific to vector implementation
   virtual real_type* getLocalData(hiop::hiopVector* x) = 0;
+  virtual const real_type* getLocalDataConst(const hiop::hiopVector* x) = 0;
   virtual void setLocalElement(hiop::hiopVector* x, local_ordinal_type i, real_type val) = 0;
   virtual real_type* createLocalBuffer(local_ordinal_type N, real_type val) = 0;
   virtual void deleteLocalBuffer(real_type* buffer) = 0;
--- a/tests/LinAlg/vectorTests.hpp
+++ b/tests/LinAlg/vectorTests.hpp
@@ -1530,14 +1530,6 @@ public:
     return reduceReturn(fail, &x);
   }
 
-  /// Method to set vector _x_ element _i_ to _value_.
-  void setLocalElement(hiop::hiopVector* x, local_ordinal_type i, real_type val)
-  {
-    x->copyFromDev();
-    real_type *xdat = x->local_data_host();
-    xdat[i] = val;
-    x->copyToDev();
-  }
   /// Returns element _i_ of vector _x_.
   real_type getLocalElement(hiop::hiopVector* x, local_ordinal_type i)
   {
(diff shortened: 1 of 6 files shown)


tests/LinAlg/vectorTest.hpp
```diff
@@ -1527,19 +1537,151 @@ class VectorTests : public TestBase
...
+      return false;
+    }
+#else
+    fail = failures;
+#endif
+    return (fail != 0);
+  }
+
+  /// Wrap new command
+  real_type* createLocalBuffer(local_ordinal_type N, real_type val)
```
CameronRutherford, line 1635: There might be a better way to do this, but this is the easiest solution that I could come up with that allows the test to be implementation agnostic. Also applies to `deleteLocalBuffer` below this function.

tests/LinAlg/vectorTest.hpp
```diff
@@ -1527,19 +1537,151 @@ class VectorTests : public TestBase
...
+      {
+        ++local_fail;
+      }
+    }
+
+    return local_fail;
+  }
+
+  /// If test fails on any rank set fail flag on all ranks
+  bool reduceReturn(int failures, hiop::hiopVector* x)
```
CameronRutherford, line 1610: This function is why I had to include `hiopVectorPar` and `hiopVectorRajaPar`. Since `get_mpi_comm()` is not a part of the `hiopVector` abstract interface, this seemed like the most reasonable solution.

tests/LinAlg/vectorTest.hpp
```diff
@@ -313,6 +320,9 @@ class VectorTests : public TestBase
     real_type* todata = getLocalData(&to);
     v.copyTo(todata);
 
+    // Since we have copied into to's data_host, we need to copy to Device
```
CameronRutherford, line 323: Is this extra copy desirable? The test was failing before I added this since the `todata` was not being copied onto the device, and the version of `verifyAnswer` that I implemented calls `copyFromDev`.

The other solution that is possible is just calling `verifyAnswer` on `todata`, but this would require creating an extra `verifyAnswer` implementation for the data type double*. This is possible, but I did not think that it would be a necessary change given the extra copy is a simpler fix.

tests/LinAlg/vectorTest.hpp
CameronRutherford: This is also why I needed the additional include to umpire at the top of the file

tests/LinAlg/vectorTest.hpp
ashermancinelli: I'm a little worried about implementation-specific stuff creeping in. I think we should find a way around this and keep hiopVectorPar and hiopVectorRajaPar details out of tests/LinAlg/vectorTest.hpp at all costs

CameronRutherford: I agree with this, but I did not want to make changes to the abstract interface without double-checking.

tests/LinAlg/vectorTest.hpp
```diff
@@ -85,11 +92,11 @@ namespace hiop { namespace tests {
  * and return `false` otherwise.
  *
  */
-class VectorTests : public TestBase
+class VectorTest : public TestBase
 {
 public:
-  VectorTests(){}
-  virtual ~VectorTests(){}
+  VectorTest(){}
```
ashermancinelli, line 98: What's the rationale for removing the plural here?

tests/LinAlg/vectorTest.hpp
ashermancinelli: @cnpetra Are you opposed to adding `get_mpi_comm` to hiopVector interface if `HIOP_USE_MPI=ON`?

tests/LinAlg/vectorTest.hpp
```diff
@@ -1527,19 +1537,151 @@ class VectorTests : public TestBase
...
+    real_type *xdat = x->local_data_host();
+    xdat[i] = val;
+    x->copyToDev();
+  }
+  /// Returns element _i_ of vector _x_.
+  real_type getLocalElement(const hiop::hiopVector* x, local_ordinal_type i)
+  {
+    x->copyFromDev();
+    return x->local_data_host_const()[i];
+  }
```
pelesh, line 1553: I suggest keeping these in a implementations specific file. While this might work for the test cases we have now, it is a short-sighted solution.

tests/LinAlg/vectorTest.hpp
```diff
@@ -1527,19 +1537,151 @@ class VectorTests : public TestBase
...
+  {
+    x->copyFromDev();
+    const local_ordinal_type N = getLocalSize(x);
+    auto* xdata = static_cast<const real_type*>(getLocalData(x));
+    
+    int local_fail = 0;
+
+    for(local_ordinal_type i = 0; i < N; ++i)
+    {
+      if(!isEqual(xdata[i], answer))
```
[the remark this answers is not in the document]
pelesh: Not with `getElement` as is implemented now. Probably the best way to implement `verifyAnswer` method(s) is indeed to copy data to the host mirror and then perform all comparisons there. 

Furthermore, method `verifyAnswer` is supposed to be implementation specific and should not be part of `vectorTests.hpp`. Since its implementation is simple and straightforward, I am not sure if it is worthwhile to put too much effort in making it general.

pelesh (changes requested):
There is a good reason why some of the helper methods are specific to vector implementation. IMHO, `get/setLocalElement` and `verifyAnswer` methods should remain implementations specific. Same goes for `getLocalData`.

I think it is possible to make _some_ of these methods implementation agnostic, but not all. I suggest having a mini-hackathon event to resolve current issues.

tests/LinAlg/vectorTest.hpp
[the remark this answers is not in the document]
ashermancinelli: sgtm 👍

pelesh:
I converted this to draft, since this PR needs more discussion (and preferably a hackathon event) before it is ready to merge.

tests/LinAlg/vectorTest.hpp
CameronRutherford: This was done in debugging, and I am happy to make it plural again if desired.

CameronRutherford:
@pelesh @ashermancinelli I have updated the changes in this PR to be in line with what was discussed during mini-hackathon. Please re-review, so that hopefully we can move onto the matrix tests. [This](https://github.com/LLNL/hiop/pull/141#discussion_r535664817) is the only change that I noticed that we did not explicitly review, so perhaps it is worth re-visiting.

tests/LinAlg/vectorTests.hpp
```diff
@@ -1527,19 +1530,80 @@ class VectorTests : public TestBase
     return reduceReturn(fail, &x);
   }
 
+  /// Returns element _i_ of vector _x_.
+  real_type getLocalElement(hiop::hiopVector* x, local_ordinal_type i)
```
CameronRutherford, line 1534: I ended up removing the const specifier for `hiop::hiopVector* x` in attempting to move this function to `vectorTests.hpp`. Since this is mainly a wrapper for getLocalData and is only used in specific situations, this seems fine to me. If not, I can easily move this and re-implement for each vector test.

pelesh (changes requested):
- Make vector test module builds cleanly and there are no warnings when compiled with `-Wconversion`.
- Address comments in the code.
- Consider adding `getLocalDataConst` method.

ashermancinelli (commented):
I've +1'ed some of slaven's comments, and I don't see any additional suggestions at the moment.

tests/testVector.cpp
```diff
@@ -72,6 +72,107 @@
 #include <hiopVectorIntRaja.hpp>
 #endif
 
+template <typename T>
+int vectorTests(T test, const char* mem_space, int rank, int numRanks, MPI_Comm comm)
+{
```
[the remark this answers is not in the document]
CameronRutherford:
```suggestion
template <typename T>
int vectorTests(T test, const char* mem_space, int rank, int numRanks, MPI_Comm comm)
{
  int rank=0;

#ifdef HIOP_USE_MPI
  int err;
  MPI_Comm comm = MPI_COMM_WORLD;
  err  = MPI_Comm_rank(comm, &rank);     
  assert(MPI_SUCCESS == err);
#endif
```
Are you suggesting something like this, where we have the MPI initialisation happen in main, but rank calculation be determined within the test? I would think that just passing the rank as a parameter would suffice, however, this change can also work.

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -325,15 +325,12 @@ void hiopVectorRajaPar::copyFromStarting(int start_index_in_this, const double*
...
   assert(n_local_ == n_ && "are you sure you want to call this?");
 #endif
   const hiopVectorRajaPar& v = dynamic_cast<const hiopVectorRajaPar&>(vec);
-  assert(start_index + v.n_local_ <= n_local_);
-  
-  auto& rm = umpire::ResourceManager::getInstance();
-  double* vv = const_cast<double*>(v.data_dev_); // scary: 
-  rm.copy(this->data_dev_ + start_index, vv, v.n_local_*sizeof(double));
+  assert(start_index + v.get_local_size() <= n_local_);
+  this->copyFromStarting(start_index, v.local_data_const(), v.get_local_size());
```
CameronRutherford, line 333: This may not be a kosher change, but I noticed it as a way to consolidate implementation. This also allows this function to take advantage of the 0 length array bounds checking in the pointer implementation of `copyFromStarting`. Might be irrelevant for this PR.

src/LinAlg/hiopLinAlgFactory.cpp
```diff
@@ -204,22 +204,36 @@ hiopMatrixSparse* LinearAlgebraFactory::createMatrixSymSparse(int size, int nnz)
 /**
  * @brief Static method to create a raw C array
  */
-double* LinearAlgebraFactory::createRawArray(int n)
```
CameronRutherford, line 207: I decided to add a default value parameter to this method in order to allow for a default value for the memory to be initialised too. I support this as a change as it prevents potential unwanted behaviour resulting from compiler default initialisation values. This being said, this does result in an extra memory transfer occurring in order to incorporate the functionality.

Since this functionality is only being added in order to better accommodate testing, I can see how it may be unnecessary as it is worse for performance. Without this change, I do not see a way to remove the createLocalBuffer and deleteLocalBuffer method implementations for the vector testing suite, so I would need to re-impliment the changes.

I also would have used an optional function parameter if we were able to use c++17 an onwards...

CameronRutherford (commented):
@pelesh @ashermancinelli please re-review. Thank you for your previous feedback!

src/LinAlg/hiopLinAlgFactory.cpp
ashermancinelli: > this does result in an extra memory transfer occurring in order to incorporate the functionality.

I don't think it does from what I can see. I would only add a default value for `value` like:
```cpp
double* LinearAlgebraFactory::createRawArray(int n, int value=0)
```
so we don't have to change existing code.

tests/testVector.cpp
```diff
@@ -73,10 +73,12 @@
 #endif
 
 template <typename T>
-int vectorTests(T test, const char* mem_space, int rank, int numRanks, MPI_Comm comm)
+int vectorTests(const char* mem_space, int rank, int numRanks, MPI_Comm comm)
```
ashermancinelli, line 76: May be overkill, but I think we should constrain this type:
```cpp
template <class T,
          typename std::enable_if<
            std::is_base_of<hiop::tests::VectorTests, T>::value,
            bool>::type = true>
int vectorTests(const char* mem_space, int rank, int numRanks, MPI_Comm comm)
```
so you can only call this with the correct test types. This should be malformed if you call with `hiop::tests::VectorTestsInt` for example.

ashermancinelli (changes requested):

src/LinAlg/hiopLinAlgFactory.cpp
CameronRutherford: There are two buffers that are created in the modified implementation [here](https://github.com/LLNL/hiop/blob/459d5493bbfc3b4be0bfbdf1512adc7e3141a2ed/src/LinAlg/hiopLinAlgFactory.cpp#L225), along with a call to `resmgr.copy()`. In the previous implementation, only one allocation is made, and no copies are performed.

pelesh: Since this PR is for unit tests refactoring, I think these changes are out of scope here. There should be a separate issue/PR if you think we need to make these changes to the source code. Same applies for the rest of the changes in `src`.

CameronRutherford: Moved changes in `src` to #157

CameronRutherford:
@pelesh @ashermancinelli this has been rebased w.r.t. develop, and I have moved out-of-scope changes to another PR. I think this may finally be ready to be merged, pending resolution of remaining comments [here](https://github.com/LLNL/hiop/pull/141#discussion_r538921709) and [here](https://github.com/LLNL/hiop/pull/141#discussion_r535649134).

ashermancinelli (approved):
Great work @CameronRutherford and thanks @pelesh for jumping into this to move us along!

pelesh (approved):
This looks good and I suggest we merge it directly to `develop` branch. Make sure you squash all commits before merging.

@cnpetra, this PR touches only testing code and could be merged without interfering with your high priority PRs.

cnpetra (approved):

PR #144, opened 2020-12-04
Fix bugs in the full KKT linear system

nychiang:
Fix bug when passing the full KKT matrix to STRUMPACK.
After this fix, HiOp with STRUMPACK passes the test problem EX6 with 5000 variables.

Note that the default ordering strategy in Metis is used in STRUMPACK, and STRUMPACK prints some mild warnings about it.
Will try different ordering schemes and/or discuss this issue with the STRUMPACK team before fixing it (probably in another PR).

### DIFF
--- a/src/LinAlg/hiopLinSolverSparseSTRUMPACK.cpp
+++ b/src/LinAlg/hiopLinSolverSparseSTRUMPACK.cpp
@@ -231,6 +231,7 @@ namespace hiop
     * initialize strumpack parameters
     */
 //    spss.options().set_matching(MatchingJob::NONE);
+    spss.options().enable_METIS_NodeNDP();
     spss.options().disable_gpu();
     spss.options().set_verbose(false);
 
--- a/src/LinAlg/hiopMatrixSparseTriplet.cpp
+++ b/src/LinAlg/hiopMatrixSparseTriplet.cpp
@@ -925,13 +925,13 @@ void hiopMatrixSparseTriplet::convertToCSR(int &csr_nnz, int **csr_kRowPtr_in, i
   double *csr_kVal = *csr_kVal_in;
   int *csr_jCol = *csr_jCol_in;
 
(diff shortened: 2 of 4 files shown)


src/LinAlg/hiopMatrixSparseTriplet.cpp
```diff
@@ -925,13 +925,13 @@ void hiopMatrixSparseTriplet::convertToCSR(int &csr_nnz, int **csr_kRowPtr_in, i
...
 
-  int *index_covert_extra_Diag2CSR_temp = new int[n];
+  int *index_covert_extra_Diag2CSR_temp = new int[n]{};
   int *nnz_each_row_tmp = new int[n]{};
 
   // set correct col index and value
   {
-    *index_covert_CSR2Triplet_in = new int[csr_nnz];
-    *index_covert_extra_Diag2CSR_in = new int[n];
+    *index_covert_CSR2Triplet_in = new int[csr_nnz]{};
```
cnpetra, line 933: I do not think the extra initialization to 0 (`{}`) is needed as you are rewriting all the elements in a loop later on...

nychiang: they are not necessary. Just try to keep consistence with the other part, e.g., line 929 in your reference above.

cnpetra: unnecessary extra work is being done, please remove

src/LinAlg/hiopLinSolverSparseSTRUMPACK.cpp
```diff
@@ -231,6 +231,7 @@ namespace hiop
     * initialize strumpack parameters
     */
 //    spss.options().set_matching(MatchingJob::NONE);
+    spss.options().enable_METIS_NodeNDP();
```
nychiang, line 234: suggested by strumpack due to the size of linear system.

PR #157, opened 2020-12-14
LinAlgFactory and Raja Vector Improvements

CameronRutherford:
In developing #141, I came across some potential improvements to the linear algebra factory, along with `hiopVectorRajaPar.cpp`. 

Changes made:
- Addition of default value for memory to be set to in linear algebra factory `createRawArray` method
- Modification of vector tests to use modified linear algebra factory method
- Modified raja vector `copyFromStarting` that uses a vector to call internal `copyFromStarting` that uses raw array data

### DIFF
--- a/src/LinAlg/hiopLinAlgFactory.cpp
+++ b/src/LinAlg/hiopLinAlgFactory.cpp
@@ -206,6 +206,10 @@ hiopMatrixSparse* LinearAlgebraFactory::createMatrixSymSparse(int size, int nnz)
  */
 double* LinearAlgebraFactory::createRawArray(int n, int value)
 {
+  if(n <= 0)
+  {
+    return NULL;
+  }
   if (mem_space_ == "DEFAULT")
   {
     double *tmp = new double[n];
@@ -216,10 +220,14 @@ double* LinearAlgebraFactory::createRawArray(int n, int value)
   {
 #ifdef HIOP_USE_RAJA
     auto& resmgr = umpire::ResourceManager::getInstance();
-    umpire::Allocator al  = resmgr.getAllocator(mem_space_);
-    double *tmp = static_cast<double*>(al.allocate(n*sizeof(double)));
-    resmgr.memset(tmp, value);
-    return tmp;
+    umpire::Allocator hal = resmgr.getAllocator("HOST");
+    umpire::Allocator dal = resmgr.getAllocator(mem_space_);
+    double *tmp = static_cast<double*>(hal.allocate(n*sizeof(double)));
+    double *tmp_dev = static_cast<double*>(dal.allocate(n*sizeof(double)));
+    std::fill(tmp, tmp + n, value);
+    resmgr.copy(tmp_dev, tmp);
+    hal.deallocate(tmp);
+    return tmp_dev;
 #else
     assert(false && "requested memory space not available because Hiop was not"
            "built with RAJA support");
--- a/src/LinAlg/hiopLinAlgFactory.cpp
+++ b/src/LinAlg/hiopLinAlgFactory.cpp
@@ -204,22 +204,28 @@ hiopMatrixSparse* LinearAlgebraFactory::createMatrixSymSparse(int size, int nnz)
 /**
  * @brief Static method to create a raw C array
  */
-double* LinearAlgebraFactory::createRawArray(int n)
+double* LinearAlgebraFactory::createRawArray(int n, int value)
 {
   if (mem_space_ == "DEFAULT")
   {
-    return new double[n];
+    double *tmp = new double[n];
+    std::fill(tmp, tmp + n, value);
+    return tmp;
   }
   else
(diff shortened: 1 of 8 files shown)


src/LinAlg/hiopLinAlgFactory.cpp
```diff
@@ -204,22 +204,36 @@ hiopMatrixSparse* LinearAlgebraFactory::createMatrixSymSparse(int size, int nnz)
...
 {
+  if(n <= 0)
+  {
+    return NULL;
+  }
   if (mem_space_ == "DEFAULT")
   {
-    return new double[n];
+    double *tmp = new double[n];
+    std::fill(tmp, tmp + n, value);
```
ashermancinelli, line 216: @CameronRutherford Can you remind me why we're `std::fill`ing for *every* allocation? Why can't the user `fill` if need be?

ashermancinelli (commented):
lgtm as long as we're sure about `std::fill`. I'm not particularly fond of that solution.

src/LinAlg/hiopLinAlgFactory.cpp
CameronRutherford: ￼We do not have to use `std::fill` for every allocation if we don't want. Instead, we could create two instances of this function. One that fills the array with a default value (using `std::fill`), and one that just initialises memory.

The reason I wanted to fill the memory with a value was to phase out usage of the internal testing function `createLocalBuffer`, however this may be too narrow of a use case to apply this change to all other parts of HiOp code that call this function.

pelesh:
It is unclear what issue(s) this PR addresses and what is its objective. I think more discussion would be desirable before we proceed with this.

tests/LinAlg/vectorTests.hpp
```diff
@@ -209,15 +209,15 @@ class VectorTests : public TestBase
     assert(Nx > Nfrom);
     x.setToConstant(two);
 
-    real_type* from_buffer = createLocalBuffer(Nx, one);
```
CameronRutherford, line 212: In addressing issue #133, I was trying to remove the function `createLocalBuffer`. If I were to use the existing implementation of `hiop::LinearAlgebraFactory::createRawArray`, there is no way to initialise device memory to a default value in an implementation agnostic way.

The solutions to this that I could come up with were:
- Leave methods `createLocalBuffer` and `deleteLocalBuffer` as is in implementation-specific tests
- Add a default value to `hiop::LinearAlgebraFactory::createRawArray` for memory to be initialised to

I supported the latter, as it seemed like a relatively easy change to implement, that would not affect any other areas of code.

Instead of how I have implemented it currently, we could write an overload of the `hiop::LinearAlgebraFactory::createRawArray` method that takes two arguments, so as to avoid using `std::fill` when it is unnecessary.

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -325,15 +325,12 @@ void hiopVectorRajaPar::copyFromStarting(int start_index_in_this, const double*
...
   assert(n_local_ == n_ && "are you sure you want to call this?");
 #endif
   const hiopVectorRajaPar& v = dynamic_cast<const hiopVectorRajaPar&>(vec);
-  assert(start_index + v.n_local_ <= n_local_);
-  
-  auto& rm = umpire::ResourceManager::getInstance();
-  double* vv = const_cast<double*>(v.data_dev_); // scary: 
-  rm.copy(this->data_dev_ + start_index, vv, v.n_local_*sizeof(double));
+  assert(start_index + v.get_local_size() <= n_local_);
+  this->copyFromStarting(start_index, v.local_data_const(), v.get_local_size());
```
CameronRutherford, line 333: I support this change, as currently there is no checking for vectors of size 0 being passed to this function. If a size 0 vector is passed, the function should return without calling `rm.copy`. By calling the internal `copyFromStarting` method, we then have this check performed accordingly, also avoiding the "scary" const_cast in the current implementation.

CameronRutherford (commented):
@pelesh I reviewed and added some more explanation behind the changes that I proposed. While there aren't any explicit bugs/issues that these changes are addressing, they still fix potential problems that I have identified through manual review of the code, along with some changes that would be positive for the unit testing framework.

If you still think that these are unnecessary changes, then we can move on.

tests/LinAlg/vectorTests.hpp
ashermancinelli: Could you just add `fillArray(T*,T);` helper method to test framework?

CameronRutherford: The existing `createLocalBuffer` method in unit testing already has the functionality of what `fillArray` would accomplish. 

It's seeming more as though this is just an unnecessary change that doesn't add any useful functionality to src, only being useful in testing. Because of this, it's probably fine just to leave this as is.

PR #158, opened 2020-12-15
Minor fixes to refactored vector tests.

pelesh:
There were quite a few very minor corrections needed to merge #141 that I fixed right away in the interest of time. @CameronRutherford please take a look and make sure you are happy with this.

I did reorganize `testVector.cpp` code to run tests for different configurations of RAJA and GPU support.

CC @ashermancinelli

### DIFF
--- a/tests/LinAlg/vectorTestsRajaPar.cpp
+++ b/tests/LinAlg/vectorTestsRajaPar.cpp
@@ -64,37 +64,30 @@
 
 namespace hiop { namespace tests {
 
-/// Returns pointer to local vector data
-real_type* VectorTestsRajaPar::getLocalData(hiop::hiopVector* x)
-{
-  x->copyFromDev();
-  return x->local_data_host();
-}
-
 /// Returns const pointer to local vector data
-const real_type* VectorTestsRajaPar::getLocalDataConst(const hiop::hiopVector* x)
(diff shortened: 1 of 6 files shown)


tests/LinAlg/vectorTestsRajaPar.cpp
```diff
@@ -64,37 +64,30 @@
...
+  const hiop::hiopVectorRajaPar* x = dynamic_cast<const hiop::hiopVectorRajaPar*>(x_in);
   x->copyFromDev();
   return x->local_data_host_const();
 }
 
 /// Method to set vector _x_ element _i_ to _value_.
-void VectorTestsRajaPar::setLocalElement(hiop::hiopVector* x, local_ordinal_type i, real_type val)
+void VectorTestsRajaPar::setLocalElement(hiop::hiopVector* x_in, local_ordinal_type i, real_type val)
 {
+  hiop::hiopVectorRajaPar* x = dynamic_cast<hiop::hiopVectorRajaPar*>(x_in);
```
CameronRutherford, line 78: Is the main reason for dynamic casting here readability?

CameronRutherford (approved):
Looks good to me. I think the solution with a seperate template function for the int tests is cleaner and more sustainable than using `std::enable_if`. I was perhaps a little too eager to use the new tool that I had learnt about!

Issue #160, opened 2020-12-16
Refactor CMake to support builds with different vendor specific GPU codes


pelesh:
Currently, HiOp CMake assumes GPU == CUDA. Furthermore, when GPU is enabled, some CUDA specific libraries are required. To add support for AMD GPUs and be able to extend this to Intel GPUs, the CMake should implement following logic:
1. When setting `HIOP_USE_GPU`:
    - Once this flag is set to `On` user should be presented choice of selecting _either_ CUDA _or_ HIP. (We should eventually support the mixture of the two, but for now, this should suffice.)
        - If CUDA is selected `HIOP_USE_MAGMA` is set to `On`. 
            - This should stay like this until there is a configuration option to disable building dense and MDS solvers.
            - If `HIOP_SPARSE` is `On`, then `HIOP_USE_STRUMPACK` is set to `On`, as well.
            - If neither Magma nor Strumpack are found, CMake should send back message that CUDA support is not enabled and turn it off.
        - If HIP is selected `HIOP_USE_HIPMAGMA` is set to `On`.
            - The rest of the logic should be similar as for the CUDA case.
        - Leave a placeholder for Intel GPU selection.
        - If neither CUDA nor HIP support are available, the CMake should return message to that end and turn `HIOP_USE_GPU` off.
     - If GPU support is off, then no options for choosing CUDA or HIP should be visible to the user.
     - RAJA/Umpire selection should be independent of GPU support selection.
2. It is user's responsibility to provide RAJA and Umpire dependencies with support for user's hardware.
    - Per @davidbeckingsale separate RAJA and Umpire libraries should be built (and expected by CMake) for CUDA, HIP and OpenMP support.

CC @ashermancinelli @cnpetra @nychiang @rothpc @nkoukpaizan

cnpetra:
lgtm

nkoukpaizan:
I have a question. Since the `HIOP_SPARSE` logic is contingent on the CUDA/HIP selection, does that mean that there are no plans to use `STRUMPACK` on the CPU, or would there be another GPU-agnostic logic that is not discussed here because we are focusing on the backend selection?

nychiang:
I think the logic is correct.
When setting `HIOP_SPARSE` and `HIOP_USE_GPU`, we need to set `HIOP_USE_STRUMPACK` and check if `STRUMPACK` is available or not. This is due to the fact that `STRUMPACK` is the only sparse solver so far we can use with CUDA.
[...]

PR #162, opened 2020-12-18
CMake GPU vendor configuration

ashermancinelli:
All, please comment where the configuration logic is incorrect - I'm sure I'm missing something. This will require more invasive changes than I have right now, but I need some feedback to make sure the logic is correct. I'll leave this as a draft for now.

CC @pelesh @cnpetra @nychiang @rothpc @nkoukpaizan

Closes: #160


ashermancinelli:
This is being discussed offline.

Issue #151, opened 2020-12-10
Embedded linear algebra kernels in hiopIterate class


pelesh:
Embedded kernels in [`addLinearDampingTermToGrad_x`](https://github.com/LLNL/hiop/blob/develop/src/Optimization/hiopIterate.cpp#L432) and [`addLinearDampingTermToGrad_d`](https://github.com/LLNL/hiop/blob/develop/src/Optimization/hiopIterate.cpp#L486) methods force execution on host (CPU). It would be good to bring these two methods in line with others in `hiopIterate` class methods and use vector kernels from the linear algebra module through an abstract interface. That would make `hiopIterate` agnostic of the memory space or execution policy.

CC @cnpetra @nychiang

cnpetra:
addressed by PR #167

PR #167, opened 2020-12-22
moved last vector kernels from hiopIterate into hiopVector

cnpetra:
Addresses and closes issue #151 

 - provides default and RAJA vectorized implementations
 - added tests in vector testing module
 - updated documentation

Tested on Summit.

### DIFF
--- a/tests/LinAlg/vectorTests.hpp
+++ b/tests/LinAlg/vectorTests.hpp
@@ -1086,6 +1086,98 @@ public:
   }
 
   /**
+   * @brief Test: For addLinearDampingTerm method, which performs a "signed" axpy:
+   * `x[i] = alpha*x[i] + sign*ct` where sign=1 when exactly one of left[i] and right[i]
+   * is 1. and sign=0 otherwise.
+   */
+  bool vectorAddLinearDampingTerm(
+      hiop::hiopVector& x,
+      hiop::hiopVector& left,
+      hiop::hiopVector& right,
+      const int rank)
+  {
+    const local_ordinal_type N = getLocalSize(&x);
+    assert(N == getLocalSize(&left));
+    assert(N == getLocalSize(&right));
+    static const real_type ct = two;
+    static const real_type alpha = quarter;
+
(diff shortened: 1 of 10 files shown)


pelesh:
I am guessing `addLinearDampingTerm` is MPI distributed? If so, it would be good to document that with a comment in the code.

tests/LinAlg/vectorTests.hpp
```diff
@@ -1085,6 +1085,98 @@ class VectorTests : public TestBase
...
+   */
+  bool vectorAddLinearDampingTerm(
+      hiop::hiopVector& x,
+      hiop::hiopVector& left,
+      hiop::hiopVector& right,
+      const int rank)
+  {
+    const local_ordinal_type N = getLocalSize(&x);
+    assert(N == getLocalSize(&left));
+    assert(N == getLocalSize(&right));
```
ashermancinelli, line 1101: I think it would be reasonable to just require size >=4 for this test rather than checking in the test and possibly leaving some corner cases untested.

```suggestion
  {
    const local_ordinal_type N = getLocalSize(&x);
    assert(N == getLocalSize(&left));
    assert(N == getLocalSize(&right));
    assert(N >= 4 && "Test `vectorAddLinearDampingTerm` must have size >=4 to test all use cases for this kernel");
```

ashermancinelli (changes requested):
lgtm! though if it's not too much work I suggest just adding an assertion and simplifying the test for `vectorAddLinearDampingTerm`

tests/LinAlg/vectorTests.hpp
cnpetra: that way one cannot test vectors with unbalanced MPI distribution (as in potentially empty local data in an otherwise quite long vector)

ashermancinelli: Okay, how should I run the test in order to test each case? I would naively just run the test on a vector with size >=4, so I think documenting how the function must be called in order to fully test would be helpful

PR #169, opened 2020-12-22
Magma 2.5.4 CI integration

ashermancinelli:
@cnpetra Adds my magma 2.5.4 builds on target clusters to build scripts. This is already passing on all ci clusters, so I think it's ready for merge.

CC @pelesh 
See #156


pelesh (approved):
Passes tests at all CI targets.

PR #172, opened 2020-12-29
Fix for double pipeline trigger

ashermancinelli:
PNNL CI would trigger two jobs, the first of which would always fail. Should be fixed here 

CC @GitBytes @pelesh

### DIFF
--- a/.github/workflows/push_mirror.yml
+++ b/.github/workflows/push_mirror.yml
@@ -12,5 +12,6 @@ jobs:
           REMOTE: ${{ secrets.GIT_REPO_URL }}
           GIT_USERNAME: ${{ secrets.GIT_USER }}
           GIT_PASSWORD: ${{ secrets.GIT_PASSWORD }}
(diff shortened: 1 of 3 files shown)


ashermancinelli:
@cnpetra you brought up that the full build matrix job leaves the PNNL jobs showing as 'blocked'. I think I can remove this in the same PR.

pelesh:
> @cnpetra you brought up that the full build matrix job leaves the PNNL jobs showing as 'blocked'. I think I can remove this in the same PR.

This was my pet peeve, too. Thanks for fixing this @ashermancinelli!

ashermancinelli:
Alright, the full build matrix will no longer stop the pipeline from showing as 'blocked':

image

In addition, you can manually run all 32 jobs or a particular subset of the full build matrix from the ui:

Screen Shot 2020-12-29 at 2 36 11 PM

The weekly run of all configurations will remain the same.

ashermancinelli:
@cnpetra I believe this is ready for review/merge

PR #174, opened 2021-01-04
remove flawed assert condition, fixes #171

### LINKED ISSUE #171: Assertion Failure
When developing ExaGO we moved to the new hiop interface currently in develop branch. After what appeared to be a successful refactor, the build failed only on intel system (marianas) with hiop develop branch (commit hashes `2e927e4` or `dbb4a7` for example) built in debug mode. Builds passed on power 9 systems (newell and ascent). Pipeline #4584 is an example [(linked here).](https://gitlab.pnnl.gov/exasgd/frameworks/exago/-/pipelines/4584) This failure went away when we rebuilt hiop in release mode.

Assertion:
```console
47: opflow: /people/manc568/projects/hiop/src/Optimization/hiopPDPerturbation.hpp:184: 
    bool hiop::hiopPDPerturbation::compute_perturb_singularity(double&, double&, double&, double&):
    Assertion `delta_cc == 0. && delta_cd == 0.' failed.
```

CC @cnpetra @pelesh @abhyshr

nychiang:
This is a flawed assert condition which should be removed from the code.

This line is only executed on marianas under GCC debug mode. 
Probably it is because some round-off errors send the logic control of inertia correction to different switch case, and then hit the  flawed assert condition. 

@ashermancinelli All the `delta_` variables have been initialized before entering this function, and hence round-off error is the last thing I can guess. Since I don't have access to marianas, I wonder if you could help us to try this fix on marianas.
It probably can converge the problem, under a different path. 

Fixes #171

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -306,6 +306,7 @@ install(FILES src/Interface/hiopInterface.hpp
   src/Optimization/hiopFilter.hpp
   src/Optimization/hiopHessianLowRank.hpp
   src/Optimization/hiopDualsUpdater.hpp
+  src/Optimization/hiopFactAcceptor.hpp
   src/LinAlg/hiopVector.hpp
   src/LinAlg/hiopVectorPar.hpp
   src/LinAlg/hiopVectorInt.hpp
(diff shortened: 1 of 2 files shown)


ashermancinelli:
A couple headers need to be installed via cmake, @nychiang do you mind if I push a commit to this pr?

nychiang:
sure. let's add those headers in this pr.

ashermancinelli (approved):
Great, thanks NY!

PR #175, opened 2021-01-14
Duals init dev

### LINKED ISSUE #97: LSQ duals initialization working with hiopNlpMDS needed
The existing LSQ duals calculator needs to be revisited and re-engineered to work with generic `hiopNlpFormulation`. Currently it assumes a `hiopNlpDenseCons` NLP: https://github.com/LLNL/hiop/blob/2252b90bc40f86b8d842da6fec30105feddbf647/src/Optimization/hiopDualsUpdater.cpp#L111

nychiang:
Add LSQ dual initialization to sparse and MDS linear system.
@cnpetra instead of using augmented system for MDS LSQ problem, I used the same formulation as the LSQ problem for dense linear system. Please have a look of it, and see if we want to add the augmented system approach.
Compare to the augmented system, the main advantage of your dense LSQ problem is that the dimension of the dense matrix is smaller, and, most importantly, many existing codes in [hiopDualsUpdater](https://github.com/LLNL/hiop/blob/5540c55e6289d625b3b951a3471e86daf1c67471/src/Optimization/hiopDualsUpdater.cpp#L218) can be reused for MDS system.

resolve #97

### DIFF
--- a/src/Optimization/hiopDualsUpdater.cpp
+++ b/src/Optimization/hiopDualsUpdater.cpp
@@ -72,54 +72,54 @@ namespace hiop
 
 hiopDualsLsqUpdate::hiopDualsLsqUpdate(hiopNlpFormulation* nlp) 
   : hiopDualsUpdater(nlp), 
-    _mexme_(nullptr), _mexmi_(nullptr), _mixmi_(nullptr), _mxm_(nullptr), 
+    mexme_(nullptr), mexmi_(nullptr), mixmi_(nullptr), mxm_(nullptr), 
     M_(nullptr),
     rhs_(nullptr), rhsc_(nullptr), rhsd_(nullptr),
-    _vec_n_(nullptr),_vec_mi_(nullptr),
+    vec_n_(nullptr),vec_mi_(nullptr),
     linSys_(nullptr),
-    lsq_dual_init_max(1e3),
+    lsq_dual_init_max_(1e3),
     augsys_update_(false) 
 {
-  hiopNlpSparse* nlpsp = dynamic_cast<hiopNlpSparse*>(_nlp);
+  hiopNlpSparse* nlpsp = dynamic_cast<hiopNlpSparse*>(nlp_);
   if(nullptr!=nlpsp)
   {
 #ifndef HIOP_SPARSE
       assert(0 && "should not reach here!");
 #endif // HIOP_SPARSE        
       augsys_update_ = true;
-      rhs_    = LinearAlgebraFactory::createVector(_nlp->n() + _nlp->m_ineq() + _nlp->m());
+      rhs_    = LinearAlgebraFactory::createVector(nlp_->n() + nlp_->m_ineq() + nlp_->m());
   }
   else
   {
-    _mexme_ = LinearAlgebraFactory::createMatrixDense(_nlp->m_eq(),   _nlp->m_eq());
-    _mexmi_ = LinearAlgebraFactory::createMatrixDense(_nlp->m_eq(),   _nlp->m_ineq());
-    _mixmi_ = LinearAlgebraFactory::createMatrixDense(_nlp->m_ineq(), _nlp->m_ineq());
-    _mxm_   = LinearAlgebraFactory::createMatrixDense(_nlp->m(), _nlp->m());
+    mexme_ = LinearAlgebraFactory::createMatrixDense(nlp_->m_eq(),   nlp_->m_eq());
+    mexmi_ = LinearAlgebraFactory::createMatrixDense(nlp_->m_eq(),   nlp_->m_ineq());
+    mixmi_ = LinearAlgebraFactory::createMatrixDense(nlp_->m_ineq(), nlp_->m_ineq());
+    mxm_   = LinearAlgebraFactory::createMatrixDense(nlp_->m(), nlp_->m());
 
-    M_      = LinearAlgebraFactory::createMatrixDense(_nlp->m(), _nlp->m());
-    rhs_    = LinearAlgebraFactory::createVector(_nlp->m());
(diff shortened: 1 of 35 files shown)


cnpetra:
Indeed there is a considerable computational advantage, only that now you have to follow-up with RAJA implementation for the sparse mat * mat^T :)

cnpetra:
re: Some minor indentation and alignment﻿ issues are corrected, i.e., use 2-space indent to replace a \tab

I am not sure it's a good idea to do it now, as least not for hiopAlgFilterIPM.cpp because it shows diffs in hundreds of lines and it's impossible to track changes. github diff also shows that the safe_mode outer control loop changed considerably, which I think is not the case.

pelesh:
> Indeed there is a considerable computational advantage, only that now you have to follow-up with RAJA implementation for the sparse mat * mat^T :)

I think there needs to be a unit test for the new kernel, as well. It is not in the PR.

cnpetra:
> > Indeed there is a considerable computational advantage, only that now you have to follow-up with RAJA implementation for the sparse mat * mat^T :)
> 
> I think there needs to be a unit test for the new kernel, as well. It is not in the PR.

PR is still a draft, there will be follow-ups

nychiang:
@pelesh I started this PR in a draft mode, since I want to share the code with @cnpetra first before I move forward.

nychiang:
Unit test for the new function in matrix kernel is added.
RAJA implementation works fine.
Pass ORNL pipelines.

Note that now I use LSQ dual initialization for MDS_Ex5 and Sparse_Ex6.
For MDS_Ex4, if we use LSQ dual initialization, some [element-wise tests](https://github.com/LLNL/hiop/blob/89bed197c3fdc30b9f0c83aaa185170129c67207/CMakeLists.txt#L432) will fail. 
I don't think this is a problem, since the test problems can still converge with LSQ initial point.
For more details, [see here](https://code.ornl.gov/ecpcitest/exasgd/hiop/-/jobs/904727)

ashermancinelli (commented):
@nychiang, do you mind if I push directly to your branch with a fix for the PNNL CI?

src/Optimization/hiopDualsUpdater.cpp
```diff
@@ -46,57 +46,106 @@
...
-  delete _mixmi;
-  delete _mxm;
-  delete M;
-  delete rhs;
-  delete rhsc; 
-  delete rhsd;
-  delete _vec_n;
-  delete _vec_mi;
+  if(linSys_) delete linSys_;
+  if(_mexme_) delete _mexme_;
```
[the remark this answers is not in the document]
cnpetra: that's my doing from a long time ago. Will take care of this later in this pr

src/Optimization/hiopDualsUpdater.cpp
```diff
@@ -46,57 +46,106 @@
...
-  rhs_copy = rhs->alloc_clone();
-  _mixme = LinearAlgebraFactory::createMatrixDense(nlpd->m_ineq(), nlpd->m_eq());
+    M_copy_ = LinearAlgebraFactory::createMatrixDense(_nlp->m(), _nlp->m());
+    rhs_copy_ = LinearAlgebraFactory::createVector(_nlp->m());
+    _mixme_ = LinearAlgebraFactory::createMatrixDense(_nlp->m_ineq(), _nlp->m_eq());
+#endif
+  }
+
+#ifdef HIOP_USE_MPI
+  long long* vec_distrib = _nlp->get_vec_distrib();
```
cnpetra, line 110: I find it the following lines a bit too complicated/much. This class should not know about `vec_distrib`. Apparently we have two "factories" for linear algebra, `LinearAlgebraFactory` and `hiopNlpFormulation`. `LinearAlgebraFactory` decides between memory spaces, but it is up to implementations of `hiopNlpFormulation` to decide whether the vectors are MPI-based: `hiopNlpSparse` does not need MPI while `hiopNlpDenseConstraints` does. I looks like it will be cleaner to allocate via `hiopNlpFormulation` `alloc_primal_vector` and other `alloc_xxx` functions and within these functions to use `LinearAlgebraFactory` to instantiate the right mem space type vectors/matrices. @pelesh @ashermancinelli @nychiang what do you think?

nychiang: Yes, I think that's a good idea. Currently we are using `LinearAlgebraFactory::createVector` without MPI information in many places of hiopKKTxxxx classes. It's better to change all of them to  `hiopNlpFormulation::allco_xxx` and let `hiopNlpFormulation` to determine the MPI usage.

ashermancinelli: How cleanly is the distribution logic in `hiopNlpFormulation` implemented? If it's pretty clean, I think it would make more sense to move that logic to `LinearAlgebraFactory`. At this point it might make sense to think about extending linalg factory with additional implementations as we discussed last quarter. I don't want to bring in a bunch of scope creep though, so we can discuss this elsewhere if it would be too much work for this pr.

cnpetra: @ashermancinelli  `LinearAlgebraFactory` already knows how to create MPI vectors of all sorts. And `hiopNlpFormulation` uses internally that functionality from `LinearAlgebraFactory` -- if/when needed, depending on the nlp formulation type. What I am saying is that in hiopKKTxxxx classes, we should use `hiopNlpFormulation` to do the allocations.

ashermancinelli: I see, this makes perfect sense. Sounds great to me.

src/Optimization/hiopDualsUpdater.cpp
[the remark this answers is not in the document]
cnpetra: I've brought those in line with coding styleguide

PR #181, opened 2021-01-25
Adapter fix

cnpetra:
This PR updates various adapter(s) to work with the latest version of HiOp's interfaces.


src/LinAlg/hiopVectorPar.cpp
```diff
@@ -795,6 +857,13 @@ bool hiopVectorPar::isfinite_local() const
   return true;
 }
 
+void hiopVectorPar::print()
```
nychiang, line 860: @cnpetra Add this function to easily display the vector in gdb. Maybe we should remove it from this PR.

cnpetra: it does not hurt to have it in there

Issue #179, opened 2021-01-22
Relax variable/constraint bound before solving the problem 


nychiang:
To have a better numerical behavior, we need to relax variable/constraint bound before solving the problem.
Add a user option to control the relaxation.

Issue #152, opened 2020-12-10
Enable using device memory space when running HiOp on GPU


pelesh:
Currently execution on GPU works only when using unified virtual memory. To enable using device memory space, issues #94, #103, #147, and #151 need to be resolved, and PR #76 needs to be merged. The remaining work may involve some refactoring in `hiopNlpFormulation` class.

This issue is to coordinate all activities needed to enable HiOp GPU execution while using device memory space.

CC @cnpetra @ashermancinelli @nychiang

cnpetra:
closing this since it has been addressed. Issue #103 is still active, but wasn't required for the resolution of this issue.

PR #182, opened 2021-01-30
scaling the problem before solving it

### LINKED ISSUE #177: Add scaling to HiOp
To have better performance for some hard problems, we need to scale the problem before solving it.
Add a user option to choose the method of scaling.

nychiang:
Ready for PR.

resolve #177


src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
```diff
@@ -459,6 +459,57 @@ double hiopMatrixRajaSparseTriplet::max_abs_value()
...
+    RAJA_LAMBDA(RAJA::Index_type itnz)
+    {
+      const int row_id = iRow[itnz];
+      if(i==row_id) {
+        norm.max(std::abs(values[itnz]));
+      }
+    });
+
+    vd[i] = norm.get();
+  }
```
nychiang, line 487: I don't think this is a sensible implementation, and it is not efficient at all.
I wonder if there is a RAJA array of type RAJA::ReduceMax, and then we can use it to recode the norm of each row.
Then we can use the following code without the outer for loop.
    RAJA::ReduceArray<RAJA::ReduceMax<hiop_raja_reduce, double>> norm_row(nrows_, 0.0); 
    RAJA::forall<hiop_raja_exec>(RAJA::RangeSegment(0, nnz_),
    RAJA_LAMBDA(RAJA::Index_type itnz)
    {
      const int row_id = iRow[itnz];
      norm_row[row_id].max(std::abs(values[itnz]));
    });
@cnpetra

nychiang:
@pelesh @ashermancinelli you may want to have a look of the new unit tests I added to the vector/matrix kernel. I have addressed the issue #193 in this PR as well. Maybe I should remove it from this PR and wait for your overall fix to unit test.

src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
cnpetra: I don't think that's available. I think this can be implemented efficiently with RAJA::forall and list segments (will require no reduce ops). Will pm you...

src/Drivers/nlpSparse_ex6.cpp
```diff
@@ -6,18 +6,18 @@
 
 /* Test with bounds and constraints of all types. For some reason this
  *  example is not very well behaved numerically.
- *  min   sum 1/4* { (x_{i}-1)^4 : i=1,...,n}
+ *  min   sum scal*1/4* { (x_{i}-1)^4 : i=1,...,n}
  *  s.t.
- *            4*x_1 + 2*x_2                     == 10
- *        5<= 2*x_1         + x_3
- *        1<= 2*x_1                 + 0.5*x_i   <= 2*n, for i=4,...,n
+ *             scal * 4*x_1 + 2*x_2                     == scal*10
```
cnpetra, line 11: what is the role of/why do we need "scal" ?

src/LinAlg/hiopMatrixRajaDense.hpp
```diff
@@ -52,6 +52,8 @@
  * @author Robert Rutherford <robert.rutherford@pnnl.gov>, PNNL
  * @author Asher Mancinelli <asher.mancinelli@pnnl.gov>, PNNL
  * @author Slaven Peles <slaven.peles@pnnl.gov>, PNNL
+ * @author Slaven Peles <slaven.peles@pnnl.gov>, PNNL
```
cnpetra, line 55: didn't know Slaven has double personalities :)

src/LinAlg/hiopMatrixDenseRowMajor.cpp
```diff
@@ -787,6 +787,48 @@ double hiopMatrixDenseRowMajor::max_abs_value()
...
+  
+  hiopVectorPar& vec=dynamic_cast<hiopVectorPar&>(ret_vec);
+  assert(m_local_==vec.get_local_size());
+  
+  for(int irow=0; irow<m_local_; irow++)
+  {
+    maxv = DLANGE(&norm, &one, &n_local_, M_[0]+(irow*n_local_), &one, nullptr);
+#ifdef HIOP_USE_MPI
+    double maxvg;
+    int ierr=MPI_Allreduce(&maxv,&maxvg,1,MPI_DOUBLE,MPI_MAX,comm_); assert(ierr==MPI_SUCCESS);
```
cnpetra, line 804: Wouldn't be better to do a single `MPI_Allreduce` (of `vec`) after the for loop?

src/LinAlg/hiopMatrixRajaDense.cpp
```diff
@@ -1376,6 +1376,70 @@ double hiopMatrixRajaDense::max_abs_value()
...
+ * @todo Consider using BLAS call (<D>LANGE)
+ */
+void hiopMatrixRajaDense::row_max_abs_value(hiopVector &ret_vec)
+{  
+  assert(ret_vec.get_size() == m());
+  ret_vec.setToZero();
+
+  auto& vec = dynamic_cast<hiopVectorRajaPar&>(ret_vec);
+  vec.copyFromDev();
+  double* vd = vec.local_data_host();
```
pelesh, line 1391: Why copy to the host mirror here?

src/LinAlg/hiopMatrixRajaDense.cpp
```diff
@@ -1376,6 +1376,70 @@ double hiopMatrixRajaDense::max_abs_value()
...
+  
+  for (int irow = 0; irow < m_local_; irow++)
+  {
+    RAJA::ReduceMax<hiop_raja_reduce, double> norm(0.0);
+    RAJA::View<const double, RAJA::Layout<2>> Mview(data, m_local_, n_local_);
+    RAJA::forall<hiop_raja_exec>(RAJA::RangeSegment(0, n_local_),
+      RAJA_LAMBDA(RAJA::Index_type j)
+      {
+        norm.max(fabs(Mview(irow,j)));
+      });
```
pelesh, line 1403: Perhaps consider parallelizing the outer loop?

cnpetra: yes, like the one for sparse matrix, parallelize over rows. I think it can be done without using reduce

src/LinAlg/hiopVectorPar.hpp
```diff
@@ -117,6 +117,11 @@ class hiopVectorPar : public hiopVector
   virtual void componentMult( const hiopVector& v );
   virtual void componentDiv ( const hiopVector& v );
   virtual void componentDiv_w_selectPattern( const hiopVector& v, const hiopVector& ix);
+  virtual void component_min ( const double constant);
```
cnpetra, line 120: extra space after "min" and before "const"... same for other functions

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -773,6 +773,80 @@ void hiopVectorRajaPar::componentDiv_w_selectPattern( const hiopVector& vec, con
...
+}
+
+/**
+ * @brief Set `this` vector elemenwise to the minimum of itself and the corresponding component of 'vec'.
+ * 
+ * @pre `this` and `vec` have same partitioning.
+ * @post `vec` is not modified
+ * 
+ */
+void hiopVectorRajaPar::component_min (const hiopVector& vec)
```
cnpetra, line 798: extra space after "min"

src/LinAlg/hiopVectorRajaPar.hpp
```diff
@@ -106,6 +106,10 @@ class hiopVectorRajaPar : public hiopVector
   virtual void componentMult( const hiopVector& v );
   virtual void componentDiv ( const hiopVector& v );
   virtual void componentDiv_w_selectPattern( const hiopVector& v, const hiopVector& ix);
+  virtual void component_min ( const double constant);
```
cnpetra, line 109: extra spaces

src/Optimization/hiopNlpFormulation.cpp
```diff
@@ -480,6 +482,38 @@ bool hiopNlpFormulation::finalizeInitialization()
...
+  const double max_grad = options->GetNumeric("scaling_max_grad");
+
+  double obj_scale = max_grad/gradf.infnorm();
+  if(obj_scale>1.)
+  {
+    obj_scale=1.;
+    return false;
+  }
+
+  nlp_scaling = new hiopNLPObjGradScaling(max_grad, c, d, gradf, Jac_c, Jac_d);
```
cnpetra, line 502: `apply_scaling` is called each time the starting procedure is called, which can be called multiple times when restarting the nlp. I've looked at the code and I think `nlp_scaling` will be added each time the nlp is restarted... looks like you may have forgotten to test `nlp_scaling` for null. 

this raises the questions whether the scaling should be recomputed when the nlp is restarted ...

src/Utils/hiopOptions.cpp
```diff
@@ -210,6 +210,20 @@ void hiopOptions::registerOptions()
...
+  {
+    vector<string> range(2); range[0]="none"; range[1]="gradient";
+    registerStrOption("scaling_type", "gradient", range,
+		      "The method used for scaling the problem before solving it."
+		      "Setting this option to 'gradient' will scale the problem, guaranteeing the maximum "
+		      "gradient at the initial point is less or equal to scaling_max_grad (default 'none')");
+    
+    registerNumOption("scaling_max_grad", 100, 1e-20, 1e+20,
+		      "The maximum initial gradient allowed in the problem."
+		      "Any gradient that is larger than this positive value will be scaled."
```
cnpetra, line 223: maybe something like this sound better?

"The problem will be rescaled is the inf-norm of the gradient at the starting point "
"will be larger than the value of this option"

src/LinAlg/hiopMatrixRajaDense.cpp
nychiang: vd is used outside the RAJA section. 
If I don't make a host mirror, line 1411 will return me a segment fault.

src/Drivers/nlpSparse_ex6.cpp
nychiang: We can easily scale EX6, and then test how gradient scaling works.

src/LinAlg/hiopMatrixRajaDense.cpp
pelesh: Correct, the reduce functionality can be implemented sequentially in that context. Parallelizing over matrix rows seems to be a good first step. We can think about optimizing this further later on.

src/LinAlg/hiopMatrixRajaDense.cpp
pelesh: I think this will be taken care of if you parallelize the outer instead of the inner loop.

src/Utils/hiopOptions.cpp
nychiang: sounds better, except for the typo.

src/Optimization/hiopNlpFormulation.cpp
nychiang: See [line 196](https://github.com/LLNL/hiop/blob/5e6b92008fbc998f200dbfd9ceb341f0494abb47/src/Optimization/hiopNlpFormulation.cpp#L196)
I think it release the memory in object nlp_transformations, and all the transformations, e.g., fixedVarsRelaxer and one I am working on for issue #179, are added each time when the nlp is restarted.

Note that [line 196](https://github.com/LLNL/hiop/blob/5e6b92008fbc998f200dbfd9ceb341f0494abb47/src/Optimization/hiopNlpFormulation.cpp#L196) will make pointer nlp_scaling as a dangling pointer, and then it is reset at  [line 502](https://github.com/LLNL/hiop/blob/5e6b92008fbc998f200dbfd9ceb341f0494abb47/src/Optimization/hiopNlpFormulation.cpp#L502)

nychiang: I think `nlp_transformations.clear();` on line [196](https://github.com/LLNL/hiop/blob/dev-scaling/src/Optimization/hiopNlpFormulation.cpp#L196) will always delete
Same question apply to other transformations, e.g., relax bounds for fixed variables, and relax var/con bounds before solving the problem (in the new branch I am working on).
Currently I can see you reset object `nlp_transformations` and add all the transformations back to it when nlp is restarted.

src/LinAlg/hiopMatrixRajaDense.cpp
nychiang: I tried to follow the suggestion given in [RAJA issue 246 ](https://github.com/LLNL/RAJA/issues/246), to implement an array of RAJA reducer. It turns out Will's solution only works on CPU.
Will move back to parallelizing over rows.

pelesh: Perhaps in this case a better reference would be RAJA tutorial on [matrix-matrix multiplication](https://raja.readthedocs.io/en/v0.13.0/tutorial/matrix_multiply.html).

nychiang:
@pelesh @cnpetra I just push the changes to RAJA code. Now I parallelize over rows (the outer loop). 
In my opinion, the main difference between parallelizing over rows and parallelizing over nonzeros is that we need to assume all the nonzeros are sorted when parallelizing over rows. I think this is fine since we have some other functions in matrix kernel require this precondition.

Therefore, similar to other functions, `hiopMatrixRajaSparseTriplet::row_max_abs_value`, can only be applied to the sorted matrices such as Hess and Jacobian, but not the sparse KKT matrix. I assemble the sparse KKT matrix by appending nonzeros in triplet form from different matrices. Therefore it is NOT sorted.

nychiang:
@cnpetra @pelesh 
I have fixed all the issues here. Ready for review.
It also addresses some issues reported in #152. For example, now I use functions from vector/matrix kernel in `hiopNlpTransform`. This allow us to remove fix variables and do scaling on GPU.

PR #185, opened 2021-02-08
Expand descriptions of style guidelines.

pelesh:
Organize style guidelines in sections and put separate examples in corresponding sections. Add new guidelines with examples.


ashermancinelli:
@cnpetra @pelesh @nychiang Looks great - we will use/enforce whatever you decide on.

cnpetra:
it would be indeed helpful if we can do something like that.

Issue #131, opened 2020-11-27
Memory space can only be set statically


ashermancinelli:
Currently, `LinearAlgebraFactory` class is managing HiOp memory space using static methods. Using static methods may be inflexible solution that could potentially limit future development, especially on hybrid hardware architectures.

pelesh:
Most likely using singleton pattern would provide enough flexibility for implementations on heterogeneous hardware. I would consider that first.

ashermancinelli:
I don't think Singleton pattern would give us anything over our current api. I believe we discussed using factory pattern offline though, is this correct?

cnpetra:
the linear algebra factory logic should be part of NlpFormulation (and it is right now to some extent) as this is the entity that needs to be assigned and ran into a particular memory space. Underneath the NlpFormulation,  the code decides memory space and types for linear algebra objects based on the user options; then uses `new_clone` and `alloc_copy` to create new objects. However, in some parts, `new_clone` and `alloc_copy` are not sufficient, as vectors/matrices of new sizes need to be created. Also, sometimes the vectors/matrices need to be in a different memory space (usually host) than the NlpFormulation is supposedly running in. I think singleton pattern would work for these situations. Factory pattern can be a bit more invasive because the specific instance of the linear algebra factory needs to be available to all the inner classes creating vectors/matrices; also, it needs to be able allocate in different memory spaces, which kinda defeats the purpose. 

In other words, NlpFormulation is already a factory pattern and makes more sense for it to rely on a  a singleton linear algebra factory.

pelesh:
Provided #241 works out well, we could get what we need with minimal changes to the existing linear algebra factory and have something like:
```c++
auto& f = hiopLinAlgFactory::getInstance(); 
hiopVector* x = f.createVector<hiopCudaPolices>(1000, "UM");
```
This would use singleton implementation to create a RAJA vector with data in unified memory. Probably there are some details to sort out, but this could work for us.

ashermancinelli:
What does this give us over static method?
```cpp
auto* vec = hiopLinAlg::creatVector<hiopCudaPolicies>(size, mem_space);
```

cnpetra:
> What does this give us over static method?
> 
> ```c++
> auto* vec = hiopLinAlg::creatVector<hiopCudaPolicies>(size, mem_space);
> ```

It appears that static-only methods do the job for now (meaning fixed policies), see PR#274

cnpetra:
closed by PR#274

PR #191, opened 2021-02-10
Primal Decomposition Solver

cnpetra:
A complete and stable implementation of the primal decomposition algorithm.

To facilitate a quick merge in `develop`, this PR does not fully update the user manual and leaves out some options -- these will be done by a follow-up PR.


src/Drivers/nlpPriDec_ex9_driver.cpp
```diff
@@ -0,0 +1,186 @@
...
+  PriDecMasterProblemEx9 pridec_problem(nx, nx, nS, S);
+  hiop::hiopAlgPrimalDecomposition pridec_solver(&pridec_problem, MPI_COMM_WORLD);
+  pridec_solver.set_initial_alpha_ratio(0.5);
+  //pridec_solver.set_tolerance(1e-6);
+  //pridec_solver.set_max_iteration(5);
+  auto status = pridec_solver.run();
+  
+  if(status!=Solve_Success){
+    if(rank==0)
+      printf("Solve was NOT successfull.");
```
cnpetra, line 147: brackets for single lines

src/Drivers/nlpPriDec_ex9_driver.cpp
```diff
@@ -0,0 +1,186 @@
...
+  }else{
+    if(rank==0)
+      printf("Solve was successfull. Optimal value: %12.5e\n",
```
cnpetra, line 150: brackets for single lines

src/Drivers/nlpPriDec_ex9_user_recourse.hpp
```diff
@@ -0,0 +1,475 @@
...
+    assert(num_cons==nx_||num_cons==0);
+    //indexes for sparse part
+    if(num_cons==0) {
+      return true;
+    }
+
+    if(iJacS!=NULL && jJacS!=NULL) {
+      int nnzit=0;
+      for(int itrow=0; itrow<num_cons; itrow++) {
+	const int con_idx = (int) idx_cons[itrow];
```
cnpetra, line 255: alignment is off

src/Drivers/nlpPriDec_ex9_user_recourse.hpp
```diff
@@ -0,0 +1,475 @@
...
+	  //sparse Jacobian ineq w.r.t x and s
+	}
+      }
+      assert(nnzit==nnzJacS);
+    }
+    //values for sparse Jacobian if requested by the solver
+    if(MJacS!=NULL) {
+      int nnzit=0;
+      for(int itrow=0; itrow<num_cons; itrow++) {
+	const int con_idx = (int) idx_cons[itrow];
```
cnpetra, line 306: alignment is off

src/Drivers/nlpSparse_ex9.hpp
```diff
@@ -0,0 +1,406 @@
...
+      }
+    }
+    //std::cout<<"gradf15 "<<gradf[15]<<std::endl;
+    return true;
+  }
+
+  /* Four constraints no matter how large n is */
+  // This constraint attempts to use existing functions
+  bool eval_cons(const long long& n, const long long& m,
+		                const double* x, bool new_x, double* cons)
```
cnpetra, line 176: alignment issue

src/Drivers/nlpSparse_ex9.hpp
```diff
@@ -0,0 +1,406 @@
...
+        cons[conidx+i*nx_+nx_-1] += x[nx_+i*nx_+j]*x[nx_+i*nx_+j];
+      }
+    }
+    
+    return true;
+  }
+  
+  bool eval_cons(const long long& n, const long long& m,
+                    const long long& num_cons, const long long* idx_cons,
+			              const double* x, bool new_x, double* cons)
```
cnpetra, line 216: alignment issues

src/Drivers/nlpSparse_ex9.hpp
```diff
@@ -0,0 +1,406 @@
...
+  {
+    return false;
+  }
+  bool eval_Jac_cons(const long long& n, const long long& m,
+                        const long long& num_cons, const long long* idx_cons,
+                        const double* x, bool new_x,
```
cnpetra, line 222: alignment issue

src/Drivers/nlpSparse_ex9.hpp
```diff
@@ -0,0 +1,406 @@
...
+                        const int& nnzJacS, int* iJacS, int* jJacS, double* MJacS)
+  {
+    return false;
+  }
+
+  bool eval_Jac_cons(const long long& n, const long long& m,
+                          const double* x, bool new_x,
+                          const int& nnzJacS, int* iJacS, int* jJacS, double* MJacS)
```
cnpetra, line 230: alignment issue

src/Drivers/nlpSparse_ex9.hpp
```diff
@@ -0,0 +1,406 @@
...
+      iJacS[nnzit] = conidx;   jJacS[nnzit++] = 1;
+      conidx++;
+
+      // --- constraint 2 body ---> 2*x_1 + x_3
+      iJacS[nnzit] = conidx;   jJacS[nnzit++] = 0;
+      iJacS[nnzit] = conidx;   jJacS[nnzit++] = 2;
+      conidx++;
+
+      // --- constraint 3 body --->   2*x_1 + 0.5*x_i, for i>=4
+      for(auto i=3; i<nx_; i++){
```
cnpetra, line 253: missing space before bracket; the following lines are indented with 4 whitespaces instead of 2

src/Interface/hiopInterfacePrimalDecomp.cpp
```diff
@@ -0,0 +1,192 @@
...
+}
+
+void hiopInterfacePriDecProblem::RecourseApproxEvaluator::
+set_xc_idx(const std::vector<int>& idx)
+{
+  assert(nc_=idx.size());
+  xc_idx_ = idx;
+}
+
+int hiopInterfacePriDecProblem::RecourseApproxEvaluator::get_S() const {return S_;}
```
cnpetra, line 180: these `get_` methods should be expanded on multiple lines as per coding styleguide

src/Optimization/hiopAlgPrimalDecomp.cpp
```diff
@@ -0,0 +1,1117 @@
...
+    MPI_Request request_;
+  private:
+    int idx;
+  };
+#endif
+
+
+
+hiopAlgPrimalDecomposition::HessianApprox::
+HessianApprox() :HessianApprox(-1)
```
cnpetra, line 96: initialization list should be on the next line

src/Optimization/hiopAlgPrimalDecomp.cpp
```diff
@@ -0,0 +1,1117 @@
...
+/* a trust region way of updating alpha ratio
+ * rkm1: true recourse value at {k-1}
+ * rk: true recourse value at k
+ */
+void hiopAlgPrimalDecomposition::HessianApprox::
+update_ratio_tr(const double rhok,
+		const double rkm1, 
+		const double rk, 
+		const double alpha_g_ratio,
+                double& alpha_ratio)
```
cnpetra, line 256: alignment issue

src/Optimization/hiopAlgPrimalDecomp.cpp
```diff
@@ -0,0 +1,1117 @@
...
+    }
+    request_ = new MPI_Request[4];   
+  #endif
+  x_ = new double[n_];
+}
+
+hiopAlgPrimalDecomposition::
+hiopAlgPrimalDecomposition(hiopInterfacePriDecProblem* prob_in,
+                           const int nc, 
+			   const std::vector<int>& xc_index,
```
cnpetra, line 428: alignment issue

src/Optimization/hiopAlgPrimalDecomp.cpp
```diff
@@ -0,0 +1,1117 @@
...
+        
+            //post recv for new index
+            req_cont_idx->post_recv(1, rank_master, comm_world_);
+            //ierr = MPI_Irecv(&cont_idx[0], 1, MPI_INT, rank_master, 1, comm_world_, &request_[0]); 	  
+          }
+        }
+      }
+
+      if(my_rank_==0)
+      {
```
cnpetra, line 840: bracket on the previous line ?!?

src/Optimization/hiopAlgPrimalDecomp.cpp
```diff
@@ -0,0 +1,1117 @@
...
+            x0[i] = x_[xc_idx_[i]];
+          }
+        } else {
+          assert(nc_==n_);
+          for(int i=0;i<nc_;i++) {
+            x0[i] = x_[i];
+          }
+        }
+
+        if(it==0) {
```
cnpetra, line 860: alignment issues, lines 840-946

src/Optimization/hiopAlgPrimalDecomp.hpp
```diff
@@ -0,0 +1,250 @@
...
+public:
+
+  //constructor
+  hiopAlgPrimalDecomposition(hiopInterfacePriDecProblem* prob_in,
+                             MPI_Comm comm_world=MPI_COMM_WORLD);
+
+  hiopAlgPrimalDecomposition(hiopInterfacePriDecProblem* prob_in,
+		             const int nc, 
+			     const std::vector<int>& xc_index,
+                             MPI_Comm comm_world=MPI_COMM_WORLD);
```
cnpetra, line 48: alignment issue

src/Optimization/hiopAlgPrimalDecomp.hpp
```diff
@@ -0,0 +1,250 @@
...
+  void set_initial_alpha_ratio(const double ratio);
+    
+  void set_max_iteration(const int max_it); 
+  
+  void set_tolerance(const double tol);
+  
+  /* Contains information of a solution step including function value 
+   * and gradient. Used for storing the solution for the previous iteration
+   */
+  struct prev_sol{
```
cnpetra, line 84: the name should probably be `PrevSol` as per coding guidestyle

src/Optimization/hiopAlgPrimalDecomp.hpp
```diff
@@ -0,0 +1,250 @@
...
+  };
+
+  /* Struct for the quadratic coefficient alpha in the recourse approximation
+   * function. It contains quantities such as s_{k-1} = x_k-x_{k-1} that is 
+   * otherwise not computed but useful for certian update rules for alpha,
+   * as well as the convergence measure. The update function is called
+   * every iteration to ensure the values are up to date.
+   * The xk here should only be the coupled x.
+   */
+  struct HessianApprox{
```
cnpetra, line 120: space before bracket

src/Optimization/hiopAlgPrimalDecomp.hpp
```diff
@@ -0,0 +1,250 @@
...
+     * objective and linear objective. 
+     * User can provide a global maximum and minimum for alpha
+     */
+    void update_ratio();
+
+    //a trust region way of updating alpha ratio
+    //rkm1: true recourse value at {k-1}
+    //rk: true recourse value at k
+    void update_ratio_tr(const double rhok,const double rkm1, const double rk, const double alpha_g_ratio,
+		         double& alpha_ratio);
```
cnpetra, line 158: alignment

nychiang (approved):

nychiang:
Do we have plan to use this solver on GPU with MPI? Seems the current implementation only supports CPU+MPI.

src/Drivers/IpoptAdapter_driver_ex9.cpp
```diff
@@ -0,0 +1,103 @@
...
+  nlp.options->SetStringValue("Hessian", "analytical_exact");
+  nlp.options->SetStringValue("duals_update_type", "linear"); 
+//  nlp.options->SetStringValue("duals_init", "zero"); // "lsq" or "zero"
+  nlp.options->SetStringValue("compute_mode", "cpu");
+  nlp.options->SetStringValue("KKTLinsys", "xdycyd");
+  // nlp.options->SetStringValue("KKTLinsys", "full");
+  // nlp.options->SetStringValue("write_kkt", "yes");
+  // nlp.options->SetIntegerValue("max_iter", 100);
+  nlp.options->SetNumericValue("mu0", 0.1);
+  hiopAlgFilterIPMNewton solver(&nlp);
```
pelesh, line 37: I would remove commented lines and document each option. For example
```suggestion
  nlp.options->SetStringValue("duals_update_type", "linear"); 
  nlp.options->SetStringValue("duals_init", "zero");   // "lsq" or "zero"
  nlp.options->SetStringValue("compute_mode", "cpu");  // cpu, hybrid or gpu
  nlp.options->SetStringValue("KKTLinsys", "xdycyd");  // xdycyd or full
  nlp.options->SetStringValue("write_kkt", "no");      // yes/no
  nlp.options->SetIntegerValue("max_iter", 100);
  nlp.options->SetNumericValue("mu0", 0.1);
  hiopAlgFilterIPMNewton solver(&nlp);
```

src/Drivers/IpoptAdapter_driver_ex9.cpp
```diff
@@ -0,0 +1,103 @@
...
+  // app->Options()->SetNumericValue("bound_frac", 1e-8);
+  // app->Options()->SetNumericValue("bound_push", 1e-8);
+  // app->Options()->SetNumericValue("slack_bound_push", 1e-24);
+  // app->Options()->SetNumericValue("bound_relax_factor", 0.);
+  // app->Options()->SetNumericValue("constr_mult_init_max", 0.001);
+  // app->Options()->SetNumericValue("kappa1", 1e-8);
+  // app->Options()->SetNumericValue("kappa2", 1e-8);
+
+  // app->Options()->SetStringValue("output_file", "ipopt.out");
+  // app->Options()->SetStringValue("derivative_test", "second-order"); //"only-second-order"
```
pelesh, line 80: I suggest removing commented out code before merging to `develop`.

src/Drivers/IpoptAdapter_driver_ex9.cpp
```diff
@@ -0,0 +1,103 @@
...
+  // Initialize the IpoptApplication and process the options
+  ApplicationReturnStatus status;
+  status = app->Initialize();
+  if( status != Solve_Succeeded ) {
+      std::cout << std::endl << std::endl << "*** Error during initialization!" << std::endl;
+      return (int) status;
+    }
```
pelesh, line 87: Brace alignment:
```suggestion
  if( status != Solve_Succeeded ) {
      std::cout << std::endl << std::endl << "*** Error during initialization!" << std::endl;
      return (int) status;
  }
```

src/Drivers/Makefile_ex9
```diff
@@ -0,0 +1,95 @@
+#########################################################################
+# For TESTING HiOp only. Not intended for any kind of production use
+#
+# Makefile for HiOp's example driver for IpoptAdapter 
+# Based on IPOPT_DIR/Ipopt/examples/Cpp_example/Makefile
```
pelesh, line 5: This should be obsoleted once HiOp CMake exports targets. See #236.

src/Drivers/nlpPriDec_ex8.cpp
```diff
@@ -0,0 +1,289 @@
...
+
+
+bool Ex8::get_vars_info(const long long& n, double *xlow, double* xupp, NonlinearityType* type)
+{
+  //assert(n>=4 && "number of variables should be greater than 4 for this example");
+  assert(n==ns);
+  //x
+  for(int i=0; i<ns; ++i) xlow[i] = 0.;
+  for(int i=0; i<ns; ++i) xupp[i] = +1e+20;
+  for(int i=0; i<ns; ++i) type[i]=hiopNonlinear;
```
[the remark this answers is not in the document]
pelesh: We should really start using clang-format.

Shout out to @ashermancinelli.

src/Drivers/nlpPriDec_ex8.hpp
```diff
@@ -0,0 +1,164 @@
...
+ */
+
+
+using namespace hiop;
+class Ex8 : public hiop::hiopInterfaceDenseConstraints
+{
+public:
+  Ex8(int ns_);
+  
+  Ex8(int ns_, int S_);
```
pelesh, line 46: Consider defining:
```suggestion
public:
  using recourse_evaluator = hiopInterfacePriDecProblem::RecourseApproxEvaluator;

  Ex8(int ns_);
  Ex8(int ns_, int S_);
```

src/Drivers/nlpPriDec_ex8.hpp
```diff
@@ -0,0 +1,164 @@
...
+  virtual bool get_MPI_comm(MPI_Comm& comm_out);
+  virtual bool eval_Jac_cons(const long long& n, const long long& m,
+	                     const long long& num_cons, const long long* idx_cons,  
+			     const double* x, bool new_x, double* Jac); 
+
+  virtual bool quad_is_defined();
+
+  virtual bool set_quadratic_terms(const int& n, 
+		                   hiopInterfacePriDecProblem::
+				         RecourseApproxEvaluator* evaluator);
```
pelesh, line 93: Defining type alias could make the rest of the class code more compact:
```suggestion
  virtual bool set_quadratic_terms(const int& n, 
		                     recourse_evaluator* evaluator);
```

src/Drivers/nlpPriDec_ex8_driver.cpp
```diff
@@ -0,0 +1,179 @@
...
+  bool selfCheck;
+  
+  if(!parse_arguments(argc, argv, selfCheck, nx, S)) {
+    usage(argv[0]);
+    return 1;
+  }
+
+  //PriDecMasterProblemEx8 pridec_problem(nx, S,MPI_COMM_WORLD);
+  PriDecMasterProblemEx8 pridec_problem(nx, S, nc,MPI_COMM_WORLD);
+  //printf("total ranks %d\n",comm_size);
```
pelesh, line 134: I suggest removing commented out code before the merge.
```suggestion
  PriDecMasterProblemEx8 pridec_problem(nx, S, nc, MPI_COMM_WORLD);
```

src/Drivers/nlpPriDec_ex9.cpp
```diff
@@ -0,0 +1,198 @@
...
+  solver.getSolution(y_);
+  
+  #ifdef HIOP_USE_MPI
+  // uncomment if want to monitor contingency computing time
+  /* t4 =  MPI_Wtime(); 
+     if(idx==0||idx==1) {
+       printf( "Elapsed time for contingency %d is %f\n",idx, t4 - t3 ); 
+       printf(" Objective for idx %d value %18.12e, xi %18.12e\n",idx,rval,xi[0]);
+     }
+  */
```
pelesh, line 163: Perhaps better solution would be to add a command line option to turn on contingency compute time monitoring.

src/Drivers/nlpPriDec_ex9_driver.cpp
```diff
@@ -0,0 +1,188 @@
+#include "nlpPriDec_ex9.hpp"
+//the solver
+#include "hiopAlgPrimalDecomp.hpp"
+
+#ifdef HIOP_USE_MAGMA
+#include "magma_v2.h"
+#endif
```
pelesh, line 7: Perhaps this should check if GPU is used?

src/Drivers/nlpPriDec_ex9_user_basecase.hpp
```diff
@@ -0,0 +1,99 @@
...
+                      const int& nnzHSS, int* iHSS, int* jHSS, double* MHSS)
+  {
+    if(!Ex6::eval_Hess_Lagr(n, m, x, new_x, obj_factor, lambda, new_lambda,
+                            nnzHSS, iHSS, jHSS, MHSS)) {
+      return false;
+    }
+    //add diagonal to the Hessian
+    if(iHSS!=nullptr && jHSS!=nullptr) {
+      //nothing to do as the parent already added (i,i) corresponding to the diagonal
+    }
```
pelesh, line 56: Is there a need to keep this `if` statement in the code?

src/Drivers/nlpPriDec_ex9_user_recourse.hpp
```diff
@@ -0,0 +1,475 @@
...
+    }
+
+    /*if(idx_<=10) {
+      for(int i=0;i<100;i++) {
+        clow[i] = 0.;
+        cupp[i] = 0.;
+      }
+    }*/
+
+    //clow[ny_-1] = 1.;
```
pelesh, line 146: Perhaps remove the commented out code?

src/Drivers/nlpPriDec_ex9_user_recourse.hpp
```diff
@@ -0,0 +1,475 @@
...
+        for(int i=1;i<nS_;i++) {
+          cons[m-1] += (x[i] + xi_[i])*(x[i] + xi_[i]);
+        }
+        for(int i=nS_;i<nx_;i++) {
+          cons[m-1] += x[i]*x[i];
+        }
+      }
+      /*if(idx_<=10 && con_idx<100) {
+        cons[con_idx] = x[con_idx]-x_[idx];
+      }*/
```
pelesh, line 217: Can we remove this?

src/Drivers/nlpPriDec_ex9_user_recourse.hpp
```diff
@@ -0,0 +1,475 @@
...
+            for(int i=nS_;i<nsparse_;i++) {
+              iJacS[nnzit] = m-1;
+              jJacS[nnzit] = i;
+              nnzit++;
+              //cons[m-1] += x[i]*x[i];
+            }
+          }
+          //sparse Jacobian ineq w.r.t x and s
+        }
+      }
```
pelesh, line 299: Consider rewriting this function in a thread-safe way. That will simplify OpenMP or CUDA implementations.

src/Drivers/nlpSparse_ex9.hpp
```diff
@@ -0,0 +1,432 @@
...
+    if(nx>3)
+      n_cons += nx-3;
+    nS_ = int(nx_/2);
+    xi_ = new double[S_*nS_];
+    x0_ = new double[nx_];
+    for(int i=0;i<nS_*S_;i++) xi_[i] = 1.0;
+    for(int i=0;i<nx_;i++) x0_[i] = 1.0;
+    //for(int i=0;i<S_;i++){
+    //  xi_[i*nS_+nS_-1] = -1.0;
+    //}
```
pelesh, line 58: Remove the commented out code.

src/Drivers/nlpSparse_ex9.hpp
```diff
@@ -0,0 +1,432 @@
...
+    if(nx>3)
+      n_cons += nx-3;
+    xi_ = new double[nS_*S_];
+    x0_ = new double[nx_];
+    for(int i=0;i<nS_*S_;i++) { 
+      xi_[i] = 1.0;
+    }
+    //for(int i=0;i<S_;i++) {
+    //  xi_[i*nS_+nS_-1] = 1.0;
+    //}
```
pelesh, line 74: Ditto.

src/Drivers/nlpSparse_ex9.hpp
```diff
@@ -0,0 +1,432 @@
...
+  bool get_sparse_blocks_info(int& n,
+                              int& nnz_sparse_Jaceq, 
+                              int& nnz_sparse_Jacineq,
+                              int& nnz_sparse_Hess_Lagr)
+  {
+    n = n_vars;
+    nnz_sparse_Jaceq = 2;
+    nnz_sparse_Jacineq = 2+2*(nx_-3)+S_*(nx_+ 2*(nx_-1)) ;
+    //printf("nnz_sparse_Jacineq %d\n",nnz_sparse_Jacineq);
+    //nnz_sparse_Hess_Lagr = nx_+S_*nx_*2+S_*nx_; //this variable should always be <= n_vars?
```
pelesh, line 136: Remove commented out code.

src/Drivers/nlpSparse_ex9.hpp
```diff
@@ -0,0 +1,432 @@
...
+    }
+    return true;
+  }
+
+  bool eval_grad_f(const long long& n, const double* x, bool new_x, double* gradf)
+  {
+    assert(n==n_vars);
+    for(int i=0;i<n_vars;i++) gradf[i] = 0.;
+    //for(int i=0;i<nx_;i++) gradf[i] = std::pow(x[i]-1.,3);
+    for(int i=0;i<nx_;i++) gradf[i] = 1.0*std::pow(x[i]-1.,3);
```
pelesh, line 162: Try:
```suggestion
    for(int i=0;i<nx_;i++)
      gradf[i] = 1.0*std::pow(x[i]-1.,3);
```

src/Drivers/nlpSparse_ex9.hpp
```diff
@@ -0,0 +1,432 @@
...
+      assert(nnzit == nnzJacS);
+    }
+    return true;
+  }
+
+  /*bool eval_Hess_Lagr(const long long& n, const long long& m,
+                           const double* x, bool new_x, const double& obj_factor,
+                           const double* lambda, bool new_lambda,
+                           const int& nnzHSS, int* iHSS, int* jHSS, double* MHSS)
+  {return false;}*/
```
pelesh, line 335: Perhaps use doxygen style comments to document this interface?

src/Drivers/nlpSparse_ex9.hpp
```diff
@@ -0,0 +1,432 @@
...
+    assert(nnzHSS == nx_+S_*nx_+S_*nx_);
+    int nnzit = 0;
+    if(iHSS!=NULL && jHSS!=NULL) {
+      for(int i=0; i<nx_; i++) { 	
+        iHSS[nnzit] = jHSS[nnzit] = i;
+        nnzit++;
+        //for(int j=0;j<S_;j++){
+        //  iHSS[nnzit] = i; jHSS[nnzit] = nx_+j*nx_+i;
+        //  nnzit++;
+        //}
```
pelesh, line 358: Remove commented code.

src/Drivers/nlpSparse_ex9.hpp
```diff
@@ -0,0 +1,432 @@
...
+          nnzit++;
+          MHSS[nnzit] = obj_factor/double(S_)+lambda[(nx_-1)+nx_*i+nx_-1]*2.;
+          nnzit++;
+        }
+      }
+      assert(nnzHSS == nnzit);
+      //for(int i=0;i<nnzHSS;i++) {
+      //  std::cout<<"MHSS "<<i<<" "<<MHSS[i];
+      //}
+      //std::cout<<std::endl;
```
pelesh, line 397: Ditto.

src/Drivers/nlpSparse_ex9_driver.cpp
```diff
@@ -0,0 +1,97 @@
...
+  Ex9 nlp_interface(nx,S,nS);
+  hiopNlpSparse nlp(nlp_interface);
+  //nlp.options->SetStringValue("duals_update_type", "linear"); // "lsq" or "linear" --> lsq hasn't been implemented yet.
+                                                            // it will be forced to use "linear" internally.
+  //nlp.options->SetStringValue("duals_init", "lsq"); // "lsq" or "zero"
+  nlp.options->SetStringValue("compute_mode", "cpu");
+  //nlp.options->SetStringValue("compute_mode", "hybrid");
+  nlp.options->SetStringValue("KKTLinsys", "xdycyd");
+  //nlp.options->SetStringValue("KKTLinsys", "full");
+  //nlp.options->SetStringValue("write_kkt", "yes");
```
pelesh, line 60: Remove commented out options and document options for user to play with (if any).

src/Interface/CMakeLists.txt
```diff
@@ -15,5 +19,7 @@ install(
   )
 
 if(HIOP_BUILD_SHARED)
-  add_library(hiopInterface OBJECT ${hiopInterface_SRC})
+  add_library(hiopInterface OBJECT ${hiopInterface_SRC} ${hiopInterface_PD})
+else()
+  add_library(hiopInterface OBJECT ${hiopInterface_PD})
 endif()
```
pelesh, line 25: I am not sure I understand this logic. @CameronRutherford, could you chime in here?

src/Interface/hiopInterface.hpp
```diff
@@ -59,6 +59,10 @@
 #define MPI_Comm int
 #endif
 
+#ifndef MPI_COMM_SELF
+#define MPI_COMM_SELF 0
+#endif
+
 #ifndef MPI_COMM_WORLD
 #define MPI_COMM_WORLD 0
 #endif
```
pelesh, line 68: I believe this is implemented in `hiopMPI.hpp` header file. I suggest including it here rather than replicating similar logic.

src/Interface/hiopInterfacePrimalDecomp.cpp
```diff
@@ -0,0 +1,216 @@
...
+                        const double* rgrad, 
+                        const double* rhess, 
+                        const double* x0)
+  : nc_(nc), S_(S)
+{
+  //assert(S>=nc);
+  rval_ = rval;
+  rgrad_ = new double[nc];
+  rhess_ = new double[nc];
+  x0_ = new double[nc];
```
pelesh, line 48: Shouldn't this be encapsulated in a vector object?

src/Interface/hiopInterfacePrimalDecomp.cpp
```diff
@@ -0,0 +1,216 @@
...
+    
+  xc_idx_.resize(nc_);
+  for(int i=0;i<nc_;i++) xc_idx_[i] = i;
+
+  memcpy(rgrad_,rgrad, nc*sizeof(double));
+  memcpy(rhess_,rhess , nc*sizeof(double));
+  memcpy(x0_,x0, nc*sizeof(double));
```
pelesh, line 55: This is not portable solution.

src/Interface/hiopInterfacePrimalDecomp.cpp
```diff
@@ -0,0 +1,216 @@
...
+  //sum 0.5 {(x_i-1)*(x_{i}-1) : i=1,...,nc} 
+  assert(rgrad_!=NULL);
+  obj_value += rval_;
+
+  for(int i=0; i<nc_; i++) {
+    obj_value += rgrad_[i]*(x[xc_idx_[i]]-x0_[i]);
+  }
+  for(int i=0; i<nc_; i++) {
+    obj_value += 0.5*rhess_[i]*(x[xc_idx_[i]]-x0_[i])*(x[xc_idx_[i]]-x0_[i]);
+  }
```
pelesh, line 100: The code in this implementation is not portable. I will not mark each instance, but rather stop here. It should be fairly straightforward to reimplement this in a portable fashion using existing linear algebra objects in HiOp.

src/Optimization/hiopAlgPrimalDecomp.hpp
```diff
@@ -0,0 +1,262 @@
...
+  double accp_tol_ = 1e-6;
+  //consecutive iteration count where NLP residual is lower than acceptable tolerance
+  int accp_count = 0;
+  //initial alpha_ratio if used
+  double alpha_ratio_=1.0;
+  //real decrease over expected decrease ratio
+  double rhok_ = 0.;
+};
+
+}; //end of namespace
```
pelesh, line 260: Don't need `;` at the end of namespace.

src/Optimization/hiopAlgPrimalDecomp.cpp
```diff
@@ -0,0 +1,1211 @@
...
+
+namespace hiop
+{
+#ifdef HIOP_USE_MPI
+  /* This struct provides the info necessary for the recourse approximation function
+   * buffer[n+1] contains both the function value and gradient w.r.t x
+   * buffer[0] is the function value and buffer[1:n] the gradient
+   * Contains send and receive functionalities for the values in buffer
+   */
+  struct ReqRecourseApprox
```
pelesh, line 16: Consider refactoring this class to contain wrappers for MPI calls and a pointer to buffer packing/unpacking object. Buffer management class should have an abstract interface and should be created using a factory. For this PR it would suffice to have just one buffer manager implementation.

src/Optimization/hiopAlgPrimalDecomp.cpp
```diff
@@ -0,0 +1,1211 @@
...
+      for(int i=0;i<n_;i++) {
+        buffer[i+1]=g[i];
+      }
+    }
+
+    MPI_Request request_;
+  private:
+    int n_;
+    double* buffer;
+  };
```
pelesh, line 56: Consider implementation like:
```suggestion
    int n_;
    MpiBufferManager* bm;
  };
```

src/Optimization/hiopAlgPrimalDecomp.cpp
```diff
@@ -0,0 +1,1211 @@
...
+   * buffer[0] is the function value and buffer[1:n] the gradient
+   * Contains send and receive functionalities for the values in buffer
+   */
+  struct ReqRecourseApprox
+  {
+    ReqRecourseApprox() : ReqRecourseApprox(1) {}
+    ReqRecourseApprox(const int& n)
+    {
+      n_=n;
+      buffer=new double[n_+1];
```
pelesh, line 22:
```suggestion
      bm_ = new MpiBufferManager(n);
```

src/Optimization/hiopAlgPrimalDecomp.cpp
```diff
@@ -0,0 +1,1211 @@
...
+    int test() 
+    {
+      int mpi_test_flag; MPI_Status mpi_status;
+      int ierr = MPI_Test(&request_, &mpi_test_flag, &mpi_status);
+      assert(MPI_SUCCESS == ierr);
+      return mpi_test_flag;
+    }
+    void post_recv(int tag, int rank_from, MPI_Comm comm)
+    {
+      int ierr = MPI_Irecv(buffer, n_+1, MPI_DOUBLE, rank_from, tag, comm, &request_);
```
pelesh, line 34: Method `buffer` in `MpiBufferManager` class returns pointer to buffer data.
```suggestion
      int ierr = MPI_Irecv(bm_->buffer, n_+1, MPI_DOUBLE, rank_from, tag, comm, &request_);
```

src/Optimization/hiopAlgPrimalDecomp.cpp
```diff
@@ -0,0 +1,1211 @@
...
+    }
+    double value(){return buffer[0];}
+    void set_value(const double v){buffer[0]=v;}
+    double grad(int i){return buffer[i+1];}
+    void set_grad(const double* g)
+    {
+      for(int i=0;i<n_;i++) {
+        buffer[i+1]=g[i];
+      }
+    }
```
pelesh, line 50: These methods should be implemented in `MpiBufferManagerClass`. It is not obvious where buffer data will reside. AMD GPUs, for example can communicate directly with MPI, so some of the buffers may be created in the device memory.

src/Optimization/hiopAlgPrimalDecomp.cpp
```diff
@@ -0,0 +1,1211 @@
...
+hiopAlgPrimalDecomposition::HessianApprox::
+HessianApprox(const int& n)
+{
+  n_=n;
+  fkm1 = 1e20;
+  fk = 1e20; 
+  xkm1 = new double[n_]; // x at k-1 step, the current step is k
+  skm1 = new double[n_]; // s_{k-1} = x_k - x_{k-1}
+  ykm1 = new double[n_]; // y_{k-1} = g_k - g_{k-1}
+  gkm1 = new double[n_]; // g_{k-1}
```
pelesh, line 110: Consider encapsulating this data in a linear algebra object. This is not portable implementation.

src/Optimization/hiopAlgPrimalDecomp.cpp
```diff
@@ -0,0 +1,1211 @@
...
+    }
+      
+    bool bret;
+    int rank_master=0; //master rank is also the solver rank
+    //Define the values and gradients as needed in the master rank
+    double rval = 0.;
+    double grad_r[nc_];
+    for(int i=0; i<nc_; i++) {
+      grad_r[i] = 0.;
+    }
```
pelesh, line 582: Even though the assumption is the master will run on host, the algorithm should be implemented in a generic fashion, not accessing raw data directly.

src/Optimization/hiopAlgPrimalDecomp.cpp
pelesh: Also, probably good idea would be to put `ReqRecourseApprox` class implementation in a separate source file.

src/Optimization/hiopAlgPrimalDecomp.cpp
```diff
@@ -0,0 +1,1211 @@
...
+    MPI_Request request_;
+  private:
+    int idx;
+  };
+#endif
+
+
+
+hiopAlgPrimalDecomposition::HessianApprox::
+HessianApprox() :
```
pelesh, line 96: I would avoid using nested classes. Better have a standalone implementation of Hessian approximation and then use composition with `hiopAlgPrimalDecomposition` class.

pelesh (changes requested):
The PR looks good. Several recommendations:
- Avoid using nested classes. Prefer composition instead.
- Classes `RecourseApproxEvaluator` and `hiopAlgPrimalDecomp` should be refactored to implement algorithms in generic form. Access to raw data should be through an abstract interface to allow for a portable implementation in the future. Even if the intent is to have these classes and their data live on the host, a generic implementation will increase this code's shelf life.
- It would be good to provide more comments in the code, so that the design intent is more clear.
- Some minor code formatting issues need to be corrected. It is probably a good time to consider using clang-format.

Addressing these should not take long. Most of the code is in a good shape.

src/Interface/CMakeLists.txt
jwang125: chiopInterface.cpp is only needed when HIOP_BUILD_SHARED is on. For pridcomp algorithms we added a cpp file for the hiopInterfacePrimalDecomp class that needs to be added to the library at all time. But add_library can only be called once for one unique library name 'hiopInterface'. So I added the else there.

src/Interface/hiopInterface.hpp
jwang125: Cool. Didn't know that. But it seems like MPI_COMM_WORLD is missing in the hiopMPI.hpp. Shall we add it to that file? Thanks.

src/Interface/hiopInterfacePrimalDecomp.cpp
jwang125: I have no opinion on this. But the protected variable 'rgrad_' inside the class 'RecourseApproxEvaluator' is of type double*. If we want to change it we need to change the protected member types.

src/Interface/hiopInterfacePrimalDecomp.cpp
jwang125: OK. Will do.

jwang125: Just to make sure I understand, should I for example use LinearAlgebraFactory::createVector(n) for all the double * ? And all the protected/private variables should be hiopVector instead of double*? Thanks.

src/Interface/hiopInterfacePrimalDecomp.cpp
jwang125: Variables are now being converted to hiopVector type.

src/Optimization/hiopAlgPrimalDecomp.cpp
jwang125: All these four double arrays are now hiopVectors.

src/Drivers/nlpPriDec_ex8.cpp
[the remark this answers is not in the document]
ashermancinelli: I'd be happy to write some scripts to take care of this, but it's up to @cnpetra. I'm sure he'd also want to write the clang-format file as well.

src/Drivers/nlpPriDec_ex9_driver.cpp
ashermancinelli: I think magma variable is set based on HIOP_USE_CUDA/GPU, so iiuc this should be good

src/Interface/CMakeLists.txt
ashermancinelli: I don't quite understand this either. Don't we just want to append `${hiopInterface_PD}` to `hiopInterface_SRC`?

src/Interface/hiopInterfacePrimalDecomp.cpp
pelesh: I think the most important part is separating platform specific code from the algorithm implementation. Platform specific code should be accessed through an abstract interface.

I think encapsulating data arrays in `hiopVector` objects is a good idea and vectors should be created using factory. We may need to update the `LinearAlgebraFactory` to meet requirements of the primal decomposition module.

src/Drivers/nlpPriDec_ex9_user_basecase.hpp
jwang125: No. Will be removed.

src/Drivers/nlpPriDec_ex9_user_recourse.hpp
jwang125: Yes.

src/Interface/CMakeLists.txt
jwang125: Sure. We can move the if statement to when hiopInterface_SRC is declared.

src/Interface/hiopInterfacePrimalDecomp.cpp
jwang125: Sounds good. Doing the conversion to hiopVector now.

src/Interface/CMakeLists.txt
CameronRutherford: Something along the lines of the following would also work:
```Cmake
add_library(hiopInterface OBJECT 
${hiopInterface_SRC}
$<$<BOOL:HIOP_BUILD_SHARED>:${hiopInterface_PD>)
```
I do think conditionally appending to `${hiopInterface_SRC}` is a better practice (the genexp can be much harder to read...), but @pelesh wanted me to chime in so I wanted to give another solution.

src/Drivers/nlpPriDec_ex9_driver.cpp
CameronRutherford: +1 to this being ok as it is clearer in context. Having a CMake variable for each TPL means the logic for them can always be taken care of in a `CMakeLists.txt` somewhere else, although functionally speaking I think it would be equivalent to just use `HIOP_USE_GPU` here.

jwang125: This has been changed as suggested.

src/Drivers/nlpPriDec_ex9_user_recourse.hpp
jwang125: Done.

src/Drivers/nlpSparse_ex9.hpp
jwang125: Done.

src/Drivers/nlpSparse_ex9.hpp
jwang125: Done.

src/Drivers/nlpSparse_ex9.hpp
jwang125: Done.

src/Drivers/nlpSparse_ex9.hpp
jwang125: Done.

src/Drivers/nlpSparse_ex9.hpp
jwang125: This function is not needed and has been deleted.

src/Drivers/nlpSparse_ex9.hpp
jwang125: Done.

src/Drivers/nlpSparse_ex9.hpp
jwang125: Thanks. Done.

src/Interface/CMakeLists.txt
jwang125: OK. Since Cameron said better to add to hiopInterface_SRC, that's what I did. Thanks.

src/Interface/hiopInterfacePrimalDecomp.cpp
jwang125: This part has been done updated with copyFrom function and rgrad, rhess, x0 are now hiopVectors.

src/Interface/hiopInterfacePrimalDecomp.cpp
jwang125: The only part not converted yet is the x[xc_idx_[i]] part, which access x through a list of integer vector. We should discuss this further.

src/Optimization/hiopAlgPrimalDecomp.hpp
jwang125: Done.

src/Optimization/hiopAlgPrimalDecomp.cpp
jwang125: Hoping to address this in discussion this week.

src/Optimization/hiopAlgPrimalDecomp.cpp
jwang125: Yes. This now uses hiopVector and calls setToZero().

src/Optimization/hiopAlgPrimalDecomp.cpp
jwang125: Hoping to discuss this issue later this week.

src/Drivers/nlpPriDec_ex9_user_recourse.hpp
jwang125: Hoping to discuss this in detail later this week.

src/Drivers/nlpSparse_ex9_driver.cpp
jwang125: For the sake of the problem, no options are currently needed. Some of them were used during debugging. I'm not sure how many we want to keep.

src/Optimization/hiopAlgPrimalDecomp.cpp
jwang125: Same as the issue above.

src/CMakeLists.txt
```diff
@@ -23,6 +23,7 @@ foreach(libtype shared static)
       $<TARGET_OBJECTS:hiopLinAlg>
       $<TARGET_OBJECTS:hiopOptimization>
       $<TARGET_OBJECTS:hiopUtils>
+      $<TARGET_OBJECTS:hiopInterface>
```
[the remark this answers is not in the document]
CameronRutherford: You should only have to add the `hiopInterface` object once in this `add_library` command. Since the object is build regardless of shared/static libs, I think you should just be able to remove `$<$<BOOL:${HIOP_BUILD_SHARED}>:$<TARGET_OBJECTS:hiopInterface>>` entirely @nychiang

src/Drivers/nlpPriDec_ex9_driver.cpp
```diff
@@ -0,0 +1,188 @@
...
+    }
+  case 2: //1 argument
+    {
+      if(std::string(argv[1]) == "-selfcheck")
+      {
+        self_check=true;
+      } else {
+        nx = atoi(argv[1]);
+        if(nx<=0) return false;
+      }
```
pelesh, line 81: We need clang-format :).

src/Drivers/nlpSparse_ex9_driver.cpp
pelesh: @jwang125, it is your call how many options to keep. The `compute_mode` one is I would suggest to keep in all examples as we are working with more and more GPU code.

src/Interface/hiopInterfacePrimalDecomp.cpp
pelesh: @jwang125, you are right. We don't have a good portable example of re-mapping vectors in HiOp code.

pelesh:
There are several implicit conversions from signed to unsigned `int` in the primal decomposition example 9:

```shell
/.../src/hiop/hiop/src/Drivers/nlpPriDec_ex9.cpp: In constructor ‘PriDecMasterProblemEx9::PriDecMasterProblemEx9(size_t, size_t, size_t, size_t)’:
/.../src/hiop/hiop/src/Drivers/nlpPriDec_ex9.cpp:19:44: warning: conversion to ‘int’ from ‘size_t {aka long unsigned int}’ may alter its value [-Wconversion]
   basecase_ = new PriDecBasecaseProblemEx9(nx_);
                                            ^~~
/.../src/hiop/hiop/src/Drivers/nlpPriDec_ex9.cpp: In member function ‘virtual hiop::hiopSolveStatus PriDecMasterProblemEx9::solve_master(hiop::hiopVector&, const bool&, const double&, const double*, const double*)’:
/.../src/hiop/hiop/src/Drivers/nlpPriDec_ex9.cpp:40:46: warning: conversion to ‘int’ from ‘size_t {aka long unsigned int}’ may alter its value [-Wconversion]
     basecase_ = new PriDecBasecaseProblemEx9(nx_);
                                              ^~~
/.../src/hiop/hiop/src/Drivers/nlpPriDec_ex9.cpp: In member function ‘virtual bool PriDecMasterProblemEx9::eval_f_rterm(size_t, const int&, const double*, double&)’:
/.../src/hiop/hiop/src/Drivers/nlpPriDec_ex9.cpp:124:47: warning: conversion to ‘int’ from ‘size_t {aka long unsigned int}’ may alter its value [-Wconversion]
   ex9_recourse = new PriDecRecourseProblemEx9(nc_, nS_,S_,x,xi);
                                               ^~~
/.../src/hiop/hiop/src/Drivers/nlpPriDec_ex9.cpp:124:52: warning: conversion to ‘int’ from ‘size_t {aka long unsigned int}’ may alter its value [-Wconversion]
   ex9_recourse = new PriDecRecourseProblemEx9(nc_, nS_,S_,x,xi);
                                                    ^~~
/.../src/hiop/hiop/src/Drivers/nlpPriDec_ex9.cpp:124:56: warning: conversion to ‘int’ from ‘size_t {aka long unsigned int}’ may alter its value [-Wconversion]
   ex9_recourse = new PriDecRecourseProblemEx9(nc_, nS_,S_,x,xi);
                                                        ^~
/.../src/hiop/hiop/src/Drivers/nlpPriDec_ex9.cpp: In member function ‘virtual bool PriDecMasterProblemEx9::eval_grad_rterm(size_t, const int&, double*, hiop::hiopVector&)’:
/.../src/hiop/hiop/src/Drivers/nlpPriDec_ex9.cpp:177:32: warning: conversion to ‘double’ from ‘size_t {aka long unsigned int}’ may alter its value [-Wconversion]
     grad_vec[i] = (x[i]-y_[i])/S_;
```
It would be good to fix these warnings before the merge.

src/Interface/hiopInterface.hpp
pelesh: @cnpetra suggested using `MPI_COMM_SELF` instead. That is why we don't have definition of `MPI_COMM_WORLD` in the `hiopMPI.hpp`.

pelesh:
Another issue to consider (perhaps out of this PR scope): Static implementation of the linear algebra factory will move all linear algebra objects data to the same memory space. UVM could be used as a transitional solution, but eventually we would want to keep some linear algebra objects specifically on the host and some on device.

Perhaps we should prioritize addressing #131 and reimplement factory as non-static class.

src/Drivers/nlpPriDec_ex9_user_recourse.hpp
pelesh: To be addressed in a separate PR.

src/Optimization/hiopAlgPrimalDecomp.cpp
pelesh: Per discussion, use `hiopVector` as a container for buffer data.

src/Optimization/hiopAlgPrimalDecomp.cpp
pelesh: We agreed the nested class approach is appropriate here.

src/Drivers/nlpPriDec_ex9_driver.cpp
ashermancinelli: @cnpetra just say the word :)

ashermancinelli:
> Perhaps we should prioritize addressing #131 and reimplement factory as non-static class.

I don't think implementing this will be too big a lift, but changing all the code that uses `LinAlgFactory::createVector(n)` into `this->linAlgFactory_.createVector(n)` might be.

nychiang:
The two new copyfrom functions can be replaced by the existing functions:
1. [copy_from_starting_at](https://github.com/LLNL/hiop/blob/467c9505db4e6fd48ba73649d017d9deff2a5b1a/src/LinAlg/hiopVectorPar.cpp#L175)
2. [startingAtCopyFromStartingAt](https://github.com/LLNL/hiop/blob/467c9505db4e6fd48ba73649d017d9deff2a5b1a/src/LinAlg/hiopVectorPar.cpp#L180)
3. [copyToStarting](https://github.com/LLNL/hiop/blob/467c9505db4e6fd48ba73649d017d9deff2a5b1a/src/LinAlg/hiopVectorPar.cpp#L202)

tests/LinAlg/vectorTestsRajaPar.cpp
```diff
@@ -132,6 +132,30 @@ real_type* VectorTestsRajaPar::createLocalBuffer(local_ordinal_type N, real_type
...
+  for(local_ordinal_type i = 0; i < N; ++i)
+    buffer[i] = val;
+  buffer[N-1] = 0;
+#ifdef HIOP_USE_GPU
+  umpire::Allocator dal = resmgr.getAllocator("DEVICE");
+  local_ordinal_type* dev_buffer = static_cast<local_ordinal_type*>(
+		                   dal.allocate(N*sizeof(local_ordinal_type)));
+  resmgr.copy(dev_buffer, buffer, N*sizeof(local_ordinal_type));
+  hal.deallocate(buffer);
+  return dev_buffer;
```
pelesh, line 152: I really like `hal` and `dal` names for host and device allocators. We should make this HiOp-wide standard :).

PR #194, opened 2021-02-17
Styleguide updates

cnpetra:
This PR establishes mandatory code styleguide for all C++ files of HiOp. 

It goes in effect immediately, but it is still open to new suggestions and/or strong objections.

### DIFF
--- a/README_developers.md
+++ b/README_developers.md
@@ -183,6 +183,17 @@ for(some_condition){
   value += 1;
 }
 ```
+### Constructor initialization list
+Use two spaces and then `:` followed by space to align the first member initialization. Use only one initialization per line with a trailing comma and align multiple initialization lines with the first declaration (essentially at four spaces).
+
+```c++
+MyClass::MyClass()
+  : member1_(nullptr),
+    other_member2_(0)
+{
+
+}
+```
 
 ## Braces
 
@@ -364,9 +375,11 @@ int c;
 int a, b, c;
 ```
 
-### Long declaration lines
-When breaking methods declaration into multiple lines:
-- have a limit of ~125 characters per line
+### Breaking down long lines
+
+Developers should have in mind a limit of ~125 characters per line. 
+
+Here are guidelines for breaking function/method declaration or calls into multiple lines:
 - keep the first argument on the same line with the method name
 - align arguments on the subsequent lines with the first argument
 - if class name or return type long, place the function name at the beginning of the line and put the return type and class name qualification on the previous line (see example below)
@@ -394,12 +407,29 @@ some_method(double beta,
 {
 }
 
+//Yes: function call may need to be broken into multiple line
+object1.method1(param1, param2).method2(param).some_method(beta,
+                                                           someobject.get_W(),
+                                                           alpha,
+                                                           X);
+                                                           
+//Yes: alternative to the above 
+object1.method1(param1, param2).method2(param).
+  some_method(beta, someobject.get_W(), alpha, X);
+                                                           
+                                                           
+                                                           
+
 // No: multiple arguments per line when using more than one line for the declaration 
 void ClassName::some_method(double beta, hiopMatrix& W,
                             double alpha, const hiopMatrix& X) const
 {
 }
 
+// No: multiple arguments per line when calling
+object1.method1(param1).method2(param).some_method(beta, someobject.get_W(),
+                                                   alpha, X);
+                                                           
 // No: arguments not aligned behind the opening parenthesis
 void ClassName::some_method(
   double beta,
@@ -408,8 +438,18 @@ void ClassName::some_method(
   const hiopMatrix& X) 
 {
 }
-```
 
+// everything above applies to constructors as well, for example,
+SomeVeryLongClassNameIMeanLong::
+SomeVeryLongClassNameIMeanLong(double beta,
+                               hiopMatrix& W,
+                               double alpha,
+                               const hiopMatrix& X)
+  : member1_(0),
+    member2_(nullptr)
+{
+}
+```
 
 ## Comments
 - Don't comment what can be expressed in code
@@ -423,6 +463,10 @@ there should be description of method arguments, template parameters (if
 applicable), return value, pre- and post-conditions. Use template as below.
 Input and output arguments should be clearly marked as such (e.g. `@param[in]`
 for inputs).
+
+Detailed comments should go in the implementation file (but not prohibited from appearing in the header file). 
+
+Documentation of pre and post conditions should appear in the interface or header file (but not prohibited from appearing in the implementation file). 
 
 ```c++
 /**
--- a/README_developers.md
+++ b/README_developers.md
@@ -1,75 +1,3 @@
-# Minimal testing procedure 
-Any PR or push to the master should go through and pass the procedures below. The shell commands need to be ran in the 'build' directory and assume bash shell. 
-
-## 1. All tests in 'make test' pass for each of these builds, with every combination of `HIOP_USE_RAJA` and `HIOP_USE_GPU`.
-1. a. MPI=ON, DEEPCHECKS=OFF, RELEASE=ON (this is the high-performance version of HiOp)
-```shell
-$> rm -rf *; cmake -DHIOP_USE_MPI=ON -DHIOP_DEEPCHECKS=OFF -DCMAKE_BUILD_TYPE=RELEASE ..
-$> make -j4; make install; make test
-```
-1. b. MPI=OFF, DEEPCHECKS=OFF, RELEASE=ON
-```shell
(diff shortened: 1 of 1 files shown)


README_developers.md
```diff
@@ -91,128 +85,356 @@ $ # For a fix
...
+ * @param <NAME> <DESCRIPTION>
+ * @tparam <NAME> <DESCRIPTION>
+ * @return <DESCRIPTION OF RETURN VALUE>
+ *
+ * @pre <PRECONDITION>
+ * @post <POSTCONDITION>
+ *
+ * LONGER DESCRIPTION, RUNTIME, EXAMPLES, ETC 
+ */
+```
```
pelesh, line 440: I would perhaps just add that detailed commens should be next to the implementation rather than in the header file. If you make changes in the implementation you are less likely to forget to update comments if they are right there.

pelesh (approved):
This is consistent and covers well what we have discussed.

CameronRutherford (approved):

README_developers.md
```diff
@@ -91,128 +85,356 @@ $ # For a fix
...
+                            const hiopMatrix& X) const
+{
+}
+
+// Yes: when return type + class name + method name + first argument would overrun the 125 characters line limit, have method name on the next line
+long_return_type SomeVeryLongClassNameIMeanLong::
+some_method(double beta,
+            hiopMatrix& W,
+            double alpha,
+            const hiopMatrix& X) const
```
ashermancinelli, line 393: Do we have a policy for inheritance lists and constructor initialization of member variables? These tend to run long as well. I recommend leading commas (kind of a boost-ism) if you feel it's worthwhile to mention.

README_developers.md
ashermancinelli: I might suggest pre and post conditions remain next to the interface - that seems important for the users to know about, no?

README_developers.md
```diff
@@ -91,128 +85,356 @@ $ # For a fix
...
+// No: opening brace on the same line
+bool MyClass::my_method(int a) {
+
+}
+``` 
+
+## RAJA specific style
+- Indent two spaces above surrounding scope
+- No spaces around template parameter
+- Line break before range segment
```
ashermancinelli, line 281: Every policy in the document besides this one would be relatively easy to automate - if you're sure about this one we can leave it but tools may have a harder time enforcing.

ashermancinelli (approved):
Lgtm! Do we intend to rewrite all of the codebase under this style, or just enforce it going forward?

README_developers.md
jwang125: I have the same question. 
hiopAverylong::
hiopAverylong(typeA a,
                       typeB b)
: nameacanbeverylong_(a),b_(b)

jwang125 (approved):

jwang125:
Correct me if I missed this please, but when calling a function with many arguments, do they need to be on a separate line each like when the function is declared? Thank you.

README_developers.md
nychiang: Do we need to add brief descriptions in the header file?
For inherited classes, do we need to add the same description to all the inherited functions?

README_developers.md
```diff
@@ -91,128 +85,356 @@ $ # For a fix
...
-  - Use `hiop_*_policy` over directly using cuda or omp policies
-  - Prefer `RAJA::Index_type` over `int` or anything else
-- Prefer braces for every block
-- Prefer no indentation for `private`, `public`, and `protected` statements
-- Each variable declaration should have it's own type declaration
+private:
+  int memeber_var_; 
+}
+```
+## Spacing
```
nychiang, line 142: I think it's better to mention no trailing whitespace.

nychiang (approved):

README_developers.md
```diff
@@ -91,128 +85,356 @@ $ # For a fix
...
+### Classes
 
-For example:
+Use Pascal case for class names (upper and lower case, no underscores).
+When the name is composite of multiple words, each word starts with a capital
+letter. If one of the "words" is an acronym use same capitalization (first
+capital and other letters lowercase).
+```c++
+class MyHiopClass;
+```
```
ashermancinelli, line 310: Does this mean we're going to move all classes to pascal case immediately (eg `hiopMatrixDense` -> `HiopMatrixDense`)? What's our timeline/plan to convert everything?

pelesh: That is a good point. I would pass the buck to @cnpetra here. My $0.02 would be to drop prefix `hiop` altogether from all HiOp class names because they are already in a `hiop` namespace.

Alternatively, we can just change the class names guideline so that each class name must begin with lowercase `hiop` like this:
```c++
class hiopMyHiopClass;
```

pelesh:
> Correct me if I missed this please, but when calling a function with many arguments, do they need to be on a separate line each like when the function is declared? Thank you.

I think @cnpetra's point is if the function declaration _or_ call is longer than 125 characters, then one should put each function argument on a separate line. If the function call can fit within 125 characters with all function arguments on the same line, then one should keep the call on the single line.

ashermancinelli:
@cnpetra do we plan on rewriting all the legacy hiop code that uses old conventions? If so, how do we plan on going about it?

I wrote a quick clang tool to identify many of these changes and suggest replacements. When run on the following block:
```cpp
class hiop_matrixDense {
  int MyMemberVariable;
};
```

it warns:

```console
$ ./src/syntax/SyntaxCheckerStandalone --strip=hiop ../test/hiop.cpp
../test/hiop.cpp:2:1: warning: Found non-pascal case class/struct/union name. Please consider renaming.
class hiop_matrixDense {
^     ~~~~~~~~~~~~~~~~
      MatrixDense
../test/hiop.cpp:3:3: warning: Found non-snake case class/struct/union member variable name. Please consider renaming.
  int MyMemberVariable;
  ^   ~~~~~~~~~~~~~~~~
      my_member_variable_
2 warnings generated.
```

It can auto-write the changes, but that seems a bit risky. Let me know what you think. cc @pelesh

README_developers.md
cnpetra: > I might suggest pre and post conditions remain next to the interface - that seems important for the users to know about, no?

I agree

cnpetra: > Do we need to add brief descriptions in the header file?
> For inherited classes, do we need to add the same description to all the inherited functions?

I think that would be good practice.

cnpetra: I've updated the doc with the above.

README_developers.md
cnpetra: i've update the guideline, take a look and let me know

README_developers.md
cnpetra: you mean "Line break before range segment" ?

cnpetra:
> Correct me if I missed this please, but when calling a function with many arguments, do they need to be on a separate line each like when the function is declared? Thank you.

I've updated the guidelines.

README_developers.md
cnpetra: as in
```c++
int member_var_  ;
```
?

README_developers.md
cnpetra: > @cnpetra do we plan on rewriting all the legacy hiop code that uses old conventions? If so, how do we plan on going about it?
> 
> I wrote a quick clang tool to identify many of these changes and suggest replacements. When run on the following block:
> 
> ```c++
> class hiop_matrixDense {
>   int MyMemberVariable;
> };
> ```
> 
> it warns:
> 
> ```
> $ ./src/syntax/SyntaxCheckerStandalone --strip=hiop ../test/hiop.cpp
> ../test/hiop.cpp:2:1: warning: Found non-pascal case class/struct/union name. Please consider renaming.
> class hiop_matrixDense {
> ^     ~~~~~~~~~~~~~~~~
>       MatrixDense
> ../test/hiop.cpp:3:3: warning: Found non-snake case class/struct/union member variable name. Please consider renaming.
>   int MyMemberVariable;
>   ^   ~~~~~~~~~~~~~~~~
>       my_member_variable_
> 2 warnings generated.
> ```
> 
> It can auto-write the changes, but that seems a bit risky. Let me know what you think. cc @pelesh

this would be nice. It would be even nicer if we can run on the diff when one commits or when one does a PR.

While I think we should enforce the guidelines on all the new code, I am not sure if it will pay off to change all the code, given the risks and the fact that the code is kinda follows the guideline already...

cnpetra:
I've tried to address all the issues raised and I have re-requested for your review.
one last round of democracy before dictatorship :)

jwang125 (approved):

README_developers.md
nychiang: For example
int member_var_;xxx
where xxx are the extra whitespace.
We have many this kind of whitespace in the current code.

cnpetra: I see, good suggestion, I've updated the doc

nychiang (approved):

ashermancinelli (approved):

pelesh (approved):

CameronRutherford (approved):

PR #195, opened 2021-02-18
Update of write_kkt

cnpetra:
Updates the CSR file IO code to write NLP sizes.

Also the documentation is updated: 
 - lower triangle elements are saved for sparse symmetric KKT linearization
 - upper triangle elems are saved for other symmetric KKT linearizations

### DIFF
--- a/src/LinAlg/load_kkt_mat.m
+++ b/src/LinAlg/load_kkt_mat.m
@@ -6,12 +6,15 @@ fclose(f);
 
 %% data loading
 m=A(1); % matrix size 
-nnz=A(2); % number of nnz of the upper triangle
-rowp = A(3:m+3); % row pointers in colidx and vals (see below)
-colidx=A(m+4:nnz+m+3); % column indexes for each nz
-vals=A(nnz+m+4:nnz+nnz+m+3); % values for each nz
-rhs=A(nnz+nnz+m+4:nnz+nnz+m+m+3); % rhs 
-sol=A(nnz+nnz+m+m+4:nnz+nnz+m+m+m+3); % solution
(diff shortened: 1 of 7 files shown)


cnpetra:
@nychiang I have second thoughts about saving lower triangle for some and upper triangle for others. maybe we should be consistent and always save upper? (and, if needed, convert from lower to upper when saving).

src/LinAlg/load_kkt_mat.m
```diff
@@ -23,9 +26,14 @@
...
-%% form the matrix and fill lower triangle
+%% form the matrix
 M = sparse(rowidx, colidx, vals, m, m);
-M=M+transpose(triu(M,1));
+% if the matrix is known to be symmetric fill, you also need to know whether
+%  A. the lower triangle was saved, in which case use
+% M=M+transpose(tril(M,-1));
+% or
+% B. the upper triangle was saved, in which case use
+% M=M+transpose(triu(M,1));
```
nychiang, line 36: how about we do a simple test to check if the input matrix is upper or lower triangular?
For example, use Matlab function `istril()` and `istriu()`. If it is not triangular, return an error.

nychiang:
> @nychiang I have second thoughts about saving lower triangle for some and upper triangle for others. maybe we should be consistent and always save upper? (and, if needed, convert from lower to upper when saving).

How about let's save the matrix in the triplet format? Matlab loads the sparse matrix via triplet format, and it can be easily converted from lower to upper when saving.

nychiang:
Otherwise, we need to spend more computation efforts to transfer a CSR lower triangular matrix to its CSR upper triangular sibling, and then convert it to a triplet format in Matlab.

PR #198, opened 2021-03-04
New CMake configuration variable: HIOP_GPU_PLATFORM

zackgalbreath:
This new drop-down menu allows users to select between different supported GPU platforms, currently CUDA vs. HIP.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@ include(FortranCInterface)
 FortranCInterface_HEADER(FortranCInterface.hpp MACRO_NAMESPACE "FC_")
 
 option(HIOP_USE_MPI "Build with MPI support" ON)
-option(HIOP_USE_GPU "Build with support for GPUs - Magma and cuda libraries" OFF)
+option(HIOP_USE_GPU "Build with support for GPUs" OFF)
 option(HIOP_TEST_WITH_BSUB "Use `jsrun` instead of `mpirun` commands when running tests" OFF)
 option(HIOP_USE_RAJA   "Build with portability abstraction library RAJA" OFF)
 option(HIOP_DEEPCHECKS "Extra checks and asserts in the code with a high penalty on performance" OFF)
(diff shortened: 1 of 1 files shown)


zackgalbreath:
Addresses #160

pelesh:
Thank you, @zackgalbreath, this is great starting point.  How do you suggest to refactor `FindHiopCudaLibraries` CMake module? With CUDA as only GPU option, we could assume that Magma, RAJA and Umpire were all built with CUDA support. With different GPU platforms we need to find (and preferably test) builds for each of these three libraries on each enabled platform (e.g. Magma::CUDA, Magma::HIP, etc.).

The current CMake code in this PR seems to have only a placeholder for HIP platform. What is the current support in CMake for HIP and ROCm packages?

zackgalbreath:
> How do you suggest to refactor `FindHiopCudaLibraries` CMake module? With CUDA as only GPU option, we could assume that Magma, RAJA and Umpire were all built with CUDA support. With different GPU platforms we need to find (and preferably test) builds for each of these three libraries on each enabled platform (e.g. Magma::CUDA, Magma::HIP, etc.).

I'll take a closer look at what you have so far and provide some suggestions on how to extend this functionality for other GPU platforms.

> The current CMake code in this PR seems to have only a placeholder for HIP platform. What is the current support in CMake for HIP and ROCm packages?

We expect HIP 4.2 (coming out in April) to include the necessary hooks for CMake to treat it as a first-class language.

ashermancinelli:
Thanks for looking into this @zackgalbreath. To extend @pelesh's comment, do you have suggestions for refactoring all of our internal targets and `Find*.cmake` modules?

zackgalbreath:
> How do you suggest to refactor `FindHiopCudaLibraries` CMake module?

I can offer some minor suggestions for improvements to `FindHiopCudaLibraries.cmake`, but in general it looks pretty good!

For instance, using the [FindPackageHandleStandardArgs macro](https://cmake.org/cmake/help/latest/module/FindPackageHandleStandardArgs.html) helps provide a common, expected interface.

Here's a link to our documentation on best practices for writing find modules:
https://cmake.org/cmake/help/latest/manual/cmake-developer.7.html#a-sample-find-module

> With CUDA as only GPU option, we could assume that Magma, RAJA and Umpire were all built with CUDA support. With different GPU platforms we need to find (and preferably test) builds for each of these three libraries on each enabled platform (e.g. Magma::CUDA, Magma::HIP, etc.).


Your hunch here is correct. Ideally Magma would expose this information for you so you could do something like

`find_package(Magma COMPONENT CUDA)`.

But alas, that does not seem to be the case.

As I see it, the options are:

1) Teaching your find module how to query Magma for its supported GPU backends
2) Upstream this improvement (exported components) to magma

Option 1 is definitely the path of least resistance.

pelesh:
Implemented in PR #246. Closing this PR.

PR #202, opened 2021-03-19
Using one norm instead of infinity norm to compute primal feasibilities in line search

nychiang:
Using one norm instead of infinity norm to measure primal infeasibilities.
Only evaluate the primal infeasibilities from the equality constraints (c) and inequality constraints (d), i.e., var/con boundary constraints doesn't contribute the the primal infeasibility computation in line search.

### DIFF
--- a/src/Optimization/hiopResidual.cpp
+++ b/src/Optimization/hiopResidual.cpp
@@ -95,22 +95,25 @@ hiopResidual::~hiopResidual()
   if(rsvu) delete rsvu;
 }
 
-double hiopResidual::computeNlpInfeasInfNorm(const hiopIterate& it, 
+double hiopResidual::compute_nlp_norms(const hiopIterate& it, 
 			       const hiopVector& c, 
 			       const hiopVector& d)
 {
   nlp->runStats.tmSolverInternal.start();
   
-  double nrmInf_infeasib;
+//  double nrmInf_infeasib;
+  double nrmOne_infeasib = 0.;
   long long nx_loc=rx->get_local_size();
   //ryc
   ryc->copyFrom(nlp->get_crhs());
   ryc->axpy(-1.0,c);
-  nrmInf_infeasib = ryc->infnorm_local();
+//  nrmInf_infeasib = ryc->infnorm_local();
+  nrmOne_infeasib += ryc->onenorm_local();
   //ryd
   ryd->copyFrom(*it.d);
   ryd->axpy(-1.0, d);
-  nrmInf_infeasib = fmax(nrmInf_infeasib, ryd->infnorm_local());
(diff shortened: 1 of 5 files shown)


src/Optimization/hiopResidual.cpp
```diff
@@ -320,6 +330,10 @@ int hiopResidual::update(const hiopIterate& it,
   int ierr = MPI_Allreduce(aux, aux_g, 6, MPI_DOUBLE, MPI_MAX, nlp->get_comm()); assert(MPI_SUCCESS==ierr);
   nrmInf_nlp_optim=aux_g[0]; nrmInf_nlp_feasib=aux_g[1]; nrmInf_nlp_complem=aux_g[2];
   nrmInf_bar_optim=aux_g[3]; nrmInf_bar_feasib=aux_g[4]; nrmInf_bar_complem=aux_g[5];
+  
+//  double sum_one_norm = 0.0;
+//  ierr = MPI_Allreduce(&nrmOne_theta, &sum_one_norm, 1, MPI_DOUBLE, MPI_SUM, nlp->get_comm()); assert(MPI_SUCCESS==ierr);
+//  nrmOne_theta = sum_one_norm;
```
nychiang, line 336: Constraints c and d are duplicated in each MPI process, and hence this MPI_Allreduce is not required.
Otherwise this will multiply the value of `nrmOne_theta` by the number of MPI processes.

This brings the issue that the current implementation of [hiopVectorPar::one_norm](https://github.com/LLNL/hiop/blob/master/src/LinAlg/hiopVectorPar.cpp#L304) may not provide the value we want, i.e., the computation is correct for the vector on the distributed `x` space, but wrong on the other spaces where we duplicates the local data.
Should we file another issue about it? Or just leave a comment before the implementation?

cnpetra: if the MPI reduce in `hiopResidual::compute_nlp_norms` does not work for all NLP formulations, we should remove it and rely instead on non-local   ` hiopVectorPar::one_norm` as your implementation of  `hiopResidual::compute_nlp_norms` does right now. 

 ` hiopVectorPar::one_norm` should work for replicated `c` and `d` though ... if it does not, then these `hiopVectorPar` vectors are not created correctly, that are not constructed with MPI_COMM_SELF ... if they are, everything should work fine

src/Optimization/hiopResidual.cpp
```diff
@@ -95,22 +95,25 @@ hiopResidual::~hiopResidual()
   if(rsvu) delete rsvu;
 }
 
-double hiopResidual::computeNlpInfeasInfNorm(const hiopIterate& it, 
+double hiopResidual::compute_nlp_norms(const hiopIterate& it,
```
cnpetra, line 98: wouldn't `compute_nlp_infeasib_onenorm` be a better name?

PR #204, opened 2021-03-26
Update of summit readme 

cnpetra:


ashermancinelli (approved):
Shri recently built our stack on summit, so perhaps he would like to review this as well. Perhaps we can also update this when the hiop package is merged upstream. LGTM.

PR #206, opened 2021-03-29
Release 0.4

cnpetra:
This PR targets `master` that will be tagged as and used for v0.4 release

The salient features of this comprehensive release are

-	Development of a sparse NLP solver and associated sparse NLP interface
-	Update of the mixed dense-sparse NLP solver to support full GPU compute mode
-	Unit testing and documentation were expanded and consolidated
-	Added partial support for device/GPU computations for the linear algebra of sparse NLPs
-	Second-order corrections to the full-Newton search directions
-	Support for the relaxing and adjustment of variables and constraints bounds
-	Implemented gradient-based scaling of the problem 
-	Least-squares initialization and computation of the duals
-	Support for inertia computation and regularization for KKT linearizations systems


nychiang:
@cnpetra I also added second-order corrections to the quasi-Newton method.

cmake/FindHiopMETIS.cmake
```diff
@@ -43,3 +43,5 @@ endif()
 
 set(METIS_INCLUDE_DIR CACHE PATH "Path to metis.h")
 set(METIS_LIBRARY CACHE PATH "Path to metis library")
+#set(METIS_LIBRARIES CACHE PATH "Path to metis library")
```
nychiang, line 46: Should we remove this line?

cmake/FindHiopSTRUMPACK.cmake
```diff
@@ -0,0 +1,42 @@
...
+  if(SCALAPACK_LIBRARY)
+    target_link_libraries(STRUMPACK INTERFACE ${SCALAPACK_LIBRARY})
+    message(STATUS "Found SCALAPACK library: ${SCALAPACK_LIBRARY}")
+  endif(SCALAPACK_LIBRARY)
+else()
+  message(STATUS "STRUMPACK was not found.")
+endif()
+
+#set(STRUMPACK_INCLUDE_DIR CACHE PATH "Path to StrumpackConfig.hpp")
+#set(STRUMPACK_LIBRARIES CACHE PATH "Path to strumpack library")
```
nychiang, line 41: Should we remove these two lines?

ashermancinelli (approved):
IMO this large of a patch is pretty impossible to thoroughly review, but I think we've taken enough care with each PR to ensure this is ready, and the functional CI is passing. LGTM. I can tag once this is merged.

cnpetra:
> review

you're right -- the goal with this PR was more to raise awareness.

nychiang (approved):

cmake/FindHiopSTRUMPACK.cmake
cnpetra: we'll do these in a follow-up so we don't clutter this PR too much

Issue #207, opened 2021-03-31
release v0.4 missing items


cnpetra:
need to release 0.4.1 because 0.4 
 
- is missing the latest pdf with user manual 
- had an outdated list of contributors (master was updated)

nychiang:
remove this [line](https://github.com/LLNL/hiop/blob/627f4d890fb35bd6620fe50ea2f4b23b45d59715/doc/src_usermanual/nlpsparse.tex#L107)

nychiang:
problm --> problem in [line](https://github.com/LLNL/hiop/blob/627f4d890fb35bd6620fe50ea2f4b23b45d59715/doc/src_usermanual/techrep_main.tex#L676)

PR #208, opened 2021-04-07
patch release 0.4 as 0.4.1

cnpetra:
This PR addresses a couple of small items forgotten when releasing 0.4. Addresses all the issues reported by issue #207 and comments made in PR#206.

Plan to do a new release as 0.4.1 (as opposed to updating 0.4, which I am not sure it's possible).


ashermancinelli:
We can update the tag if you want, but it seems better to just have a new patch release like you suggest.￼

ashermancinelli (approved):

doc/src_usermanual/nlpsparse.tex
```diff
@@ -104,7 +104,6 @@ \subsubsection{Specifying the optimization problem}
 \begin{lstlisting}
 bool eval_Jac_cons(const long long& n, const long long& m, 
 			          const double* x, bool new_x,
-			          const long long& nsparse, const long long& ndense,
```
pelesh, line 107: Why was this line removed? Was this a copy-paste error?

pelesh:
In `techrep_main.tex`, line 266, C++ requirement should be C++11.

pelesh:
On page 22, memory space option, the "default" option is sequential code. It is the no-RAJA option. The other three match Umpire memory spaces HOST, DEVICE and UM.

pelesh (changes requested):
I would suggest fixing C++ standard and memory space options description in the manual. Other than that, it is ready to merge.

doc/src_usermanual/nlpsparse.tex
cnpetra: was copy-paste error indeed

nychiang (approved):

PR #216, opened 2021-04-22
development of feasibility restoration

nychiang:
Restore feasibility when line search fails, i.e., step size is too small.

Note:
1. Pass valgrind tests.
2. Pass 3 ExaGO OPFLOW tests 

@cnpetra

### DIFF
--- a/src/LinAlg/hiopVectorRajaPar.cpp
+++ b/src/LinAlg/hiopVectorRajaPar.cpp
@@ -2,47 +2,47 @@
 // Produced at the Lawrence Livermore National Laboratory (LLNL).
 // LLNL-CODE-742473. All rights reserved.
 //
-// This file is part of HiOp. For details, see https://github.com/LLNL/hiop. HiOp 
-// is released under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause). 
+// This file is part of HiOp. For details, see https://github.com/LLNL/hiop. HiOp
+// is released under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause).
 // Please also read “Additional BSD Notice” below.
 //
-// Redistribution and use in source and binary forms, with or without modification, 
+// Redistribution and use in source and binary forms, with or without modification,
 // are permitted provided that the following conditions are met:
-// i. Redistributions of source code must retain the above copyright notice, this list 
+// i. Redistributions of source code must retain the above copyright notice, this list
 // of conditions and the disclaimer below.
-// ii. Redistributions in binary form must reproduce the above copyright notice, 
-// this list of conditions and the disclaimer (as noted below) in the documentation and/or 
+// ii. Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the disclaimer (as noted below) in the documentation and/or
 // other materials provided with the distribution.
-// iii. Neither the name of the LLNS/LLNL nor the names of its contributors may be used to 
-// endorse or promote products derived from this software without specific prior written 
+// iii. Neither the name of the LLNS/LLNL nor the names of its contributors may be used to
+// endorse or promote products derived from this software without specific prior written
 // permission.
 //
-// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY 
-// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES 
-// OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT 
-// SHALL LAWRENCE LIVERMORE NATIONAL SECURITY, LLC, THE U.S. DEPARTMENT OF ENERGY OR 
-// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR 
-// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS 
-// OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED 
-// AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT 
-// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+// OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+// SHALL LAWRENCE LIVERMORE NATIONAL SECURITY, LLC, THE U.S. DEPARTMENT OF ENERGY OR
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+// OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+// AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
 // EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 //
 // Additional BSD Notice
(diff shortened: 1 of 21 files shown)


nychiang:
Just rebased this branch to the latest develop branch.

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -341,6 +341,30 @@ void hiopVectorRajaPar::copyFromStarting(int start_index, const hiopVector& src)
...
+ */
+void hiopVectorRajaPar::copy_from_starting_at(const double* v, int start_index_in_v, int nv)
+{
+  // If nothing to copy, return.
+  if(nv == 0)
+    return;
+
+  auto& rm = umpire::ResourceManager::getInstance();
+  double* vv = const_cast<double*>(v); // <- cast away const
+  rm.copy(data_dev_, vv + start_index_in_v, nv*sizeof(double));
```
cnpetra, line 365: it's unusual that for a `copy` method to require the source to be non-const ...

src/Optimization/hiopFRProb.cpp
```diff
@@ -0,0 +1,852 @@
...
+  }
+
+  mu_ = mu;
+  zeta_ = std::sqrt(mu_);
+
+  return true;
+}
+
+bool hiopFRProbSparse::force_update(double obj_value,
+                                         const int n,
```
cnpetra, line 775: alignment

src/Utils/hiopOptions.cpp
```diff
@@ -240,6 +240,16 @@ void hiopOptions::registerOptions()
 		      "Factor to decrease the constraint violation in second order correction.");
   }
 
+  // feasibility restoration
+  {
+    registerNumOption("kappa_resto", 0.9, 0, 1.0,
+		      "Factor to decrease the constraint violation in feasibility restoration. (default 0.9)");
```
cnpetra, line 246: alignment

src/LinAlg/hiopVectorRajaPar.cpp
pelesh: It is casting to/from void pointer somewhere I believe, that is why it fails for const pointers.

nychiang: I copy this line from [here](https://github.com/LLNL/hiop/blob/a7aa4e0dc030b3030c82c872308d2e753129cffe/src/LinAlg/hiopVectorRajaPar.cpp#L313). Other lines in the file also shows that non-const source is passed to umpire.
@pelesh @ashermancinelli  This is required by umpire, isn't it?

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -413,7 +437,7 @@ void hiopVectorRajaPar::copyToStarting(int start_index, hiopVector& dst)
  * @pre start_index + vec.n_local_ <= n_local_
  * @pre `this` and `vec` are not distributed
  */
-void hiopVectorRajaPar::copyToStarting(hiopVector& vec, int start_index/*_in_dest*/)
+void hiopVectorRajaPar::copyToStarting(hiopVector& vec, int start_index/*_in_dest*/) const
 {
```
pelesh, line 441:
```suggestion
void hiopVectorRajaPar::copyToStarting(hiopVector& vec, int start_index_in_dest) const
{
```
Why not just expand the variable name instead of using a comment?

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -426,7 +450,7 @@ void hiopVectorRajaPar::copyToStarting(hiopVector& vec, int start_index/*_in_des
   rm.copy(v.data_dev_ + start_index, this->data_dev_, this->n_local_*sizeof(double));
 }
 
-void hiopVectorRajaPar::copyToStartingAt_w_pattern(hiopVector& vec, int start_index/*_in_dest*/, const hiopVector& select)
+void hiopVectorRajaPar::copyToStartingAt_w_pattern(hiopVector& vec, int start_index/*_in_dest*/, const hiopVector& select) const
 {
```
pelesh, line 454:
```suggestion
void hiopVectorRajaPar::copyToStartingAt_w_pattern(hiopVector& vec, int start_index_in_dest, const hiopVector& select) const
{
```
Ditto.

src/LinAlg/hiopVectorRajaPar.cpp
cnpetra: `const void*` would not require the `const_cast`, but apparently is just a `void* ` ...

cnpetra: https://github.com/LLNL/Umpire/blob/8d7fba7eaaefe194a90d1b26b8b7fb9138d485c3/src/umpire/ResourceManager.cpp#L383

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -1165,6 +1205,24 @@ double hiopVectorRajaPar::logBarrier_local(const hiopVector& select) const
...
+ * @brief Sum all elements
+ */
+double hiopVectorRajaPar::sum_local() const
+{
+  double* data = data_dev_;
+  RAJA::ReduceSum< hiop_raja_reduce, double > sum(0.0);
+  RAJA::forall< hiop_raja_exec >(
+    RAJA::RangeSegment(0, n_local_),
+		RAJA_LAMBDA(RAJA::Index_type i)
+    {
```
pelesh, line 1218:
```suggestion
    RAJA::RangeSegment(0, n_local_),
    RAJA_LAMBDA(RAJA::Index_type i)
    {
```
Use spaces instead of tabs for indentation.

src/Optimization/hiopFRProb.cpp
```diff
@@ -0,0 +1,852 @@
...
+        last_row_idx = row_idx;
+        k = k-2;
+      } else {
+        // x
+        iJacS[k] = row_idx;
+        jJacS[k] = jcol_d[k_base];
+        k--;
+        k_base--;
+      }
+    }
```
pelesh, line 364: This does not seem to be portable nor a thread-safe implementation. Is it expected this function will run sequentially and on CPU only?

src/Optimization/hiopAlgFilterIPM.cpp
```diff
@@ -2012,4 +2050,98 @@ int hiopAlgFilterIPMBase::apply_second_order_correction(hiopKKTLinSys* kkt,
...
+                               *_c,
+                               *_d,
+                               *it_curr->get_yc(),
+                               *it_curr->get_yd(), //lambda,
+                               _mu,
+                               _alpha_dual,
+                               _alpha_primal)) {
+      solver_status_ = Error_In_FR;
+      return false;
+    }
```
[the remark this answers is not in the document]
nychiang: These member variables are used for ages and probably we need to replace them carefully. For example, I see `c_` and `d_` are used[ in the head file here](https://github.com/LLNL/hiop/blob/develop/src/Optimization/hiopAlgFilterIPM.hpp#L92)
We probably can create a new issue about this, and fix all the member variables in that ticket.

src/Optimization/hiopFRProb.cpp
pelesh: There are several instances like this in `hiopFRProbSparse` class where member method access raw data. It would certainly be worthwhile to encapsulate these computations in specialized matrix kernels and keep `hiopFRProbSparse` implementation generic.

src/Optimization/hiopAlgFilterIPM.cpp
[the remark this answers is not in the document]
nychiang: I mean member variables in other files as well.

src/Optimization/hiopFRProb.cpp
nychiang: Good point, and I guess I know how to do it.
It seems that we cannot test the corresponding RAJA implementation due to the lack of `gpu` mode for sparse instance. Therefore probably I will create a new ticket to address this issue.
Will discuss with @cnpetra tomorrow.

src/Optimization/hiopNlpFormulation.cpp
```diff
@@ -807,6 +814,24 @@ void hiopNlpFormulation::copy_EqIneq_to_cons(const hiopVector& yc_in,
...
+  assert(cons.get_size() == n_cons);
+  assert(yc_in.get_size() + yd_in.get_size() == n_cons);
+  //split whole lambda array to eq/ineq
+  for(int i=0; i<n_cons_eq; ++i) {
+    yc_arr[i] = cons_arr[cons_eq_mapping_[i]];
+  }
+  for(int i=0; i<n_cons_ineq; ++i) {
+    yd_arr[i] = cons_arr[cons_ineq_mapping_[i]];
+  }
+}
```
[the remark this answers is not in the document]
nychiang: yes, we realized this issue. It happens whenever `cons_eq_mapping_` and `cons_ineq_mapping_ ` are used. For example, [here](https://github.com/LLNL/hiop/blob/e5204ae1a450b351788812f400e4caa7259db4b7/src/Optimization/hiopNlpFormulation.cpp#L781)

src/LinAlg/hiopVectorRajaPar.cpp
pelesh: I remember we needed to cast away `const` to use Umpire copy.  Then again, that was three major versions ago. I'll check again.

CC @davidbeckingsale

src/Optimization/hiopFRProb.cpp
cnpetra: as per our discussion, to prioritize the merging of this, on which other things depend, @nychiang please fill an issue with this and we'll follow up with a matrix method and associated device implementation for this.

cnpetra:
> I believe it would be worthwhile to implement class `hiopFRProbSparse` in a generic way and encapsulate all access to raw data within custom linear algebra kernels. Same should be done in similar instances elsewhere.
> 
> I would also suggest implementing new kernels in a thread-safe and portable way. I am afraid it will only be more work if we decide to leave this for later.

I opened issue https://github.com/LLNL/hiop/issues/232 and documented a few places that need to be rebased on vector and matrix kernels. We need to merge this PR asap as it blocks other activities, like `hiopFRProbMDS` and running on Summit.

Issue #215, opened 2021-04-20
Refactor linear algebra library CMake


ashermancinelli:
Library creation logic is repeated throughout the subdirectories of `src/`. Library creation through a single macro is the best-practice established in SUNDIALS and ExaGO.

ashermancinelli:
@CameronRutherford and I will look at this as per our discussion today.

@cnpetra @nychiang @pelesh

PR #217, opened 2021-04-22
Use macros for CMake library creation

ashermancinelli:
@cnpetra I kept the old build system intact with a suffix of `_legacy`. If something goes wrong, we can direct users to rename `libhiop_legacy.so -> libhiop.so`, etc. Exposes cmake target `HiOp::HiOp` and each of `HiOp::SHARED` and `HiOp::STATIC` if they are enabled.

Closes: #215


ashermancinelli:
The patch ended up being smaller than expected, so I did all the components in one go.

src/Drivers/CMakeLists.txt
```diff
@@ -39,5 +39,5 @@ endif()
 
 if(HIOP_BUILD_SHARED AND NOT HIOP_USE_GPU)
   add_executable(nlpMDS_cex4.exe nlpMDS_ex4.c)
-  target_link_libraries(nlpMDS_cex4.exe hiop_shared)
+  target_link_libraries(nlpMDS_cex4.exe HiOp::SHARED)
```
pelesh, line 42:
```suggestion
  target_link_libraries(nlpMDS_cex4.exe HiOp::HiOp)
```
I am not sure if having `SHARED` target is a good idea. I would think that `HiOp::HiOp` target should install shared, static or both libraries depending on configuration option.

src/LinAlg/CMakeLists.txt
```diff
@@ -46,17 +46,52 @@ if(HIOP_WITH_KRON_REDUCTION)
...
-# Link linear algebra target to Umpire & RAJA when enabled
 if(HIOP_USE_RAJA)
   target_link_libraries(hiopLinAlg PUBLIC umpire RAJA OpenMP::OpenMP_CXX)
 endif()
+# ---
+
+set(hiopLinAlg_HEADERS
+  hiopLinAlgFactory.hpp
+  hiopLinSolver.hpp
+  hiopLinSolverIndefDenseLapack.hpp
```
pelesh, line 60:
```suggestion
set(hiopLinAlg_HEADERS
  hiopLinSolver.hpp
  hiopLinSolverIndefDenseLapack.hpp
```
I believe we should not have linear algebra factory headers in HiOp API. These should be used internally within HiOp, but not exposed to applications linking to HiOp.

src/Drivers/CMakeLists.txt
ashermancinelli: IIUC the C interface needs the shared library which is why this test is only built `if(HIOP_BUILD_SHARED)` - @cnpetra is this correct?. `HiOp::HiOp` will point to shared by default **_and to static only if shared is not being built_** so that the alias target always points to a valid library. This is why I don't just link against `HiOp::HiOp`.

src/LinAlg/CMakeLists.txt
ashermancinelli: Great catch!

ashermancinelli:
As per offline discussion, I'll be updating this with the legacy build system removed.

pelesh (changes requested):
Applications segfault when HiOp is built as shared lib.

ashermancinelli:
The segfaulting was an issue with CI scripts on marianas and this has been resolved.

CameronRutherford (approved):

pelesh:
This branch needs a little more work. will reopen once is ready for a PR.

Issue #205, opened 2021-03-26
unknown error in Ex3 with MPI


nychiang:
When running EX3 with 5 variables on multiple MPI processes, some MPI function returns wrong values.
For example, in the latest `develop` branch ( 3/26/2021, commit [9eb6362](https://github.com/LLNL/hiop/commit/9eb6362a7d59cdd137cef396b52a83a9f0e50611) , negative values are returned when computing infinity norm ([here](https://github.com/LLNL/hiop/blob/9eb6362a7d59cdd137cef396b52a83a9f0e50611/src/Optimization/hiopResidual.cpp#L148))

In branch `ls-prob-fix`, HiOp is frozen at the 1st iteration.

NOTE that this problem only happens in EX3 with 5 variables. One cannot overserve this issue if there are more than 6 variables.

nychiang:
Update:
this bug comes from removing fixed variables. In this 5 variables example, one MPI process has 0 variable after removing fixed variable. As a result, the distributed vector is not created correctly [here](https://github.com/LLNL/hiop/blob/627f4d890fb35bd6620fe50ea2f4b23b45d59715/src/Optimization/hiopNlpFormulation.cpp#L315), and MPI_ALL_REDUCE is not matched later.

@cnpetra seems this is not a quick fix. For all the MPI functions in the vector kernel, we need to test if its size is zero or not. If it is zero, add dummy MPI_Allreduce function to return an appropriate value.
I remember @pelesh mentioned a similar bug, i.e., vector with size zero, in another issue.

cnpetra:
that makes a lot of sense

nychiang:
@cnpetra 
I think PR #205 hasn't fixed this problem yet, and it still has mpi mismatches in the code.
When we solver Ex3 with 1 mpi or 2 mpi processes, we got different results. That is, the inf_pr computed with 2 mpi processes is wrong.
It seems this bug exists for a while, as it can be reproduced in v0.3 ([7e8adae](https://github.com/LLNL/hiop/commit/7e8adae).

cnpetra:
will take a look

cnpetra:
I cannot reproduce nor see anything wrong with ex3 on one and two ranks. Can you please email me the parameters/size of ex3 and the log file?

cnpetra:
thanks, I reproduced and fixed the issue.

PR #221, opened 2021-04-26
issue205 fix

cnpetra:
fixes a small issue in matrix class when having 0 local vars in the mpi rank

fixes Issue #205

### DIFF
--- a/src/LinAlg/hiopMatrixDenseRowMajor.cpp
+++ b/src/LinAlg/hiopMatrixDenseRowMajor.cpp
@@ -371,15 +371,13 @@ void hiopMatrixDenseRowMajor::print(FILE* f,
   }
 }
 
-#include <unistd.h>
-
 /*  y = beta * y + alpha * this * x  
  *
  * Sizes: y is m_local_, x is n_local_, the matrix is m_local_ x n_global_, and the
  * local chunk is m_local_ x n_local_ 
 */
-void hiopMatrixDenseRowMajor::timesVec(double beta, hiopVector& y_,
-			       double alpha, const hiopVector& x_) const
+void hiopMatrixDenseRowMajor::
+timesVec(double beta, hiopVector& y_, double alpha, const hiopVector& x_) const
 {
   hiopVectorPar& y = dynamic_cast<hiopVectorPar&>(y_);
   const hiopVectorPar& x = dynamic_cast<const hiopVectorPar&>(x_);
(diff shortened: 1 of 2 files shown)


src/LinAlg/hiopMatrixDenseRowMajor.cpp
```diff
@@ -371,15 +371,13 @@ void hiopMatrixDenseRowMajor::print(FILE* f,
...
-
 /*  y = beta * y + alpha * this * x  
  *
  * Sizes: y is m_local_, x is n_local_, the matrix is m_local_ x n_global_, and the
  * local chunk is m_local_ x n_local_ 
 */
-void hiopMatrixDenseRowMajor::timesVec(double beta, hiopVector& y_,
-			       double alpha, const hiopVector& x_) const
+void hiopMatrixDenseRowMajor::
+timesVec(double beta, hiopVector& y_, double alpha, const hiopVector& x_) const
```
nychiang, line 380: Each argument on a separate line, i.e., using style guide.
Same comment applied to the other functions in this file.

src/Optimization/hiopResidual.cpp
```diff
@@ -58,11 +58,13 @@ hiopResidual::hiopResidual(hiopNlpFormulation* nlp_)
...
   rd = nlp->alloc_dual_ineq_vec();
-  rxl = rx->alloc_clone();
-  rxu = rx->alloc_clone();
-  rdl = rd->alloc_clone();
-  rdu = rd->alloc_clone();
+  rd->setToZero();
+  rxl = rx->new_copy();
+  rxu = rx->new_copy();
+  rdl = rd->new_copy();
+  rdu = rd->new_copy();
```
nychiang, line 67: I also fixed this in my fr branch. I think it is better to PR it here and then I will rebase my fr to the develop branch.

nychiang (approved):

src/LinAlg/hiopMatrixDenseRowMajor.cpp
cnpetra: this is actually valid, arguments can stay on the same line as long as the line is not too long... styleguide needs maybe to be updated to make this more explicit

src/Optimization/hiopResidual.cpp
cnpetra: ok, will merge

PR #223, opened 2021-04-29
Fix scaling

nychiang:
fix gradient scaling if the input Jacobian has empty rows.


cnpetra:
can you fix the spaces? looks good otherwise. I was about to ask you what to do about empty rows, I guess this will fix it.

PR #225, opened 2021-04-30
Fix row_max_abs_value() in RAJA implementation

nychiang:
Fix sparse/dense RAJA implementation and the implementation of gradient scaling.
Now it supports empty rows in Jacobian and passes the tests with 3 exago problems.

Note that I also changes the file ascentVariable.sh, to use the same setting as exago on ascent.


src/Drivers/CMakeLists.txt
```diff
@@ -13,12 +13,18 @@ target_link_libraries(nlpDenseCons_ex3.exe hiop)
...
+target_link_libraries(nlpMDS_ex6.exe hiop)
+
 if(HIOP_USE_RAJA)
   if(HIOP_USE_GPU)
     set_source_files_properties(nlpMDSForm_raja_ex4.cpp nlpMDS_ex4_raja_driver.cpp PROPERTIES LANGUAGE CUDA)
+    set_source_files_properties(nlpMDSForm_raja_ex6.cpp nlpMDS_ex6_raja_driver.cpp PROPERTIES LANGUAGE CUDA)
   endif()
   add_executable(nlpMDS_ex4_raja.exe nlpMDS_ex4_raja_driver.cpp nlpMDSForm_raja_ex4.cpp)
   target_link_libraries(nlpMDS_ex4_raja.exe hiop)
+  add_executable(nlpMDS_ex6_raja.exe nlpMDS_ex6_raja_driver.cpp nlpMDSForm_raja_ex6.cpp)
```
cnpetra, line 26: I think we should add MDS Ex6 to the suite of tests

src/Drivers/nlpMDSForm_ex6.hpp
```diff
@@ -0,0 +1,651 @@
+#ifndef HIOP_EXAMPLE_EX6
```
cnpetra, line 1: `HIOP_EXAMPLE_EX6_MDS` to void conflicts with the macro from the sister sparse file

cnpetra (approved):
beside the comments I left, we need to 

- rename all the files starting with "nlpMDSForm" start with "nlpMDS" (this inconsistency applies both to ex4 and ex6)
- what was the exact cause of the crash in the two  `row_max_abs_value` methods? RAJA ranges won't accept zero lengths? Or something related to the pointer arithmetic in HiOp

nychiang:
I want to discuss the name of this example. I can see 3 options:
1. Follow the tradition and name it as Ex9? (I think Ex8 is used by Frank). I don't think that is a good idea, since it may ruin others' work.
2. Using MDS_Ex6, to follow the names nlpMDS_ex4 and nlpMDS_ex5. 
3. How about MDS_Ex1?

nychiang:
The exact cause of the crash in the two row_max_abs_value methods:
1. Sparse: It is an empty matrix with dimension 0x0. Therefore `idx_start` is not generated [in this function](https://github.com/LLNL/hiop/blob/3fcb788d223eec24c0241680070c4a9a5ec71ef3/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp#L500). Then later RAJA access `idx_start[0]` in the loop
2. Dense: A vector of length 0 is created [(here)](https://github.com/LLNL/hiop/blob/3fcb788d223eec24c0241680070c4a9a5ec71ef3/src/LinAlg/hiopMatrixMDS.hpp#L176) .

pelesh:
What is (supposed to be) the difference between MDS_Ex6 and MDS_Ex4 examples? I get nothing when I diff 4 vs. 6 files.

src/Drivers/nlpMDSForm_ex6.hpp
pelesh: Alternatively:
```c++
#pragma once
```

pelesh:
> beside the comments I left, we need to
> 
> * what was the exact cause of the crash in the two  `row_max_abs_value` methods? RAJA ranges won't accept zero lengths? Or something related to the pointer arithmetic in HiOp

I believe it is Umpire that will not let you access zero length array. This is imho appropriate behavior. When using raw pointers in CPU implementation, there is no way to detect the array length, but if the pointer arithmetic is correct, no damage will be done.

nychiang:
> What is (supposed to be) the difference between MDS_Ex6 and MDS_Ex4 examples? I get nothing when I diff 4 vs. 6 files.

@pelesh 
I make an empty row in the sparse Jacobian in Ex6, see [Ex6](https://github.com/LLNL/hiop/blob/27214b0a69d4337959b0ff2a9d6a0c6942ed7e84/src/Drivers/nlpMDSForm_ex6.hpp#L27) v.s. [Ex4](https://github.com/LLNL/hiop/blob/27214b0a69d4337959b0ff2a9d6a0c6942ed7e84/src/Drivers/nlpMDSForm_ex4.hpp#L27).
You probably use command `diff nlpMDSForm_ex6.hpp nlpMDSForm_ex6.hpp` and it returns nothing. ( This is exactly what I did after I saw your comment. I was shocked and was wondering where are my changes. Then I opened both files in vim and confirmed that all my changes are there. :-) )

src/Drivers/CMakeLists.txt
pelesh: Perhaps we should think about subdirectory structure in `src/Drivers`? It is getting crowded there.

pelesh:
I figured out how to diff correctly :). Sorry for the noise. Still, the differences between Ex4 and Ex6 are minimal. Would it be possible to add a parameter to Ex4 so one can to turn a row in the sparse Jacobian to a zero row through a command line option? That would be a more compact solution

src/Drivers/CMakeLists.txt
ashermancinelli: We could move current `tests/` directory to `tests/unit/` and move `src/Drivers` to `tests/functionality`?

src/Drivers/CMakeLists.txt
```diff
@@ -13,12 +13,18 @@ target_link_libraries(nlpDenseCons_ex3.exe hiop)
 add_executable(nlpMDS_ex4.exe nlpMDS_ex4_driver.cpp)
 target_link_libraries(nlpMDS_ex4.exe hiop)
 
+add_executable(nlpMDS_ex6.exe nlpMDS_ex6_driver.cpp)
+target_link_libraries(nlpMDS_ex6.exe hiop)
+
 if(HIOP_USE_RAJA)
   if(HIOP_USE_GPU)
     set_source_files_properties(nlpMDSForm_raja_ex4.cpp nlpMDS_ex4_raja_driver.cpp PROPERTIES LANGUAGE CUDA)
+    set_source_files_properties(nlpMDSForm_raja_ex6.cpp nlpMDS_ex6_raja_driver.cpp PROPERTIES LANGUAGE CUDA)
```
ashermancinelli, line 22: [nitpick] I think this would look a little better
```suggestion
    set_source_files_properties(
      nlpMDSForm_raja_ex4.cpp
      nlpMDS_ex4_raja_driver.cpp
      nlpMDSForm_raja_ex6.cpp
      nlpMDS_ex6_raja_driver.cpp
      PROPERTIES LANGUAGE CUDA
      )
```

src/Drivers/CMakeLists.txt
ashermancinelli: Actually, my suggestion is out of scope for this PR.

ashermancinelli (approved):
Other than my nitpick and the test organization, lgtm.

cnpetra:
> I want to discuss the name of this example. I can see 3 options:
> 
> 1. Follow the tradition and name it as Ex9? (I think Ex8 is used by Frank). I don't think that is a good idea, since it may ruin others' work.
> 2. Using MDS_Ex6, to follow the names nlpMDS_ex4 and nlpMDS_ex5.
> 3. How about MDS_Ex1?

Ex9 is taken by PriDec solver's driver. You can use nlpMDS_ex10 if you want. Or adopt Slaven's suggestion to modify ex4. nlpMDS_ex6 is also good as long asthet underlying MDS NL resembles the sparse NLP from sparse ex4, even though it seems it is more line ex4. Up to you.

src/Drivers/CMakeLists.txt
cnpetra: @ashermancinelli That structure of `tests/` looks good, but I prefer `Drivers/` to stay in the current location and have the current content as they are user oriented, while `tests/` not so much. In `tests/functionality` we can add tests additional tests based on `src/Drivers/`.

src/Drivers/nlpMDS_ex4.hpp
```diff
@@ -517,8 +537,12 @@ class Ex4OneCallCons : public Ex4
 	  for(int i=0; i<nd; i++) cons[con_idx] += y[i];
 
 	} else if(con_idx==ns+1) {
-	  cons[con_idx] = x[1];
-	  for(int i=0; i<nd; i++) cons[con_idx] += y[i];
+	  if(empty_sp_row_) {
```
cnpetra, line 540: alignment issue ?

nychiang: I just fixed this paragraph.

nychiang:
@pelesh @ashermancinelli @ashermancinelli 
MDS_ex6 is removed. I add an additional user option to MDS_ex4, to control if the optimization problem has empty row in sparse constraint Jacobian. Please have a look of the usage.
If everyone agree with the change, I will record the solutions with empty sparse Jacoian and add them into self check.

cnpetra:
lgtm

src/Drivers/nlpMDS_ex4_driver.cpp
```diff
@@ -52,7 +63,7 @@ static bool parse_arguments(int argc, char **argv,
     return false; //5 or more arguments
   }
 
-  if(self_check && n_sp!=400 && n_de!=100)
+  if(self_check && (n_sp!=400 || n_de!=100 || empty_sp_row) )
```
pelesh, line 66: I believe this logic will force to set empty sparse row. I think we would want this example to test cases with and without an empty row.

src/LinAlg/hiopMatrixRajaDense.cpp
```diff
@@ -1385,6 +1385,9 @@ void hiopMatrixRajaDense::row_max_abs_value(hiopVector &ret_vec)
 {  
   assert(ret_vec.get_size() == m());
   ret_vec.setToZero();
+  if(0 == m()) {
+    return;
+  }
```
pelesh, line 1390:
```suggestion
  if(0 == m_local_) {
    return;
  }  
```
I know this won't change anything because HiOp matrices are not partitioned by rows, but I would make this check explicitly with respect to _local_ row numbers.

src/Drivers/nlpMDS_ex4_driver.cpp
pelesh: Why not add third test for example 4? For example:
```cmake
  add_test(NAME NlpMixedDenseSparse4_3 COMMAND ${RUNCMD} bash -c "$<TARGET_FILE:nlpMDS_ex4.exe> 400 100 0 -empty_sp_row -selfcheck)
```

src/Drivers/nlpMDS_ex4_raja_driver.cpp
```diff
@@ -55,7 +64,7 @@ static bool parse_arguments(int argc, char **argv,
     return false; //5 or more arguments
   }
 
-  if(self_check && n_sp!=400 && n_de!=100)
+  if(self_check && (n_sp!=400 || n_de!=100 || empty_sp_row) )
```
pelesh, line 67: Same here -- we could allow an empty row for tests and add third test fro example 4.

pelesh:
> If everyone agree with the changes, I will record the solutions with empty sparse Jacoian and add them into self check.

Yes, that would be great! You can then add a test for that case, as well. For this PR, I don't think you need to add the "deep check" variant of the test.

nychiang:
All the comments have been addressed.

ashermancinelli:
Fails at ORNL:
```
/gpfs/wolf/proj-shared/csc359/ci/144700/src/Drivers/nlpMDS_raja_ex4.cpp:1:10: fatal error: nlpMDSForm_raja_ex4.hpp: No such file or directory
 #include "nlpMDSForm_raja_ex4.hpp"
          ^~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

nychiang:
> Fails at ORNL:
> 
> ```
> /some/path/src/Drivers/nlpMDS_raja_ex4.cpp:1:10: fatal error: nlpMDSForm_raja_ex4.hpp: No such file or directory
>  #include "nlpMDSForm_raja_ex4.hpp"
>           ^~~~~~~~~~~~~~~~~~~~~~~~~
> compilation terminated.
> ```
That bug has been fixed in commit `4df3951`, after I noticed if from PNNL pipeline.
There is a delay in ORNL pipeline and hence you didn't see the latest result. Now you can see it pass all the pipeline tests on PNNL/ORNL. :-)

pelesh (approved):
All pipelines pass, looks great ...

ashermancinelli:
@nychiang Great thanks for following up! lgtm

PR #226, opened 2021-05-04
CMake refactor

ashermancinelli:
* Move library creation logic under `src` directory
* Collect interface headers in `src/*` subdirectories
* Use `hiop_tpl` over `hiop_math` for clarity and to conform to best-practices
* Allow building of shared and static libraries independently
* Push logic to later in the CMakeLists.txt and keep source/header lists at the top to make it easier on future contributors
* Expose `HiOp::HiOp` target
* Update build variables on PNNL cluster

@cnpetra this refactor was deemed simpler and more understandable than the previous one in an offline conversation with @pelesh and I.


cnpetra:
it appears the C examples are not building properly, see github checks

```CMake Error at src/Drivers/CMakeLists.txt:47 (add_executable):
    Target "nlpMDS_cex4.exe" links to target "HiOp::SHARED" but the target was
    not found.  Perhaps a find_package() call is missing for an IMPORTED
    target, or an ALIAS target is missing?
```

scripts/marianasVariables.sh
```diff
@@ -24,5 +26,11 @@ module load gcc/7.3.0
...
-
-spack env activate exago-v0-99-2-hiop-v0-3-99-2-marianas --with-view
+module load magma-2.5.4-gcc-7.3.0-vgkvbvm
+module load openblas-0.3.12-gcc-7.3.0-dzz6rfy
+module load raja/0.13.0-gcc-7.3.0-pwmbk4o
+module load umpire-4.1.2-gcc-7.3.0-qqotfxd
+module load suite-sparse/5.8.1-gcc-7.3.0-uivxrx7
+module load coinhsl/2015.06.23-gcc-7.3.0-kvdofab
+module load metis-5.1.0-gcc-7.3.0-ymmhgpk
+module load camp-0.1.0-gcc-7.3.0-lfmsuz3
```
pelesh, line 36: I would suggest using spack environment here rather than loading spack modulae one by one.

src/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+add_subdirectory(Interface)
+add_subdirectory(Optimization)
+add_subdirectory(LinAlg)
+add_subdirectory(Drivers)
```
pelesh, line 4: I am not sure if this matters, but I would add `Drivers` directory at the very end of this file after HiOp libraries are built. If nothing else, it makes the logic of the CMake code easier to follow.

src/Drivers/CMakeLists.txt
```diff
@@ -39,5 +45,5 @@ endif()
 
 if(HIOP_BUILD_SHARED AND NOT HIOP_USE_GPU)
```
pelesh, line 46:
```suggestion
if(HIOP_BUILD_SHARED)
```
Do we need to disable shared library build when enabling GPU support? I believe this is remnant of an older version of build system.

@cnpetra, this may cause the test failure you observed.

scripts/marianasVariables.sh
ashermancinelli: CMake links against the environment's hiop library as opposed to the one currently being built - otherwise I would much have rathered the original 😄

src/CMakeLists.txt
ashermancinelli: Great point!

src/Drivers/CMakeLists.txt
ashermancinelli: I think you're entirely correct - @cnpetra can you confirm?

scripts/marianasVariables.sh
pelesh: I believe this is fixable by using correct rpath settings. I'll tinker with this a little more.

src/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+add_subdirectory(Interface)
+add_subdirectory(Optimization)
+add_subdirectory(LinAlg)
+add_subdirectory(Utils)
+add_subdirectory(Drivers)
+
```
pelesh, line 6: I would move this line to the very bottom of the CMake file. Everything that is in `Drivers` directory links to the HiOp library, so I would make sure the library is built and the aliases are set correctly before building the examples in the `Drivers` dir. I think CMake has its way in managing order in which different targets are built, but I would just make that order explicit in this code.

scripts/marianasVariables.sh
ashermancinelli: Rpath isn't so much the issue iiuc - the alternative would be `spack env activate <name> --without-view` and then `spack load -r <dependency>` for every dependency which isn't any more readable than loading modules imo

src/CMakeLists.txt
ashermancinelli: Again, good point

scripts/marianasVariables.sh
ashermancinelli: I also don't know how rpath and ld library path are prioritized, so you may be right and the external libhiop might never be found if we better manage our rpaths.

ashermancinelli:
Passing CI everywhere

nychiang (approved):

src/Drivers/CMakeLists.txt
cnpetra: In short, I am not sure... I do not recall whether the shared library does not work on GPUs or there are other limitation(s) preventing enabling shared library for GPUs. @michel2323 maybe you can shed some light? 

In any case, it looks like we need a C RAJA using shared library example in the near future to ensure that the latest bells and whistles in HiOp work with C interface and shared library (which are the preamble for interfacing with Julia)

michel2323: Nothing that I remember. I only remember the CMake issues. A while ago I tried HiOp using CUDA internally and it's all good so far, even through Julia.

ashermancinelli:
This is building share and static with GPU enabled in CI, so I think this is good to go

scripts/marianasVariables.sh
ashermancinelli: I have some new spack environments coming which should make this a lot easier. I think this is good for this PR, but I appreciate the suggestion!

pelesh (approved):

pelesh:
CC @rothpc @nkoukpaizan

cnpetra (approved):

Issue #220, opened 2021-04-23
TPL CMake logic in top-level cmake lists


ashermancinelli:
Logic for third party libraries should not be in the top-level cmakelists.txt, but hidden in many other cmake files to break up the logic. [See SUNDIALS top-level cmake linked here](https://github.com/LLNL/sundials/blob/master/CMakeLists.txt#L167). Logic for each third party library ought to be modular and in seperate cmake files.

cnpetra:
looks like this is also address, please confirm?

PR #234, opened 2021-05-14
Export Targets in CMake

### LINKED ISSUE #219: No Exported CMake Targets
Best practice for a cmake project is to export cmake targets in the install directory so other project may more easily integrate the former project. HiOp does not have this capability.

CameronRutherford:
Closes #219, Related to #220 

cc @ashermancinelli @pelesh 

@balos1 I have based this work this after what you have done in Sundials CMake, and I would appreciate your review in making sure this is the an adequete adaptation for HiOp's CMake.

Further discussion may be warranted, but this should be a good start!

### libhiop.so not linking against libmagma.so
@pelesh pointed out that the output `libhiop.so` should link against all libraries in it's r path. However:
```bash
[rcruther <build_dir>/src]$ ldd libhiop.so  | grep magma
    libmagma.so => not found
```

This could be something that I am doing wrong in CMake when linking against Magma, or in our find module. This could also be taken care of in #220, or wrapped in a separate issue.

Note that the magma library is being exported in:
- `<install_dir>/share/hiop/cmake/HiOpConfig.cmake` as `Magma`
- `<install_dir>/share/hiop/cmake/HiOpTargets.cmake` as `HiOp::Magma` (`HiOp::Magma -> HiOp::hiop_tpl -> HiOp::HiOp`)

### Correct TPL Exporting + Remove Duplication

As seen with the Magma example, it seems as though I am exporting a Magma target that is identical under two different CMake names: `Magma` and `HiOp::Magma`

@balos1 this is where you could be of most help I assume:

How do we export TPL targets? Using Magma as an example, do we:
- Explicitly define interface in `cmake/HiOpConfig.cmake.in`
- Rely on namespace based exporting with `install(TARGETS Magma EXPORT hiop-targets)`

What determines the appropriate export configuration, and why?


src/CMakeLists.txt
```diff
@@ -27,16 +27,13 @@ foreach(libtype shared static)
       )
     target_link_libraries(hiop_${libtype} PUBLIC hiop_tpl)
     set_target_properties(hiop_${libtype} PROPERTIES OUTPUT_NAME hiop)
-    install(TARGETS hiop_${libtype} DESTINATION lib)
+    install(TARGETS hiop_${libtype} DESTINATION lib EXPORT hiop-targets)
   endif()
 endforeach()
 
 # Create default library pointed to by HiOp::HiOp
 add_library(HiOp::HiOp ALIAS ${hiop_default_library_name})
```
ashermancinelli, line 35: Will this target be exported? I think we want this to be the cmake interface for the end user so they don't concern themselves with shared/static unless they need to.

tests/CMakeLists.txt
```diff
@@ -28,23 +28,17 @@ endif()
 
 # Build vector test
 add_executable(testVector ${testVector_SRC})
-target_link_libraries(testVector PRIVATE hiop)
+target_link_libraries(testVector PRIVATE HiOp::HiOp)
 
 # Build matrix test
 add_executable(testMatrixDense ${testMatrix_SRC})
-target_link_libraries(testMatrixDense PRIVATE hiop)
+target_link_libraries(testMatrixDense PRIVATE HiOp::HiOp)
```
ashermancinelli, line 35: +1 for all these updates

src/CMakeLists.txt
CameronRutherford: This will be exported in `<install_dir>/share/hiop/cmake/HiOpTargets.cmake`. All TPLs and libraries that have the `EXPORT hiop-targets` are exported similarly. When I built and installed, I had the following exported for these targets:

```CMake
# Create imported target HiOp::hiop_shared
add_library(HiOp::hiop_shared SHARED IMPORTED)

set_target_properties(HiOp::hiop_shared PROPERTIES
  INTERFACE_LINK_LIBRARIES "HiOp::hiop_tpl;HiOp::Magma"
)

# Create imported target HiOp::hiop_static
add_library(HiOp::hiop_static STATIC IMPORTED)

set_target_properties(HiOp::hiop_static PROPERTIES
  INTERFACE_LINK_LIBRARIES "HiOp::hiop_tpl;HiOp::Magma"
)
```

CameronRutherford: `HiOp::hiop_tpl` is also created as a target, and it is linked against all the libraries that `hiop_tpl` links against.

ashermancinelli: Right, I don't really want users interacting with `_shared` or `_static` unless they have to - is `HiOp::HiOp` exported?

pelesh:
> @pelesh pointed out that the output libhiop.so should link against all libraries in it's r path. However:
> ```
> [rcruther <build_dir>/src]$ ldd libhiop.so  | grep magma
>     libmagma.so => not found
> ```
> This could be something that I am doing wrong in CMake when linking against Magma, or in our find module. This could also be taken care of in #220, or wrapped in a separate issue.

I would suggest submitting a separate issue and linking it to this PR. The #220 does not document the issue with Magma library rpath you've just found.

CameronRutherford:
Closing in favor of #236 @pelesh

Issue #233, opened 2021-05-12
Copy eq and ineq to full size cons body/multipliers


cnpetra:
Externalize the host code to matrix and vector kernels (may need to add new method in the vector and matrix interfaces, as well as, CPU and RAJA implementations)

In class `hiopNlpFormulation`, the following functions only works on host:

1. `copy_EqIneq_to_cons`

2. `copy_cons_to_EqIneq` 

Need to move these functions into matrix/vector kernels.

nychiang:
fixed in PR #235

Issue #150, opened 2020-12-10
Implicit conversions long long <--> int in linear algebra module


pelesh:
There are several implicit conversions from `int` to `long long` in the linear algebra module and related unit tests. This may cause some issues down the road. This has been confirmed to cause (at least part of the) problems in #103. 

A number of these conversions happens in sparse matrix methods, where arguments of the method are of `long long` type, while sparse matrix rows and columns are indexed by `int`s. None of these should be difficult to fix.

Perhaps it would be a good idea to add flag `-Wconversion` to the HiOp build configuration and ensure HiOp builds without  warnings. 

CC @cnpetra @nychiang @ashermancinelli

ashermancinelli:
I believe we have some option for developer warnings. Are you suggesting enabling some warnings regardless of whether that option is enabled or not?

cnpetra:
not sure if it;s clear why the interfaces are with `long long` (on short I would like to keep them like this for now)... the good news is these `long long` only store numbers that also fit in `int` and conversions from `long long` (not involving pointers) should not pose a problem

pelesh:
> I believe we have some option for developer warnings. Are you suggesting enabling some warnings regardless of whether that option is enabled or not?

It looks to me we dropped `-Wconversion` flag from the developer's warnings. I think we only check for `float` <-> `double` conversiosn. I may be wrong.

pelesh:
> not sure if it;s clear why the interfaces are with `long long` (on short I would like to keep them like this for now)... the good news is these `long long` only store numbers that also fit in `int` and conversions from `long long` (not involving pointers) should not pose a problem

For as long as we make sure all conversions are _explicit_, we should be fine. Potential issues may arise with mapping vectors:
```c++
  // keep track of the constraints indexes in the original, user's formulation
  long long *cons_eq_mapping_, *cons_ineq_mapping_; 
```

ashermancinelli:
> > I believe we have some option for developer warnings. Are you suggesting enabling some warnings regardless of whether that option is enabled or not?
> 
> It looks to me we dropped `-Wconversion` flag from the developer's warnings. I think we only check for `float` <-> `double` conversiosn. I may be wrong.

No, you're right I didn't realize we dropped it. In the PR to address this I think we should add it back and turn `HIOP_DEVELOPER_MODE=ON` in the build script so we at least see these warnings in CI.

cnpetra:
[...]

PR #235, opened 2021-05-17
Issue 232 233 fix

### LINKED ISSUE #232: FR (sparse) problem contains host only code
Externalize the host code to matrix and vector kernels (may need to add new method in the vector and matrix interfaces, as well as, CPU and RAJA implementations)

More specifically, in class `hiopFRProbSparse ` :
- Jacobian method
- `get_vars_info`, `get_cons_info` -> the type for loops
- `eval_f` -> for loop
- Hessian method

nychiang:
This PR resolves #232 and #233.
It moves some customized functions into the hiop vector/matrix kernel, in order to support multi-threads/GPU enviroment.

This is still a draft!

### DIFF
--- a/src/LinAlg/hiopLinAlgFactory.cpp
+++ b/src/LinAlg/hiopLinAlgFactory.cpp
@@ -109,7 +109,7 @@ hiopVector* LinearAlgebraFactory::createVector(
  * Creates int vector with operator new by default, RAJA vector when memory space
  * is specified.
  */
-hiopVectorInt* LinearAlgebraFactory::createVectorInt(int size)
+hiopVectorInt* LinearAlgebraFactory::createVectorInt(hiopInt size)
 {
   if(mem_space_ == "DEFAULT")
   {
--- a/src/LinAlg/hiopLinAlgFactory.hpp
+++ b/src/LinAlg/hiopLinAlgFactory.hpp
@@ -78,7 +78,7 @@ public:
   /**
    * @brief Static method to create local int vector.
    */
-  static hiopVectorInt* createVectorInt(int size);
+  static hiopVectorInt* createVectorInt(hiopInt size);
 
   /**
    * @brief Static method to create a dense matrix.
--- a/src/LinAlg/hiopMatrixRajaSparseTriplet.hpp
+++ b/src/LinAlg/hiopMatrixRajaSparseTriplet.hpp
@@ -246,7 +246,7 @@ public:
                           const hiopMatrixSparse& Jac_d,
                           int* iJacS,
                           int* jJacS,
-                          double* MJacS);
+                          double* MJacS) {assert("not implemented"&&0);}
 
   virtual void set_Hess_FR(const hiopMatrixSparse& Hess,
                            int* iHSS,
@@ -401,7 +401,7 @@ public:
                            int* iHSS,
                            int* jHSS,
                            double* MHSS,
-                           const hiopVector& add_diag);
+                           const hiopVector& add_diag){assert("not implemented"&&0);}
 
 };
 } //end of namespace
--- a/src/LinAlg/hiopMatrixRajaSparseTriplet.hpp
+++ b/src/LinAlg/hiopMatrixRajaSparseTriplet.hpp
@@ -241,8 +241,22 @@ public:
     assert(false && "not needed / implemented");
   }
 
-  virtual long long numberOfOffDiagNonzeros() {assert("not implemented"&&0);return 0;};
-      
+  virtual long long numberOfOffDiagNonzeros() const {assert("not implemented"&&0);return 0;};
+  virtual void set_Jac_FR(const hiopMatrixSparse& Jac_c,
+                          const hiopMatrixSparse& Jac_d,
+                          int* iJacS,
+                          int* jJacS,
+                          double* MJacS);
+
+  virtual void set_Hess_FR(const hiopMatrixSparse& Hess,
+                           int* iHSS,
+                           int* jHSS,
+                           double* MHSS,
+                           const hiopVector& add_diag)
+  {
+    assert(false && "not needed / implemented");
+  }
+
   virtual hiopMatrixSparse* alloc_clone() const;
   virtual hiopMatrixSparse* new_copy() const;
 
@@ -372,10 +386,22 @@ public:
     return true;
   }
   
-  virtual long long numberOfOffDiagNonzeros() {
+  virtual long long numberOfOffDiagNonzeros() const {
     assert(false && "not needed / implemented");
     return 0;
   }
+
+  virtual void set_Jac_FR(const hiopMatrixSparse& Jac_c,
+                          const hiopMatrixSparse& Jac_d,
+                          int* iJacS,
+                          int* jJacS,
+                          double* MJacS) {assert("not implemented"&&0);};
+
+  virtual void set_Hess_FR(const hiopMatrixSparse& Hess,
+                           int* iHSS,
+                           int* jHSS,
+                           double* MHSS,
+                           const hiopVector& add_diag);
 
 };
 } //end of namespace
--- a/src/LinAlg/hiopMatrixSparse.hpp
+++ b/src/LinAlg/hiopMatrixSparse.hpp
@@ -198,7 +198,21 @@ public:
   virtual const int* i_row() const = 0;
   virtual const int* j_col() const = 0;
   virtual const double* M()  const = 0;
-  virtual long long numberOfOffDiagNonzeros() = 0;
+  virtual long long numberOfOffDiagNonzeros() const = 0;
+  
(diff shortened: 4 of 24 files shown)


nychiang:
@pelesh @cnpetra 
tried to move some customized functions from `hiopFRProb` to `hiopVector` and `hiopMatrix`.
I am still using raw pointers in `hiopFRProb`, in order to reuse the standard C++ interface and consequently the other functions in `hiopAlgIPM`,  `hiopNlpFormulation` and 'hiopNlpTransformer". 
Similar to `nlpMDS_ex4_raja`, all the problem are passed by raw pointers from users. 
I think this change should be fine, but it doesn't look nice. 
Please have a look of it. After you two agree with the changes, I will complete the corresponding unit tests for the new functions I introduced in the vector/matrix kernel.

nychiang:
Unlike the original code where I extended the matrices by a loop of nonzeros, now I tried to extend the matrix in a loop of rows. This will resolve the multi-thread issue and hence the RAJA implementation should be similar.

nychiang:
@pelesh @cnpetra @ashermancinelli 
In order to remove CPU-based raw pointers [`cons_eq_mapping_` and `cons_ineq_mapping_`](https://github.com/LLNL/hiop/blob/177d971aee13f7cf3ba3292e050948c5238ea769/src/Optimization/hiopNlpFormulation.cpp#L398-L421), I plan to declare them as hiop integer vector of type long long.
Then I can reuse the existing interface which requires `long long*` as an input. 
For example, I can change the following code
https://github.com/LLNL/hiop/blob/177d971aee13f7cf3ba3292e050948c5238ea769/src/Optimization/hiopNlpFormulation.cpp#L638-L642
to 
```
bool bret = interface_base.eval_cons(nlp_transformations.n_pre(),
                                     n_cons,n_cons_eq,
                                     cons_eq_mapping_->local_data_const(),
                                     xx->local_data_const(), 
                                     new_x,
                                     cc->local_data());
```

Should I just change the source code in `hiopVectorInt`, to use `long long` array instead of `int` array to save the data internal?
Otherwise I can create a new class `hiopVectorLongLong`, but it mainly did the same thing as `hiopVectorInt`.

Any preference?

ashermancinelli:
@pelesh Is there any reason we don't template the integer type for hiopVectorInt?

pelesh:
> Should I just change the source code in `hiopVectorInt`, to use `long long` array instead of `int` array to save the data internal?
> Otherwise I can create a new class `hiopVectorLongLong`, but it mainly did the same thing as `hiopVectorInt`.
> 
> Any preference?

I think the issue here is that HiOp sparse matrices store their integer data (matrix indices) as `int`s, but the matrix API uses `long long` type to access matrix elements. There are implicit conversions at several places in the code because of that.

The rationale for using `int` type in the integer vector was to (try to) avoid implicit conversions, but I realize now this is a bit more involved issue.

Before replacing one hard-wired type with another, I would check if creating class template for `hiopVectorInt` would be an easy option. It is a small simple class, so that should not be too difficult and it would give us some flexibility in the future too.

nychiang:
Thanks for the quick responses.
I think templating the integer type may be the best solution here.
Here is the short path to the relevant issue #150.

nychiang:
@pelesh @ashermancinelli 
I use `typedef` in `hiopVectorInt` to check if creating a class template is an easy option.
However, the new commit gets some runtime CUDA error on Mariana, while it pass all the unit tests on newell and ORNL pipelines. Seem I met the same `long long - int` conversion error, which was reported in the last comment in #150.

In function [`copy_from_two_vec_w_pattern` ](https://github.com/LLNL/hiop/blob/26c860f16f6afa55fdc347741443e8fb2ebbab59/src/LinAlg/hiopVectorRajaPar.cpp#L517-L526), I need to have access to `long long` arrays `id1` and `id2`. I believe this fails the RAJA function on mariana.

Is there any way to use long long inside RAJA?

pelesh:
> In function [`copy_from_two_vec_w_pattern` ](https://github.com/LLNL/hiop/blob/26c860f16f6afa55fdc347741443e8fb2ebbab59/src/LinAlg/hiopVectorRajaPar.cpp#L517-L526), I need to have access to `long long` arrays `id1` and `id2`. I believe this fails the RAJA function on mariana.
> 
> Is there any way to use long long inside RAJA?

I don't think that RAJA has any issues with dealing with `long long` type. There may be an issue with possible implicit type conversion in CUDA. I cc @davidbeckingsale as he might have a better advice here.

@nychiang, could you provide more detailed info how to reproduce this issue? When I build `copy_from_two_vec_wpattern`, I get the unit test to pass, but the example 4 hangs, whether I cast to `int` or use `long long`.

nychiang:
@pelesh I haven't implement the unit tests for `copy_from_two_vec_w_pattern` yet. As you mentioned, it fails no matter if we cast it to`int` or `long long`, but only on Mariana.
Before ex4 hangs, I got error message:
20: CUDAassert: an illegal memory access was encountered /qfs/projects/exasgd/spack-install/linux-centos7-broadwell/gcc-7.3.0/raja/0.13.0-pwmbk/include/RAJA/policy/cuda/MemUtils_CUDA.hpp 183
20: terminate called after throwing an instance of 'std::runtime_error'
20:   what():  CUDAassert
20: [dl07:13519] *** Process received signal ***
20: [dl07:13519] Signal: Aborted (6)
20: [dl07:13519] Signal code:  (-6)
20: [dl07:13519] [ 0] /usr/lib64/libpthread.so.0(+0xf630)[0x7fd06d91e630]
20: [dl07:13519] [ 1] /usr/lib64/libc.so.6(gsignal+0x37)[0x7fd057619387]
20: [dl07:13519] [ 2] /usr/lib64/libc.so.6(abort+0x148)[0x7fd05761aa78]
20: [dl07:13519] [ 3] /share/apps/gcc/7.3.0/lib64/libstdc++.so.6(_ZN9__gnu_cxx27__verbose_terminate_handlerEv+0x125)[0x7fd057f5b205]
20: [dl07:13519] [ 4] /share/apps/gcc/7.3.0/lib64/libstdc++.so.6(+0x8efd6)[0x7fd057f58fd6]
20: [dl07:13519] [ 5] /share/apps/gcc/7.3.0/lib64/libstdc++.so.6(+0x8f021)[0x7fd057f59021]
20: [dl07:13519] [ 6] /share/apps/gcc/7.3.0/lib64/libstdc++.so.6(+0x8f264)[0x7fd057f59264]
20: [dl07:13519] [ 7] /people/svcexasgd/gitlab/66227/build/src/Drivers/nlpMDS_ex4_raja.exe(_Z19RAJA_ABORT_OR_THROWPKc+0x64)[0x4308ab]
20: [dl07:13519] [ 8] /people/svcexasgd/gitlab/66227/build/src/Drivers/nlpMDS_ex4_raja.exe(_ZN4RAJA10cudaAssertE9cudaErrorPKcib+0x67)[0x4312f4]
20: [dl07:13519] [ 9] /people/svcexasgd/gitlab/66227/build/src/libhiop.so(_ZN4hiop17hiopVectorRajaPar27copy_from_two_vec_w_patternERKNS_10hiopVectorERKNS_13hiopVectorIntES3_S6_+0xb70)[0x7fd06e37bfe2]
20: [dl07:13519] [10] /people/svcexasgd/gitlab/66227/build/src/libhiop.so(_ZN4hiop18hiopNlpFormulation21user_callback_iterateEiddRKNS_10hiopVectorES3_S3_S3_S3_S3_S3_S3_ddddddi+0x225)[0x7fd06e3f87d1]
20: [dl07:13519] [11] /people/svcexasgd/gitlab/66227/build/src/libhiop.so(_ZN4hiop22hiopAlgFilterIPMNewton3runEv+0xd7e)[0x7fd06e412848]
20: [dl07:13519] [12] /people/svcexasgd/gitlab/66227/build/src/Drivers/nlpMDS_ex4_raja.exe[0x415e45]
20: [dl07:13519] [13] /usr/lib64/libc.so.6(__libc_start_main+0xf5)[0x7fd057605555]
20: [dl07:13519] [14] /people/svcexasgd/gitlab/66227/build/src/Drivers/nlpMDS_ex4_raja.exe[0x414f49]
20: [dl07:13519] *** End of error message ***

nychiang:
Create a new PR, using @pelesh's temporary fix.
Would like to close this PR, and move our discussions there.

pelesh:
> Create a new PR #238 , using @pelesh's temporary fix.
> Would like to close this PR, and move our discussions there.

Please keep this PR. The purpose of the branch I created was to verify that the two kernels that are failing are producing correct indices and to narrow down the problem to casting from `long long` to `int` on GPUs. The instrumentation in `issue-232-233-237-fix` branch was never meant to be used as a workaround.

cnpetra (approved):

PR #238, opened 2021-05-21
Issue 232 233 237 fix

nychiang:
This PR replaces PR #235, since @pelesh adds a temporary fix, and now the code works on all the platform.

Now I define `hiopInt` as type of `long long` and it works fine.
However, if I define it as `int` then we fail to compile the code, since we still use `long long` in the interface. 
For example
```
virtual bool eval_cons(const long long& n, const long long& m, 
		       const long long& num_cons, const long long* idx_cons,  
		       const double* x, bool new_x, 
		       double* cons)=0;
```

Things we need to discuss here are:
1. how to define/use temporary integer type in hiop
2. should we address this issue in this PR

@pelesh @cnpetra

### DIFF
--- a/src/LinAlg/hiopLinAlgFactory.cpp
+++ b/src/LinAlg/hiopLinAlgFactory.cpp
@@ -109,7 +109,7 @@ hiopVector* LinearAlgebraFactory::createVector(
  * Creates int vector with operator new by default, RAJA vector when memory space
  * is specified.
  */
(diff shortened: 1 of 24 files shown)


pelesh (changes requested):
imho this branch should not be merged. The code I added was for debugging purposes and should not be included even as a workaround. I am sorry if my earlier communication was misleading. I suggest closing this MR.

PR #239, opened 2021-05-27
Feasibility restoration problem for MDS system

### LINKED ISSUE #255: Another bug in RAJA function that only happens on marianas
Branch `fr-mds-dev` passes all the unit test on newell and ascent, but gets problems again on marianas.
This is not about the `long long` --> `int` conversion bug, and I created branch `raja-marianas-bug-fix` to reproduce and narrow down the bug.

in short, I received CUDA assertion error [on this line ](https://github.com/LLNL/hiop/blob/96f793f6311c0950bc9e360e4a78f8c3d8a5942e/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp#L1138).
Note that I add this fake line to show the problem is not caused by accessing array out of bounds. 
'this' matrix is initialized [here](https://github.com/LLNL/hiop/blob/96f793f6311c0950bc9e360e4a78f8c3d8a5942e/tests/testMatrixSparse.cpp#L246), given a space to save 105 non-zeros.

The error message is " CUDAassert: an illegal memory access was encountered /qfs/projects/exasgd/spack-install/linux-centos7-broadwell/gcc-7.3.0/raja/0.13.0-pwmbk/include/RAJA/policy/cuda/MemUtils_CUDA.hpp 183"

I am guessing here we have another implicit type conversion that fails CUDA.

PS: FYI, the RAJA::Index_type type is an alias to the C++ type std::ptrdiff_t, and std::ptrdiff_t is the signed integer type of the result of subtracting two pointers.

@pelesh @cnpetra @ashermancinelli

nychiang:
This PR implements the feasibility restoration problem for MDS system.
It also:
1. implements the interface for warm-start.
2. fixes memory leak in the current unit test
3. implements missing functions in RAJA matrix kernel 
4. also closes #255

### DIFF
--- a/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
+++ b/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
@@ -601,6 +601,44 @@ void hiopMatrixRajaSparseTriplet::copyFrom(const hiopMatrixSparse& dm)
   assert(false && "this is to be implemented - method def too vague for now");
 }
 
+/// @brief copy to 3 arrays.
+/// @pre these 3 arrays are not nullptr
+void hiopMatrixRajaSparseTriplet::copy_to(int* irow, int* jcol, double* val)
+{
+  assert(irow && jcol && val);
+  auto& resmgr = umpire::ResourceManager::getInstance();
+  resmgr.copy(irow, iRow_);
+  resmgr.copy(jcol, jCol_);
+  resmgr.copy(val, values_);
+}
+
+void hiopMatrixRajaSparseTriplet::copy_to(hiopMatrixDense& W)
+{
+  assert(W.m() == nrows_);
+  assert(W.n() == ncols_);
+  W.setToZero();
+  
+  RAJA::View<double, RAJA::Layout<2>> WM(W.local_data(), W.m(), W.n());
+
+  if(row_starts_host==NULL)
+    row_starts_host = allocAndBuildRowStarts();
+  assert(row_starts_host);
+  
+  int num_rows = this->nrows_;
+  int* idx_start = row_starts_host->idx_start_;
+  int* jCol = jCol_;
+  double* values = values_;
+  RAJA::forall<hiop_raja_exec>(RAJA::RangeSegment(0, this->nrows_),
+    RAJA_LAMBDA(RAJA::Index_type i)
+    {
+      for(int k=idx_start[i]; k<idx_start[i+1]; k++)
+      {
+        WM(i, jCol[k]) += values[k];
+      }
+    }
+  );
+}
+
 #ifdef HIOP_DEEPCHECKS
 /// @brief Ensures the rows and column triplet entries are ordered.
 bool hiopMatrixRajaSparseTriplet::checkIndexesAreOrdered() const
@@ -842,6 +880,7 @@ hiopMatrixRajaSparseTriplet::allocAndBuildRowStarts() const
   // build rsi on the host, then copy it to the device for simplicity
   int it_triplet = 0;
   rsi->idx_start_[0] = 0;
+
   for(int i = 1; i <= this->nrows_; i++)
   {
     rsi->idx_start_[i]=rsi->idx_start_[i-1];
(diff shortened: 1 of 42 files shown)


nychiang:
I will squash the commits before merging it.

nychiang:
@jwang125 I use `size_type` to replace `long long` in `hiopInterfacePrimalDecomp`.

nychiang:
I replaced `long long` by `index_type` and `size_type` in the tex files.
It is out of scope, but it's a quick fix for me. I search for the keyword `long long` and replace them one by one manually.

cnpetra:
> @jwang125 I use `size_type` to replace `long long` in `hiopInterfacePrimalDecomp`.

It's my doing as I forgot to replace those when I rebased Frank's branch

nychiang:
Fix the RAJA bug to access member variables. Thanks @ashermancinelli for all the valgrind testings and suggestions.
Now this branch passes all the CI tests.
Ready to review.

src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
```diff
@@ -1013,6 +1052,171 @@ hiopMatrixRajaSparseTriplet::RowStartsInfo::~RowStartsInfo()
...
+  int n_d = J_d.ncols_;
+  assert(n_c == n_d);
+  assert(ncols_ == n_c + 2*m_c + 2*m_d);
+
+  int nnz_Jac_c_new = nnz_Jac_c + 2*m_c;
+  int nnz_Jac_d_new = nnz_Jac_d + 2*m_d;
+
+  assert(nnz_ == nnz_Jac_c_new + nnz_Jac_d_new);
+  
+  if(J_c.row_starts_host == nullptr){
```
cnpetra, line 1088: space before bracket

src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
```diff
@@ -1013,6 +1052,171 @@ hiopMatrixRajaSparseTriplet::RowStartsInfo::~RowStartsInfo()
...
+    J_c.row_starts_host = J_c.allocAndBuildRowStarts();
+  }
+  assert(J_c.row_starts_host);
+  int* Jc_row_st = J_c.row_starts_host->idx_start_;
+
+  if(J_d.row_starts_host == nullptr){
```
cnpetra, line 1094: space before bracket

tests/LinAlg/matrixTestsDenseRowMajor.cpp
```diff
@@ -156,6 +156,16 @@ const real_type* MatrixTestsDenseRowMajor::getLocalDataConst(hiop::hiopMatrixDen
   return amat->local_data_const();
 }
 
+/// Returns const pointer to local data block of matrix _A_.
+real_type* MatrixTestsDenseRowMajor::getLocalData(hiop::hiopMatrixDense* A)
+{
+  auto* amat = dynamic_cast<hiop::hiopMatrixDense*>(A);
+  if(amat == nullptr)
+    THROW_NULL_DEREF;
```
cnpetra, line 164: nothing bad with this, but we haven't use exceptions in the core of hiop. I would just return null_ptr...

tests/LinAlg/matrixTestsRajaDense.cpp
```diff
@@ -162,6 +162,15 @@ const real_type* MatrixTestsRajaDense::getLocalDataConst(hiop::hiopMatrixDense*
   return amat->local_data_const();
 }
 
+/// Returns pointer to local data block of matrix _A_ in current memory space.
+real_type* MatrixTestsRajaDense::getLocalData(hiop::hiopMatrixDense* A)
+{
+  auto* amat = dynamic_cast<hiop::hiopMatrixRajaDense*>(A);
+  if(amat == nullptr)
```
cnpetra, line 169: same here

cnpetra (changes requested):
lgtm... I left a couple of very small comments

tests/LinAlg/matrixTestsDenseRowMajor.cpp
nychiang: I probably would keep this line, in order to keep consistence [with the other code in this file.](https://github.com/LLNL/hiop/blob/c559149f6cef1b71ebdea03cb56f820f29c2fde1/tests/LinAlg/matrixTestsRajaSparseTriplet.cpp)
`THROW_NULL_DEREF` is [defined here](https://github.com/LLNL/hiop/blob/c559149f6cef1b71ebdea03cb56f820f29c2fde1/tests/LinAlg/testBase.hpp#L58)

src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
nychiang: Fixed. I will wait for others' feedback before pushing the commit.

tests/LinAlg/matrixTestsDenseRowMajor.cpp
ashermancinelli: We kept the exceptions out of hiop proper, but we found them to be really helpful when diagnosing bugs in our test suite and we probably won't miss the 2% performance hit. We can revert this, but I don't think we need to imo

cnpetra: ok

cnpetra (approved):

tests/LinAlg/matrixTestsSparse.hpp
```diff
@@ -939,6 +939,118 @@ class MatrixTestsSparse : public TestBase
...
+    C.setToConstant(C_val);
+    D.setToConstant(D_val);
+
+
+    const auto* C_iRow = getRowIndices(&C);
+    const auto* C_jCol = getColumnIndices(&C);
+    auto C_nnz = C.numberOfNonzeros();
+    const auto* D_iRow = getRowIndices(&D);
+    const auto* D_jCol = getColumnIndices(&D);
+    auto D_nnz = D.numberOfNonzeros();
```
pelesh, line 1002: I would really spell out data types here rather than using `auto`.

nychiang:
@cnpetra @pelesh I have addressed all the comments in the latest commit. PNNL CI is not working right now, but I think this easy change should be able to pass all the unit tests.

Issue #237, opened 2021-05-20
RAJA vector kernel copy_from_two_vec_w_pattern segfaults on GPU


pelesh:
Vector kernel [`hiopVectorRajaPar::copy_from_two_vec_w_pattern`](https://github.com/LLNL/hiop/blob/issue-232-233-fix/src/LinAlg/hiopVectorRajaPar.cpp#L490) segfaults when executed on GPU. The issue can be reproduced in unit test [vector_copy_from_two_vec](https://github.com/LLNL/hiop/blob/issue-232-233-fix/tests/LinAlg/vectorTests.hpp#L384).

The backtrace is:
```shell
(gdb) bt
#0  0x00002aaac114b387 in raise () from /usr/lib64/libc.so.6
#1  0x00002aaac114ca78 in abort () from /usr/lib64/libc.so.6
#2  0x00002aaac090b205 in __gnu_cxx::__verbose_terminate_handler ()
    at ../../../../libstdc++-v3/libsupc++/vterminate.cc:95
#3  0x00002aaac0908fd6 in __cxxabiv1::__terminate (handler=<optimized out>)
    at ../../../../libstdc++-v3/libsupc++/eh_terminate.cc:47
#4  0x00002aaac0909021 in std::terminate () at ../../../../libstdc++-v3/libsupc++/eh_terminate.cc:57
#5  0x00002aaac0909264 in __cxxabiv1::__cxa_throw (obj=obj@entry=0x10275b0, 
    tinfo=0x696740 <_ZTISt13runtime_error@@GLIBCXX_3.4>, 
    dest=0x2aaac091e450 <std::runtime_error::~runtime_error()>)
    at ../../../../libstdc++-v3/libsupc++/eh_throw.cc:93
#6  0x00002aaaaad30d0a in RAJA_ABORT_OR_THROW (str=0x2aaaaad9f911 "CUDAassert")
    at /.../src/spack/opt/spack/linux-centos7-broadwell/gcc-7.3.0/raja-0.12.1-sflepy572gx7kfwoz2h7gjva76gjezd3/include/RAJA/util/macros.hpp:152
#7  RAJA::cudaAssert (code=<optimized out>, line=line@entry=183, abort=true, 
[...]

PR #242, opened 2021-05-31
replaced `long long` with generic `hiop::int_type`

cnpetra:
now using `hiop::int_type` instead of `long long` 

addresses Issues #150 and  #237 

there are no warnings under `-Wconversion` and passes all CI pipelines.


cnpetra:
passes all test except  VectorTest on device memory space

```--- FAIL: Test vectorSetElement on rank 0
```

cnpetra:
@nychiang : I based it on your `issue-232-233-fix`, I am curious if fixes the crashes on marianna

cnpetra:
> passes all test except VectorTest on device memory space
> 
> ```
> --- FAIL: Test vectorSetElement on rank 0
> ```

this issue is caused by an incorrect implementation of the operator[] in hiopVectorIntRaja... will open a separate issue for it. I pushed a temp fix https://github.com/LLNL/hiop/pull/242/commits/da4a3684d11f5be0d7d828d5296bd596f1593f5c so that all the tests are passed

pelesh:
I would perhaps just suggest using name such as `index_type` or `ordinal_type` to specify this is a type for linear algebra objects indices. This is what we would (ideally) want to change with a flip of a switch. We probably don't want other `int`s in the code to be affected by setting vector and matrix index type.

cnpetra:
`int_type` is used both as "index" and "length"/size. Ideally should have `index_type` and `size_type` . `ordinal_type` is also not suitable since the name is too generic...

pelesh:
Taking this off the public github… As far as I can see, the size and the index type should be the same as it is in STL. Size is max index + 1.

imho any name is better than int_type 😃

Cheers,
Slaven

From: Cosmin G Petra ***@***.***>
Reply-To: LLNL/hiop ***@***.***>
Date: Tuesday, June 1, 2021 at 09:09
To: LLNL/hiop ***@***.***>
Cc: "Peles, Slaven" ***@***.***>, Mention ***@***.***>
Subject: Re: [LLNL/hiop] replaced `long long` with generic `hiop::int_type` (#242)

Check twice before you click! This email originated from outside PNNL.


int_type is used both as "index" and "length"/size. Ideally should have index_type and size_type . ordinal_type is also not suitable since the name is too generic...

src/Optimization/hiopNlpFormulation.cpp
```diff
@@ -282,9 +282,9 @@ bool hiopNlpFormulation::finalizeInitialization()
 
   dFixedVarsTol = fixedVarTol;
   
-  long long nfixed_vars=nfixed_vars_local;
+  int_type nfixed_vars=nfixed_vars_local;
 #ifdef HIOP_USE_MPI
-  int ierr = MPI_Allreduce(&nfixed_vars_local, &nfixed_vars, 1, MPI_LONG_LONG, MPI_SUM, comm); 
+  int ierr = MPI_Allreduce(&nfixed_vars_local, &nfixed_vars, 1, MPI_INT, MPI_SUM, comm);
```
pelesh, line 287: Probably good idea to implement index traits class to manage issues like this. That should be straightforward.

src/Optimization/hiopAlgFilterIPM.cpp
```diff
@@ -1965,8 +1967,8 @@ int hiopAlgFilterIPMBase::apply_second_order_correction(hiopKKTLinSys* kkt,
                                                         double &grad_phi_dx,
                                                         int &num_adjusted_bounds)
 {
-  int max_soc_iter = nlp->options->GetNumeric("max_soc_iter");
-  int kappa_soc = nlp->options->GetNumeric("kappa_soc");
+  int max_soc_iter = nlp->options->GetInteger("max_soc_iter");
+  double kappa_soc = nlp->options->GetNumeric("kappa_soc");
```
pelesh, line 1971: This seems to be fixing a bug out of this PR scope.

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -1871,43 +1871,43 @@ void hiopVectorRajaPar::copyFromDev() const
...
     });
 
-  long long nrm = sum.get();
+  int_type nrm = sum.get();
 
 #ifdef HIOP_USE_MPI
-  long long nrm_global;
-  int ierr = MPI_Allreduce(&nrm, &nrm_global, 1, MPI_LONG_LONG, MPI_SUM, comm_);
+  int_type nrm_global;
+  int ierr = MPI_Allreduce(&nrm, &nrm_global, 1, MPI_INT, MPI_SUM, comm_);
```
pelesh, line 1888: Having index traits to set correct MPI type would be desirable.

src/LinAlg/hiopVectorPar.cpp
```diff
@@ -1031,37 +1031,37 @@ void hiopVectorPar::print(FILE* file, const char* msg/*=NULL*/, int max_elems/*=
...
     if(data_[i]<val) {
       ret_num++;
     }
   }
 
 #ifdef HIOP_USE_MPI
-  long long ret_num_global;
-  int ierr=MPI_Allreduce(&ret_num, &ret_num_global, 1, MPI_LONG_LONG, MPI_SUM, comm_);
+  int_type ret_num_global;
+  int ierr=MPI_Allreduce(&ret_num, &ret_num_global, 1, MPI_INT, MPI_SUM, comm_);
```
pelesh, line 1045: Hard-wired MPI_INT type.

src/LinAlg/hiopMatrixSparseTriplet.cpp
```diff
@@ -706,11 +706,11 @@ void hiopMatrixSparseTriplet::print(FILE* file, const char* msg/*=NULL*/,
     if(NULL==msg) {
       if(numranks>1) {
         fprintf(file,
-                "matrix of size %lld %lld and nonzeros %lld, printing %d elems (on rank=%d)\n",
+                "matrix of size %d %d and nonzeros %d, printing %d elems (on rank=%d)\n",
                 m(), n(), numberOfNonzeros(), max_elems, myrank_);
       } else {
         fprintf(file,
-                "matrix of size %lld %lld and nonzeros %lld, printing %d elems\n",
+                "matrix of size %d %d and nonzeros %d, printing %d elems\n",
```
pelesh, line 713: Is there a reason not to use fstream here? The `<<` operator is overloaded for all int types.

src/LinAlg/hiopMatrixComplexSparseTriplet.cpp
```diff
@@ -475,10 +475,10 @@ namespace hiop
       
       if(NULL==msg) {
 	if(numranks>1)
-	  fprintf(file, "matrix of size %lld %lld and nonzeros %lld, printing %d elems (on rank=%d)\n", 
+	  fprintf(file, "matrix of size %d %d and nonzeros %d, printing %d elems (on rank=%d)\n", 
 		  m(), n(), numberOfNonzeros(), max_elems, myrank_);
 	else
-	  fprintf(file, "matrix of size %lld %lld and nonzeros %lld, printing %d elems\n", 
+	  fprintf(file, "matrix of size %d %d and nonzeros %d, printing %d elems\n",
```
pelesh, line 481: Consider using fstream instead of fpritntf.

src/LinAlg/hiopLinAlgFactory.hpp
```diff
@@ -70,54 +70,46 @@ class LinearAlgebraFactory
...
+  static double* createRawArray(int_type n);
 
   /**
    * @brief Static method to delete a raw C array
    */
   static void deleteRawArray(double* a);
 
   /// Method to set memory space ID
-  static void set_mem_space(const std::string mem_space);
+  static void set_mem_space(const std::string& mem_space);
```
pelesh, line 112: Out of scope, but good catch. :)

src/Utils/hiopOptions.cpp
```diff
@@ -233,7 +233,7 @@ void hiopOptions::registerOptions()
 
   // second order correction
   {
-    registerNumOption("max_soc_iter", 4, 0, 1e6,
+    registerIntOption("max_soc_iter", 4, 0, 1000000,
```
ashermancinelli, line 236: This also seems a tad out of scope.

ashermancinelli (commented):
I'd approve except that I agree with @pelesh that some of the options-related changes seem a bit unrelated to the core of this PR

src/LinAlg/hiopMatrixSparseTriplet.cpp
ashermancinelli: I think fstream is usually more appropriate, but if we do any real formatting with this, fstreams would be pretty painful and verbose

src/Interface/hiopInterface.h
```diff
@@ -1,38 +1,44 @@
 // The C interface header used by the user. This needs a detailed user documentation.
 
+/** The hiop_int_type type defined below needs to match the C++ type hiop::int_type defined for 
+ * HiOp in the C++ header file hiop_defs.hpp
+ */
+typedef int hiop_int_type;
```
pelesh, line 6: Would it be possible to use `hiop::int_type` here?
```suggestion
typedef int hiop::int_type;
```

src/Drivers/nlpSparse_ex7_driver.cpp
```diff
@@ -127,7 +127,7 @@ static bool self_check(long long n, double objval)
     if(n_saved[it]==n) {
       found=true;
       if(fabs( (objval_saved[it]-objval)/(1+objval_saved[it])) > relerr) {
-	printf("selfcheck failure. Objective (%18.12e) does not agree (%d digits) with the saved value (%18.12e) for n=%lld.\n",
+	printf("selfcheck failure. Objective (%18.12e) does not agree (%d digits) with the saved value (%18.12e) for n=%d.\n",
```
pelesh, line 130: Consider using fstream istead of fprintf.

src/Drivers/nlpSparse_ex7_driver.cpp
```diff
@@ -138,7 +138,7 @@ static bool self_check(long long n, double objval)
   }
 
   if(!found) {
-    printf("selfcheck: driver does not have the objective for n=%lld saved. BTW, obj=%18.12e was obtained for this n.\n", n, objval);
+    printf("selfcheck: driver does not have the objective for n=%d saved. BTW, obj=%18.12e was obtained for this n.\n", n, objval);
```
pelesh, line 141: Ditto

src/Drivers/nlpDenseCons_ex3_driver.cpp
```diff
@@ -141,7 +141,7 @@ static bool self_check(long long n, double objval)
     if(n_saved[it]==n) {
       found=true;
       if(fabs( (objval_saved[it]-objval)/(1+objval_saved[it])) > relerr) {
-	printf("selfcheck failure. Objective (%18.12e) does not agree (%d digits) with the saved value (%18.12e) for n=%lld.\n",
+	printf("selfcheck failure. Objective (%18.12e) does not agree (%d digits) with the saved value (%18.12e) for n=%d.\n",
```
pelesh, line 144: Consider using fstream.

pelesh:
The current version of HiOp has different types for global and local vector and dense matrix indices, similar as in [Tpetra](https://github.com/trilinos/Trilinos/tree/master/packages/tpetra). This PR sets global and local indices to be of the same type. Is that intentional?

pelesh (commented):
I would really use this opportunity to make `int` <-> `long long` replacement work out of box. To do so one only needs to set:
- generic MPI int type
- replace type specific fprintf calls with fstream implementation

This requires minimal additional effort.

src/Interface/hiopInterface.h
nychiang: I think this is specialized for C interface. Can we define it in hiop_defs.hpp?

nychiang:
LGTM. My only worry is the hard coded `MPI_INT`, but we may fix it later when we need to set `hiop::int_type=long long`.
Tried to run all the examples in my Mac and all work fine.
Should we use develop as the base of this PR, and deleted PR #235 ?

nychiang (approved):

src/Interface/hiopInterface.h
cnpetra: `hiopInterface.h` is C code, cannot use C++ code like `using` or including `hiop_defs.hpp` ... a common C include file with defines of  common types between C and C++ interfaces would be ideal, but the typedef should do the job for now.

src/Utils/hiopOptions.cpp
cnpetra: fixing a bug is never out of scope :)

cnpetra:
thanks, good catches @nychiang and @pelesh

nychiang:
BTW, I see this PR changes all the `long long` to `hiop::int_type`. Do we have plan to change `int` to `hiop::int_type`?
For example, [member variable nnz_offdiag](https://github.com/LLNL/hiop/blob/da4a3684d11f5be0d7d828d5296bd596f1593f5c/src/LinAlg/hiopMatrixSparseTriplet.hpp#L364) is returned by function [numberOfOffDiagNonzeros](https://github.com/LLNL/hiop/blob/da4a3684d11f5be0d7d828d5296bd596f1593f5c/src/LinAlg/hiopMatrixSparseTriplet.hpp#L349).

nychiang:
@cnpetra I saw you changed the base branch couple of times and merged branch `develop` into `indextype-change-dev`, after you merge branch`issue-232-233-fix` into `develop`.
Since you created branch `indextype-change-dev` from `issue-232-233-fix`, I think the easiest way to make all the code correct is to rebase `indextype-change-dev` onto `develop` through `issue-232-233-fix`, i.e., using command `git rebase --onto develop issue-232-233-fix indextype-change-dev`. 
Then you can have a fast forward and wont see any conflicts. In addition, all the commits from `issue-232-233-fix` will be removed from this PR, and the commit record of this PR will start from `replaced 'long long' with generic hiop::int_type 4c7d4d8`.

cnpetra:
> @cnpetra I saw you changed the base branch couple of times and merged branch `develop` into `indextype-change-dev`, after you merge branch`issue-232-233-fix` into `develop`.
> Since you created branch `indextype-change-dev` from `issue-232-233-fix`, I think the easiest way to make all the code correct is to rebase `indextype-change-dev` onto `develop` through `issue-232-233-fix`, i.e., using command `git rebase --onto develop issue-232-233-fix indextype-change-dev`.
> Then you can have a fast forward and wont see any conflicts. In addition, all the commits from `issue-232-233-fix` will be removed from this PR, and the commit record of this PR will start from `replaced 'long long' with generic hiop::int_type 4c7d4d8`.


> @cnpetra I saw you changed the base branch couple of times and merged branch `develop` into `indextype-change-dev`, after you merge branch`issue-232-233-fix` into `develop`.
> Since you created branch `indextype-change-dev` from `issue-232-233-fix`, I think the easiest way to make all the code correct is to rebase `indextype-change-dev` onto `develop` through `issue-232-233-fix`, i.e., using command `git rebase --onto develop issue-232-233-fix indextype-change-dev`.
> Then you can have a fast forward and wont see any conflicts. In addition, all the commits from `issue-232-233-fix` will be removed from this PR, and the commit record of this PR will start from `replaced 'long long' with generic hiop::int_type 4c7d4d8`.

yes, should have rebased... just squashed merged, guess it's the same end result

PR #246, opened 2021-06-01
Fix buildsystem and RAJA policies regarding HIP

ashermancinelli:
Updates to build system to enable building recent develop branch on HIP platform.


ashermancinelli:
Ping @pelesh @CameronRutherford

cmake/FindHiopHipLibraries.cmake
```diff
@@ -0,0 +1,52 @@
...
+  ${HIPBLAS_LIBRARY_DIR}/..
+  PATH_SUFFIXES
+  include)
+
+if(HIPBLAS_LIBRARY)
+  add_library(Hipblas INTERFACE)
+  target_link_libraries(Hipblas INTERFACE ${HIPBLAS_LIBRARY})
+  target_include_directories(Hipblas INTERFACE ${HIPBLAS_INCLUDE_DIR})
+  message(STATUS "Found Hipblas include: ${HIPBLAS_INCLUDE_DIR}")
+  message(STATUS "Found Hipblas library: ${HIPBLAS_LIBRARY}")
```
CameronRutherford, line 34: Since this is a dependency that is not a part of an export set, we should (similar to COINHSL [here](https://github.com/LLNL/hiop/commit/8916a52daae1cb73d011089a66516682fb2f0e4d#diff-ea5d33bbfc6af651063ada4999653ea00a02dc47c347e5304356b6af6ed00281R44)) add:
```CMake
install(TARGETS Hipblas EXPORT hiop-targets)
```
Another option is to add logic in `cmake/HiOpConfig.cmake.in`, however for Hipblas specifically what I suggested seems best.

cmake/FindHiopHipLibraries.cmake
```diff
@@ -0,0 +1,52 @@
...
+else()
+  message(STATUS "Hipblas was not found.")
+endif()
+
+# for now we rely on MAGMA for GPUs computations
+include(FindHiopMagma)
+target_link_libraries(Magma    INTERFACE Hipblas)
+
+# Find hip cmake targets
+find_package(hip REQUIRED)
```
CameronRutherford, line 44: This is where we should add something in `cmake/HiOpConfig.cmake.in`. Thankfully it is simple in this case. A consuming project will error when linking against HiOp here with `hip::device Magma` not found. Now, if the project has `find_package(hip REQUIRED)` somewhere in their CMake then this is no issue.

In [cmake/HiopConfig.cmake.in](https://github.com/LLNL/hiop/blob/develop/cmake/HiOpConfig.cmake.in#L28), currently there is logic for exporting the CUDA config that depends on `HIOP_USE_GPU`, and this should be modified to be similar to main CMakeLists.txt in terms of logic + any additional configurations:

```CMake
if(@HIOP_USE_GPU@) 
  if(@HIOP_USE_CUDA@)
    # Existing CUDA config
  endif()
  if(@HIOP_USE_HIP@)
    find_package(hip REQUIRED)
    add_definitions(-DHAVE_HIP) # For hipmagma
  endif()
endif()
```

src/Drivers/nlpMDS_raja_ex4.cpp
```diff
@@ -6,18 +6,9 @@
...
-  using ex4_raja_exec   = RAJA::cuda_exec<RAJA_CUDA_BLOCK_SIZE>;
-  using ex4_raja_reduce = RAJA::cuda_reduce;
-  #define RAJA_LAMBDA [=] __device__
-#else
-  using ex4_raja_exec   = RAJA::omp_parallel_for_exec;
-  using ex4_raja_reduce = RAJA::omp_reduce;
-  #define RAJA_LAMBDA [=]
-#endif
-
+#include <hiop_raja_defs.hpp>
```
CameronRutherford, line 9: Since we include this, why not just replace `ex4_raja_exec` with `hiop::hiop_raja_exec` for the moment?

CameronRutherford (changes requested):
Apart from those comments, LGTM :)

src/Drivers/nlpMDS_raja_ex4.cpp
```diff
@@ -199,12 +190,10 @@ bool Ex4::get_vars_info(const long long& n, double *xlow, double* xupp, Nonlinea
       }
     });
 
-  /// Using OpenMP execution policy for nonlinearity type setting
-  RAJA::forall<RAJA::omp_parallel_for_exec>(RAJA::RangeSegment(0, n),
-    [=](RAJA::Index_type i)
-    {
-      type[i] = hiopNonlinear;
-    });
+  for (int i = 0; i<n; ++i)
```
ashermancinelli, line 193: I will replace this with a seq host policy to keep it in a raja loop.

cmake/FindHiopHipLibraries.cmake
```diff
@@ -0,0 +1,54 @@
...
+  target_include_directories(Hipblas INTERFACE ${HIPBLAS_INCLUDE_DIR})
+  message(STATUS "Found Hipblas include: ${HIPBLAS_INCLUDE_DIR}")
+  message(STATUS "Found Hipblas library: ${HIPBLAS_LIBRARY}")
+else()
+  message(STATUS "Hipblas was not found.")
+endif()
+
+# for now we rely on MAGMA for GPUs computations
+include(FindHiopMagma)
+target_link_libraries(Magma    INTERFACE Hipblas)
```
CameronRutherford, line 41: This line is causing an issue with exportingn as Hipblas is linked against, but not a part of an export set. Adding:

```CMake
install(TARGETS Hipblas EXPORT hiop-targets)
```

should fix :)

src/Drivers/nlpMDS_raja_ex4.cpp
CameronRutherford: Fair enough. I figured there was a reason, but I wanted to ask the question :)

CameronRutherford (approved):

ashermancinelli:
@cnpetra We used this branch for our Spock development session and I've tested it there quite thoroughly, I believe this is ready for review/merge.

cnpetra (approved):
I like that you introduced the raja defs header... that should keep us moving forward in the short term with minimal to no changes.

pelesh:
When trying to configure with CUDA support, I gettwo variables that seem to be setting the same config option: `CMAKE_CUDA_ARCHITECTURES` and `HIOP_NVCC_ARCH`. Do we need them both? I get this with CMake 3.19.

CameronRutherford:
@pelesh with CMake 3.18 just directly modifying `CMAKE_CUDA_ARCHITECTURES` is sufficient. I support removing `HIOP_NVCC_ARCH` completely in favour of this (we have already done so in ExaGO).

src/Drivers/nlpMDS_raja_ex4.cpp
```diff
@@ -6,17 +6,9 @@
...
-  #define RAJA_CUDA_BLOCK_SIZE 128
-  using ex4_raja_exec   = RAJA::cuda_exec<RAJA_CUDA_BLOCK_SIZE>;
-  using ex4_raja_reduce = RAJA::cuda_reduce;
-  #define RAJA_LAMBDA [=] __device__
-#else
-  using ex4_raja_exec   = RAJA::omp_parallel_for_exec;
-  using ex4_raja_reduce = RAJA::omp_reduce;
-  #define RAJA_LAMBDA [=]
-#endif
+#include <hiop_raja_defs.hpp>
```
pelesh, line 9: I don't think we should include `hiop_raja_defs.hpp` in the example. I suggest we keep this header out of HiOp API altogether. There is no reason why a user would want to use same policies as in HiOp library. User should be advised to use _compatible_ policies, but not necessarily the same. For example, for certain models different CUDA block size might be needed.

src/Drivers/nlpMDS_raja_ex4.cpp
```diff
@@ -198,9 +190,9 @@ bool Ex4::get_vars_info(const size_type& n, double *xlow, double* xupp, Nonlinea
       }
     });
 
-  /// Using OpenMP execution policy for nonlinearity type setting
-  RAJA::forall<RAJA::omp_parallel_for_exec>(RAJA::RangeSegment(0, n),
-    [=](RAJA::Index_type i)
+  // Must be a sequential host policy for now
```
pelesh, line 193: Make sure people understand this is a temporary solution. 
```suggestion
  // Use a sequential policy for host computations for now
```

src/Drivers/nlpMDS_ex4.c
```diff
@@ -307,17 +307,17 @@ int main(int argc, char **argv) {
...
+  double* Q = (double*)malloc(nd*nd*sizeof(double));
   for(i=0; i<nd*nd; i=i+1) Q[i] = 1e-8;
   for(i=0; i<nd; i=i+1) Q[i + i*nd] += 2.0;
   for(i=1; i<nd-1; i=i+1) {
     Q[i+1 + i*nd] += 1.0;
     Q[i + (i+1)*nd] += 1.0;
   }
 
-  double* Md = malloc(ns*nd*sizeof(double));
+  double* Md = (double*)malloc(ns*nd*sizeof(double));
```
pelesh, line 318: Should we use space between cast expression and the `malloc` function?
```suggestion
  double* Md = (double*) malloc(ns*nd*sizeof(double));
```
If so, perhaps we should update developer guidelines accordingly?

pelesh (approved):
I suggested a few minor changes, but other than that, the PR looks good.

ashermancinelli:
@pelesh +1 to removing hiop nvcc arch in favor of cmake variable. I'll update that.

src/Drivers/nlpMDS_ex4.c
ashermancinelli: I don't really care either way, @cnpetra do you have an opinion?

cnpetra: looks better with the space, but it's up to you

PR #252, opened 2021-06-09
Fix unit test for ex8 and ex9 in cmake file

nychiang:
According to [`Driver/CMakeLists.txt`](https://github.com/LLNL/hiop/blob/1d249ca947e99e23f75db03c6f178fcc4c83c25f/src/Drivers/CMakeLists.txt#L56), Ex8 and Ex9 are compiled if MPI is used.

However, the current `CMakeLists.txt` adds unit test for Ex8 and Ex9 using different if statement ([see here](https://github.com/LLNL/hiop/blob/1d249ca947e99e23f75db03c6f178fcc4c83c25f/CMakeLists.txt#L418-L423)).

This is a quick fix, using the if conditions in `Driver/CMakeLists.txt`.


cnpetra:
yes, good catch. Originally we wanted to have serial (non-MPI) ex8 and ex9 as well. Not sure if that is still the case @jwang125 ... I am fine to have them only with MPI.

ashermancinelli (approved):

nychiang:
@jwang125 I am going to close this PR, since I am fine to have these two tests only with MPI. You may want to change the code so that they don't reply on MPI, but I think we should file a new PR for that.

jwang125:
Serial run made it easier to test on some platforms and have separate functions. I think keeping the functions but not in the test is fine.

PR #260, opened 2021-06-16
Header install hotfix

ashermancinelli:
`hiop_defs.hpp` now includes `hiop_types.h` as of commit `37ea1152`. Some installed headers include `hiop_defs.hpp`, but `hiop_types.h` is not installed, so dependents of hiop fail to build. Is this intended behavior? If so, this patch should fix it.


ashermancinelli:
Looks like `hiopInterfacePrimalDecomp.hpp` wasn't installed either, added that too

cnpetra (approved):

cnpetra:
good catch

PR #268, opened 2021-06-25
Remove multiple warnings (minor issues).

### LINKED ISSUE #267: Multiple warnings when building develop branch
Several warnings pop up when building `develop` branch (without deep checks or developer mode enabled):
```shell
(...)
/hiop/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp(330): warning: variable "nx" was declared but never referenced
/hiop/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp(339): warning: variable "M1nnz" was declared but never referenced
/hiop/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp(340): warning: variable "M2nnz" was declared but never referenced
/hiop/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp(615): warning: variable "num_rows" was declared but never referenced
/hiop/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp(775): warning: variable "nx" was declared but never referenced
/hiop/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp(1056): warning: variable "irow_c" was declared but never referenced
/hiop/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp(1058): warning: variable "irow_d" was declared but never referenced
/hiop/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp(1072): warning: variable "nnz_Jac_d_new" was declared but never referenced
(...)
/hiop/src/Drivers/nlpMDS_raja_ex4.hpp(46): warning: overloaded virtual function "hiop::hiopInterfaceBase::get_starting_point" is only partially overridden in class "Ex4"
/hiop/src/Optimization/hiopNlpFormulation.hpp(460): warning: type qualifier on return type is meaningless
/hiop/src/Optimization/hiopNlpFormulation.hpp(461): warning: type qualifier on return type is meaningless
/hiop/src/Optimization/hiopNlpFormulation.hpp(462): warning: type qualifier on return type is meaningless
/hiop/src/Optimization/hiopNlpFormulation.hpp(463): warning: type qualifier on return type is meaningless
```
These seem to be minor issues, but should probably be fixed nevertheless.

pelesh:
Minor fixes to remove compilation warnings in `develop` branch.

Fixes #267.

### DIFF
--- a/src/Drivers/nlpMDS_raja_ex4.cpp
+++ b/src/Drivers/nlpMDS_raja_ex4.cpp
@@ -639,12 +639,18 @@ bool Ex4::get_starting_point(const size_type& global_n, double* x0)
   return true;
 }
 
-bool Ex4::get_starting_point(const size_type& n, const size_type& m,
-    double* x0,
-    bool& duals_avail,
-    double* z_bndL0, double* z_bndU0,
-    double* lambda0)
+bool Ex4::get_starting_point(const size_type& n, 
+                             const size_type& m,
+                             double* x0,
+                             bool& duals_avail,
+                             double* z_bndL0,
+                             double* z_bndU0,
+                             double* lambda0,
+                             bool& slacks_avail,
+                             double* )
 {
+  slacks_avail = false;
+  
   if(sol_x_ && sol_zl_ && sol_zu_ && sol_lambda_)
   {
     duals_avail = true;
--- a/src/Drivers/nlpMDS_raja_ex4.hpp
+++ b/src/Drivers/nlpMDS_raja_ex4.hpp
@@ -158,11 +158,28 @@ public:
   /* Implementation of the primal starting point specification */
   bool get_starting_point(const size_type& global_n, double* x0);
 
-  bool get_starting_point(const size_type& n, const size_type& m,
-      double* x0,
-      bool& duals_avail,
-      double* z_bndL0, double* z_bndU0,
-      double* lambda0);
+  bool get_starting_point(const size_type& n,
+                          const size_type& m,
+                          double* x0,
+                          bool& duals_avail,
+                          double* z_bndL0,
+                          double* z_bndU0,
+                          double* lambda0,
+                          bool& slacks_avail,
+                          double* ineq_slack);
+
+  bool get_starting_point(const size_type& n,
+                          const size_type& m,
+                          double* x0,
+                          double* z_bndL0, 
+                          double* z_bndU0,
+                          double* lambda0,
+                          double* ineq_slack,
+                          double* vl0,
+                          double* vu0)
+  {
+    return false;
+  }
 
   /* The public methods below are not part of hiopInterface. They are a proxy
    * for user's (front end) code to set solutions from a previous solve. 
--- a/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
+++ b/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
@@ -213,10 +213,10 @@ void hiopMatrixRajaSparseTriplet::timesVec(double beta,
       y[i] *= beta;
     });
 
-#ifndef NDEBUG
-  auto nrs = nrows_;
-  auto ncs = ncols_;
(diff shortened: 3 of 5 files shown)


nychiang:
I think some variables are compiled only in debug mode. 
For example, [see here](https://github.com/LLNL/hiop/pull/268/commits/4e159874e129800cc5f38b543348b38a2352a58a#diff-a4293458926f3ca06d19d411a184b3c60887ecde5f1ad97b3ea6a94008709abaR370). 
If we remove the definition of `M1nnz`, we also need to correct the corresponding lines.

pelesh:
> I think some variables are compiled only in debug mode.
> For example, [see here](https://github.com/LLNL/hiop/pull/268/commits/4e159874e129800cc5f38b543348b38a2352a58a#diff-a4293458926f3ca06d19d411a184b3c60887ecde5f1ad97b3ea6a94008709abaR370).
> If we remove the definition of `M1nnz`, we also need to correct the corresponding lines.

So I noticed :). I believe I sorted it out.

pelesh:
For some reason mirroring to newell/marianas pipelines failed. Tests were not run there. Ascent pipeline passes.

nychiang (approved):
lgtm. Some fixes are out of the scope but good to have, e.g., add space behind and in the front of "=".
I believe we have the same issues in other places. Should we use @ashermancinelli's script to fix all these kinds of issues? We can open another PR for that.

pelesh:
Originally, @juandiegomontoya used preprocessor directives to avoid defining debug variables and thus eliminate compiler warnings in the release build:
```c++
#ifndef NDEBUG
  int num_rows = nrows_;
  int num_cols = ncols_;
#endif
```
I used `__attribute__` instead, but apparently this is not quite portable:
```c++
  // num_rows and num_columns are used in assert statements only
  int num_rows __attribute__ ((unused)) = nrows_;
  int num_cols __attribute__ ((unused)) = ncols_;
```
Usual go-to solution would be to do something like this:
```c++
  int num_rows = nrows_;
  (void) num_rows;
```
It seems that this does not fool the CUDA compiler and it returns an unused variable warning nevertheless. I am not sure why this happens, because this code should be compiled by the host compiler anyway. I believe using preprocessor guards are our best path forward, especially since there are only a few instances where these issues appear.

CC @ashermancinelli @CameronRutherford

JuanDiegoMontoya:
I did some research on this, and it looks like there is no easy solution that works with all compilers for all versions of C++. The closest thing I could find is this, which supports all major compilers and MSVC for C++17 and higher:
```cpp
#if defined(__GNUC__) || defined(__CUDACC__)
#define UNUSED __attribute__((unused))
#elif (defined(_MSVC_LANG) && _MSVC_LANG >= 201703L) || __cplusplus >= 201703L
#define UNUSED [[maybe_unused]]
#else
#define UNUSED
#endif

void foo()
{
    int UNUSED bar = 1;
    assert(bar);
}
```

A possibly portable solution is to use a template function, though I haven't tested this with NVCC:
```cpp
template<typename T>
void UNUSED(T&&) { }

void foo()
{
    int bar = 1;
    UNUSED(bar);
    assert(bar);
}
```

pelesh:
I think for our needs using preprocessor directives checking if `NDEBUG` is defined is probably best path forward. There are only 6 places in the code where there are variables defined that are used only within `assert`s, so we can change this at a later time without too much stress.

pelesh:
> For some reason mirroring to newell/marianas pipelines failed. Tests were not run there. Ascent pipeline passes.

BTW, we fixed the issue with CI, so tests pass on marianas and newell pipelines.

cnpetra:
> I think for our needs using preprocessor directives checking if `NDEBUG` is defined is probably best path forward. There are only 6 places in the code where there are variables defined that are used only within `assert`s, so we can change this at a later time without too much stress.

yes, ifdefs should be used for now.

cnpetra (approved):

Issue #266, opened 2021-06-24
Disable Build Matrix


ashermancinelli:
The build script is extremely fragile concerning the build matrix. This should be removed for now. Additionally, reproducing the configuration used in CI should be documented.

ashermancinelli:
There are a few corners of the build system and CI config that are pretty unseemly. This issue is to track their resolution.

cnpetra:
seems that this has been addressed

PR #269, opened 2021-06-25
Update dev workflow

ashermancinelli:
I disabled the build matrix, as we now have better ways to test multiple configurations and the configuration was woefully out of date.

After reviewing the documentation for Summit and developers, I updated the docs to reflect recent changes to HiOp and our build system. The development workflow on supported CI platforms is documented as well. I tested the summitVariables.sh and was able to build everything there. Looking for feedback on the documentation, as I may have removed some docs that others want to keep.

Closes: #266


README_developers.md
```diff
@@ -18,6 +18,48 @@ $ git checkout -b my-bug-fix
...
+```console
+$ cd /path/to/hiop
+$ export MY_CLUSTER=newell # or marianas or ascent
+$ source ./scripts/${MY_CLUSTER}Variables.sh
+$ mkdir build && cd build
+$ cmake ..
+$ ccmake . # Optional, if you would like to further configure the build
+$ make -j 16
+$ make test
+```
```
pelesh, line 36: Perhaps leaving `ccmake` outside the core instructions would better emphasize it is optional?
```suggestion
$ make -j 16
$ make test

Optionally, you may want to modify default CMake configuration with
`ccmake` or `cmake-gui` tools.
```

pelesh (approved):
This looks good. I like how much shorter is the documentation now.

ashermancinelli:
@cnpetra Could we sneak this rather small PR in before the other two on the docket?

jwang125:
Hi guys, just tested this on summit. Here's my experience. I tried MY_CLUSTER=summit ./BUILD.sh and it compiles great no issue. I noticed make test was run as well and the MPI tests failed, so I thought HIOP_TEST_WITH_BSUB needs to be on. I went into /build and cmake can't find mpi compiler. I did source summitVariables.sh it compiles great again. I 'bsub'ed make test and this time 80% tests failed. I'm not sure if this is intended or perhaps I did things wrong.

nychiang:
You can run unit test from folder `tests`. For example, `jsrun -n 1 -g 1 build/tests/testVector`.
This can bring you more details about the failure.
Sounds to me your nvblas.conf is not loaded correctly and hence 80% of the tests fail.

jwang125:
nvblas is in the the correct folder. Here's the output of Vectortest. Seems like mpi compiler or library issue. I've not seen this error before. And this error didn't appear when using ./BUILD.sh. But it has different tests failing. 

[NVBLAS] NVBLAS_CONFIG_FILE environment variable is set to '/ccs/home/wang125/software/hiop/nvblas.conf'
[NVBLAS] Config parsed
--------------------------------------------------------------------------
Sorry!  You were supposed to get help about:
    opal_init:startup:internal-failure
But I couldn't open the help file:
    /opt/ibm/spectrum_mpi/jsm_pmix/..//container/../share/openmpi/help-opal-runtime.txt: No such file or directory.  Sorry!
--------------------------------------------------------------------------
--------------------------------------------------------------------------
Sorry!  You were supposed to get help about:
    orte_init:startup:internal-failure
But I couldn't open the help file:
    /opt/ibm/spectrum_mpi/jsm_pmix/..//container/../share/openmpi/help-orte-runtime: No such file or directory.  Sorry!
--------------------------------------------------------------------------
--------------------------------------------------------------------------
Sorry!  You were supposed to get help about:
    mpi_init:startup:internal-failure
But I couldn't open the help file:
    /opt/ibm/spectrum_mpi/jsm_pmix/..//container/../share/openmpi/help-mpi-runtime.txt: No such file or directory.  Sorry!
--------------------------------------------------------------------------
*** An error occurred in MPI_Init
*** on a NULL communicator
*** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
***    and potentially your MPI job)
[f12n15:45225] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!

cnpetra:
> @cnpetra Could we sneak this rather small PR in before the other two on the docket?

yes, this has highest priority and will merge as soon as we sort out issues reported by @jwang125...

cnpetra:
when doing the spack-based build, isn't the user supposed to download coinhsl on his own (and let spack know where the source files are)? Or spack takes care of it?

nychiang:
@cnpetra It seems that on the predefined platform, i.e., asent, summit, Mariana and newell, we can use exasgd-coinhsl/2015.06.23, which is installed by spack.
@ashermancinelli Can we use spack view, instead of loading modules separately? I remember you suggest doing that way.

ashermancinelli:
@cnpetra you're right, when installing via spack you'd need coinhsl in your cwd, but I've already installed all the dependencies on all 4 platforms. 

I built and ran all tests and they all passed when I was in a bsub allocation, so I'm not sure what's going on with the test failures. I can look on Monday though.

nychiang:
Like ascentVariables.sh, should we add line
`EXTRA_CMAKE_ARGS="$EXTRA_CMAKE_ARGS -DCMAKE_CUDA_ARCHITECTURES=70 -DHIOP_TEST_WITH_BSUB=ON"`
to summitVariables.sh?

ashermancinelli:
> @cnpetra It seems that on the predefined platform, i.e., asent, summit, Mariana and newell, we can use exasgd-coinhsl/2015.06.23, which is installed by spack.
> 
> @ashermancinelli Can we use spack view, instead of loading modules separately? I remember you suggest doing that way.

Yes, if you want to install via spack yourself a view is great. I was concerned with the difficulty some of our team ran into with using spack directly, so I had spack generate these modulefiles. I think modulefiles will be the best solution going forward. It will allow you to swap in another version of a dependency without requiring an entirely new spack environment, and it's extremely clear what versions of each dependency are being used.

ashermancinelli:
> nvblas is in the the correct folder. Here's the output of Vectortest. Seems like mpi compiler or library issue. I've not seen this error before. And this error didn't appear when using ./BUILD.sh. But it has different tests failing. 
> 
> 
> 
> [NVBLAS] NVBLAS_CONFIG_FILE environment variable is set to  '/ccs/home/wang125/software/hiop/nvblas.conf'
> 
> [NVBLAS] Config parsed
> 
> 
> 
> Sorry!  You were supposed to get help about:
> 
>     opal_init:startup:internal-failure
> 
> But I couldn't open the help file:
> 
>     /opt/ibm/spectrum_mpi/jsm_pmix/..//container/../share/openmpi/help-opal-runtime.txt: No such file or directory.  Sorry!
> 
> 
> 
> Sorry!  You were supposed to get help about:
> 
>     orte_init:startup:internal-failure
> 
> But I couldn't open the help file:
> 
>     /opt/ibm/spectrum_mpi/jsm_pmix/..//container/../share/openmpi/help-orte-runtime: No such file or directory.  Sorry!
> 
> 
> 
> Sorry!  You were supposed to get help about:
> 
>     mpi_init:startup:internal-failure
> 
> But I couldn't open the help file:
> 
>     /opt/ibm/spectrum_mpi/jsm_pmix/..//container/../share/openmpi/help-mpi-runtime.txt: No such file or directory.  Sorry!
> 
> 
> 
> *** An error occurred in MPI_Init
> 
> *** on a NULL communicator
> 
> *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
> 
> ***    and potentially your MPI job)
> 
> [f12n15:45225] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!

Could you send me the exact commands you used to produce these failures? I can look into it tomorrow. This may be an issue with my documentation not being thorough enough, so I apologize for that.

jwang125:
Yes. And I have a better idea now what was wrong.
First, here's what I did. First with bsub test on:
>git clone hiop, checkout update-dev-workflow
source ./scripts/summitVariables.sh
mkdir build && cd build
cmake -DHIOP_TEST_WITH_BSUB=ON ..
make -j 16     <- successful
bsub -Pcsc359 -W10 -nnodes1 
> make test
>crtl+D

and wait for the results. 80%+ failed.
Second without bsub test on:
>git clone hiop, checkout update-dev-workflow
source ./scripts/summitVariables.sh
mkdir build && cd build
 cmake ..
make -j 16     <- successful
make test

Only 20% failed! Only the mpi tests failed.
So I went into the build/test folder. I think this is because somehow jsrun is not working with this build. when TEST_WITH_BSUB is on mpirun is replaced with jsrun and that's why so many more tests failed. The error I printed in the previous post is from jsrun. If TEST_WITH_BSUB is off, mpirun -n 1 ./testVector works. However, mpirun -n 2 ./testVector does not. And the error is:

bash-4.2$ mpirun -n 2 testVector 
stty: standard input: Inappropriate ioctl for device
testVector: /lib64/libstdc++.so.6: version `CXXABI_1.3.8' not found (required by testVector)
testVector: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.21' not found (required by testVector)
testVector: /lib64/libstdc++.so.6: version `CXXABI_1.3.8' not found (required by /autofs/nccs-svm1_sw/summit/.swci/1-compute/opt/spack/20180914/linux-rhel7-ppc64le/gcc-7.4.0/openmpi-4.0.3-hf3xf5weredkclm4lkxadtizibv7z6i6/lib/libmpi_cxx.so.40)
testVector: /lib64/libstdc++.so.6: version `CXXABI_1.3.9' not found (required by /autofs/nccs-svm1_sw/summit/.swci/1-compute/opt/spack/20180914/linux-rhel7-ppc64le/gcc-7.4.0/openmpi-4.0.3-hf3xf5weredkclm4lkxadtizibv7z6i6/lib/libmpi_cxx.so.40)
Primary job terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
mpirun detected that one or more processes exited with non-zero status, thus causing
the job to be terminated. The first process to do so was:
  Process name: [[43267,1],1]
  Exit code:    1

To sum up, two issues. One is that jsrun doesn't work (it works on develop) and the other is mpirun not working. The error seems to say standard library is not available? I'm not sure.

cnpetra (approved):
lgtm minus the trouble @jwang125 has, would like to troubleshoot that before merging.

ashermancinelli:
I'm skeptical that the GLIBCXX issue is with this setup, looks like you don't have the GCC module loaded as it's picking up the system libstdc++.

ashermancinelli:
I'll have to investigate what's going on, but to get a known working configuration, you can use the following:

```console
$ source scripts/summitVariables.sh
$ mkdir build && cd build
$ cmake -C ../scripts/defaultCIBuild.cmake ..
$ make -j 16
$ make test
```

as described in the documentation I wrote for this branch. The tests should pass by default however.

pelesh:
> I'm skeptical that the GLIBCXX issue is with this setup, looks like you don't have the GCC module loaded as it's picking up the system libstdc++.

I agree. @jwang125, you can also look into `CMakeCache.txt` file to see what compilers CMake has set for you. Search e.g. for variables `CMAKE_CXX_COMPILER` and `MPI_C_COMPILER`. You can also view these in `ccmake` or `cmake-gui` tools when using advanced mode.

jwang125:
I tried what Asher suggested, and again the MPI tests failed like previously.
$ source scripts/summitVariables.sh
$ mkdir build && cd build
$ cmake -C ../scripts/defaultCIBuild.cmake ..
$ make -j 16
$ make test
I then tried in tests 
$ mpirun -n 2 ./testVector
and get the error ./testVector: error while loading shared libraries: libnvblas.so.10: cannot open shared object file: No such file or directory. I added the library libnvblas.so.10 to LD_LIBRARY_PATH and it doesn't work. bsub tests return the same result. nvblas.conf is automatically generated by cmake.
I then tried
$jsrun -n2 -a1 -g1 -c1 ./testVecotr
and it doesn't work and I don't think it's supposed to work. 
To sum it up, I can't get MPI tests to run. I tried my own set of parameters which uses cuda/11.0.3 and it works. I've tried deleting and reclone. I would have time if you want a quick 10 minute meeting so that I don't hold up the merge.

scripts/defaultBuild.sh
```diff
@@ -33,8 +11,9 @@ defaultBuild()
     echo
     echo Build step
     echo
-    pushd $BUILDDIR                             || exit 1
-    cmake $CMAKE_OPTIONS ..                     || exit 1
+    local SRCDIR=$PWD
+    pushd $BUILDDIR || exit 1
+    cmake -C $SRCDIR/scripts/defaultCIBuild.cmake $EXTRA_CMAKE_ARGS .. || exit 1
```
CameronRutherford, line 16: `$EXTRA_CMAKE_ARGS` should come before the cache cmake. This will allow default CUDA architecture configuration, with it being customizable through cluster scripts.

scripts/summitVariables.sh
```diff
@@ -0,0 +1,75 @@
...
+NVBLAS_GPU_LIST ALL
+NVBLAS_TILE_DIM 2048
+NVBLAS_AUTOPIN_MEM_ENABLED
+EOD
+export NVBLAS_CONFIG_FILE=$PWD/nvblas.conf
+echo "Generated $PWD/nvblas.conf"
+
+export EXTRA_CMAKE_ARGS="$EXTRA_CMAKE_ARGS \
+  -DCMAKE_CUDA_ARCHITECTURES=70 \
+  -DPETSC_DIR=$PETSC_DIR"
```
CameronRutherford, line 72: Don't need to specify PETSC here.

CameronRutherford (changes requested):

scripts/defaultCIBuild.cmake
```diff
@@ -0,0 +1,30 @@
...
+set(HIOP_USE_CUDA ON CACHE BOOL "")
+
+message(STATUS "Enabling HiOp's Sparse Interface")
+set(HIOP_SPARSE ON CACHE BOOL "")
+
+message(STATUS "Enabling HiOp's deepchecking")
+set(HIOP_DEEPCHECKS ON CACHE BOOL "")
+
+message(STATUS "Setting default cuda architecture to 60")
+set(CMAKE_CUDA_ARCHITECTURES 60 CACHE STRING "")
```
CameronRutherford, line 28: Meant to leave this on the review, but something like this should work:
```CMake
if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
  message(STATUS "Setting default cuda architecture to 70")
  set(CMAKE_CUDA_ARCHITECTURES 70 CACHE STRING "")
else()
  message(STATUS "Using cuda architecture ${CMAKE_CUDA_ARCHITECTURES} from cache")
endif()
```

jwang125:
I made the changes Cameron suggested, same results with mpi tests failing. But now that Cameron mentions it, I have this warning for basically all tests:

[100%] Linking CXX executable testVector
/usr/bin/ld: warning: libcudart.so.10.1, needed by /autofs/nccs-svm1_sw/summit/.swci/1-compute/opt/spack/20180914/linux-rhel7-ppc64le/gcc-7.4.0/openmpi-4.0.3-hf3xf5weredkclm4lkxadtizibv7z6i6/lib/libmpi_cxx.so, may conflict with libcudart.so.10.2
[100%] Built target nlpMDS_ex4_raja.exe
/usr/bin/ld: warning: libcudart.so.10.1, needed by /autofs/nccs-svm1_sw/summit/.swci/1-compute/opt/spack/20180914/linux-rhel7-ppc64le/gcc-7.4.0/openmpi-4.0.3-hf3xf5weredkclm4lkxadtizibv7z6i6/lib/libmpi_cxx.so, may conflict with libcudart.so.10.2

Could it be openmip-4.0.3 and cuda/10.2.89 are have some compatibility issues hence the can't find library error? I have to load cuda/10.2.89 based on everything else in the summitVariables.sh.

scripts/defaultBuild.sh
ashermancinelli: I don't know about that - the extra cmake args come after the initial cache script *so that* they override whatever is set in that cmake cache script

scripts/defaultCIBuild.cmake
ashermancinelli: Refer to my comment above - just adding `-DCMAKE_CUDA_ARCHITECTURES=XX` will override whatever I have in this cache script

scripts/defaultBuild.sh
CameronRutherford: The argument here was more for clarity I suppose. When building with the current CMake cache, CMake shows that you are using the default architecture 70, and then under the hood re-sets it to whatever you have overridden it to. If this behaviour is OK, then I don't have a problem, but I liked having the arch information printed with the cache build when running with my suggestion.

ashermancinelli:
@jwang125 I tested this out again myself, and you're right, I can't get the MPI- and GPU-enabled tests to run either. It seems that there is some issue with jsrun interacting with the mpirun commands, probably with some PMI fabric setting. I'll email Phil Roth about this.

ashermancinelli:
The issues on summit seem to be valid and we're working with ORNL team on them - is it sufficient to document these issues and merge this so we have the build system improvements in develop @cnpetra? We can just document the summit issues and move from there since the summit environment is only available to ExaSGD HiOp developers anyways.

ashermancinelli:
I've documented this in #272 and engaged ORNL team, so imo I think this is ready.

jwang125:
Thank you Asher!

PR #273, opened 2021-07-02
Fix some MPI and host/device memory issues in MDS FR problem

nychiang:
Fix some bugs happened when MDS FR is used on device.
Now MDS FR problem uses the same `compute_mode` and `mem_space` as the base problem.
In addition, change the MPI `comm` in FR problem to the one used by base problem.

This PR solves the `copy_to` problem found in some exago+pridecomp tests.


nychiang:
BTW, I noticed that function [hiopNlpFormulation::finalizeInitialization](https://github.com/LLNL/hiop/blob/4e3faa9963e83cd7f2a388bfcc1e402dc86fc81d/src/Optimization/hiopNlpFormulation.cpp#L168) uses host memory to set up the problem. Therefore the interface function [interface_base.get_vars_info](https://github.com/LLNL/hiop/blob/4e3faa9963e83cd7f2a388bfcc1e402dc86fc81d/src/Optimization/hiopNlpFormulation.cpp#L220) only uses host memory.

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -1947,9 +1947,16 @@ void hiopVectorRajaPar::set_array_from_to(hiopInterfaceBase::NonlinearityType* a
   if(end - start == 0)
     return;
   
-  auto& rm = umpire::ResourceManager::getInstance();
   hiopInterfaceBase::NonlinearityType* vv = const_cast<hiopInterfaceBase::NonlinearityType*>(arr_src); // <- cast away const
```
cnpetra, line 1950: I think the cast is not needed anymore

jwang125 (approved):
Work as describe.

src/LinAlg/hiopVectorRajaPar.cpp
nychiang: Done!

cnpetra (approved):

cnpetra:
reopening

PR #278, opened 2021-07-09
Add support to PARDISO solver in CPU mode

nychiang:
Support using PARDISO to solve the symmetric indefinite sparse linear system in HiOp.

Right now it passes both nlpSparseEx6 and Ex7.


cnpetra:
that was quick!  👍

nychiang:
@cnpetra per our offline discussion, the interface for unsymmetric matrix is added.

src/Interface/hiopVersion.hpp
```diff
@@ -61,6 +61,13 @@ struct hiopVersion
       false;
 #endif
 
+  static constexpr bool usePARDISO =
+#ifdef HIOP_USE_PARDISO
+      true;
+#else
+      false;
+#endif
+
```
pelesh, line 70: I would really configure that as a HiOp option.

pelesh:
As a general comment -- after @cnpetra defined `index_type` and `size_type` as HiOp's type aliases for matrix/vector indices, it seems developers are not using them consistently. Since this is new addition to HiOp code, I would strongly recommend using agreed upon type aliases before this PR is merged.

cnpetra:
> As a general comment -- after @cnpetra defined `index_type` and `size_type` as HiOp's type aliases for matrix/vector indices, it seems developers are not using them consistently. Since this is a new addition to HiOp code, I would strongly recommend using agreed upon type aliases before this PR is merged.

agree, but the PARDISO class can still use internally `int`s or whatever PARDISO API is using.

nychiang:
I agree.
However, it seems more works need to be done if we want to use `index_type` and `size_type` to access all the vector/matrix indices. Some member functions/variables of vec/mat are still using `int` internally.
For example,  in class `hiopMatrixSparseTriplet`, we define 
```
int* iRow_; ///< row indices of the nonzero entries
inline int* i_row() { return iRow_; }
```
Another PR?

cnpetra:
> I agree.
> However, it seems more works need to be done if we want to use `index_type` and `size_type` to access all the vector/matrix indices. Some member functions/variables of vec/mat are still using `int` internally.
> For example, in class `hiopMatrixSparseTriplet`, we define
> 
> ```
> int* iRow_; ///< row indices of the nonzero entries
> inline int* i_row() { return iRow_; }
> ```
> 
> Another PR?

you're right, these changes need to be done in many other files, starting with the sparse matrix interface and propagated down the child classes. Definitely should be done in a separate PR.

src/LinAlg/hiopLinSolverSparsePARDISO.cpp
```diff
@@ -0,0 +1,523 @@
...
+    double* drhs = rhs_->local_data();
+
+    int phase = 33;
+    int nrhs = 1;
+  
+    pardiso_d(pt_ , &maxfct_, &mnum_, &mtype_, &phase,
+              &n_, kVal_, kRowPtr_, jCol_,
+              NULL, &nrhs, 
+              iparm_, &msglvl_,
+
```
cnpetra, line 328: is this an (unnecessary) extra line?

src/LinAlg/hiopLinSolverSparsePARDISO.cpp
```diff
@@ -0,0 +1,523 @@
...
+      std::cout << "PardisoSolver ERROR during pardisoinit:" << error_ << "." << std::endl;
+      assert(false);
+    }
+
+    /* Numbers of processors, value of OMP_NUM_THREADS */
+    char *var = getenv("OMP_NUM_THREADS");
+    if(var != NULL) {
+      sscanf( var, "%d", &num_threads_ );
+    } else {
+      printf("Set environment OMP_NUM_THREADS before using PARDISO.");
```
cnpetra, line 415: wouldn't be better to just let the solver work with 1 thread (assuming it does not give an error if the `OMP_NUM_THREADS` is not set)?

src/LinAlg/hiopLinSolverSparsePARDISO.cpp
```diff
@@ -0,0 +1,523 @@
...
+    pardiso_d(pt_ , &maxfct_, &mnum_, &mtype_, &phase,
+              &n_, kVal_, kRowPtr_, jCol_,
+              NULL, &nrhs, 
+              iparm_, &msglvl_,
+              
+              drhs, dx, &error_, dparm_);
+
+    if ( error_ != 0) {
+      printf ("PardisoSolver - ERROR during backsolve: %d\n", error_ );
+      assert(false);
```
cnpetra, line 516: return false

src/LinAlg/hiopLinSolverSparsePARDISO.cpp
```diff
@@ -0,0 +1,523 @@
...
+    }
+
+    nlp_->runStats.linsolv.tmTriuSolves.stop();
+    return 1;
```
cnpetra, line 520: return true

src/Optimization/hiopKKTLinSysSparse.cpp
```diff
@@ -251,12 +254,34 @@ namespace hiop
       if(nlp_->options->GetString("compute_mode")=="cpu")
       {
         nlp_->log->printf(hovScalars,
-                          "KKT_SPARSE_XYcYd linsys: alloc MA57 size %d (%d cons)\n",
+                          "KKT_SPARSE_XYcYd linsys: alloc sparse solver with matrix size %d (%d cons)\n",
```
cnpetra, line 257: it's a good idea to report which sparse linear solver is being allocated/used.

cnpetra (changes requested):
looks good, I left  few small comments

src/LinAlg/hiopLinSolverSparsePARDISO.cpp
nychiang: which line? 328 or 329? seems to me both are required.

cnpetra: 328 seems to be an empty line

nychiang: got it!

nychiang:
@cnpetra @pelesh all the comments have been addressed, except for using hiop option in "hiopVersion.hpp".
Right now "hiopVersion.hpp" is only used to provide information under xSDK R5 policy. (@ashermancinelli can confirm if this is the case)
I think we should keep its existing format, rather than re-engineering the files and using a hiopOption object in this head file.

cnpetra (approved):

cnpetra:
We need a test for this code, see issue #298, but we can work on it after this is merged

pelesh:
> Another PR?

Perhaps good idea would be to document in github issues instances where linear algebra is using hard-wired type for vector/matrix indices. We can wait to collect multiple issues and then address them in a single PR.

I would suggest, however, to make sure all new code is using index type aliases as implemented in #242.

pelesh (changes requested):
I think once a better solution is found for enabling Pardiso (perhaps through HiOp options?), this could be merged.

I don't have access to Pardios right now, so I wasn't able to build/run the code.

cnpetra:
> I think once a better solution is found for enabling Pardiso (perhaps through HiOp options?), this could be merged.
> 
> I don't have access to Pardios right now, so I wasn't able to build/run the code.

options for using/switching to pardiso are implemented.

the `usePARDISO` variable you reported in `hiopVersion.hpp` is just for the purpose of building the version (string) info as per  xSDK R5 policy. All the others linear solvers have it.

PR #259, opened 2021-06-16
RAJA implementation for sparse matrix

nychiang:
This PR is ready for review.
Note that I will change the base from `fr-mds-dev` to develop once we merge #239 

In this PR, I addressed the following functions and their corresponding unit tests:
1. `hiopMatrixRajaSparseTriplet::copyRowsFrom` 
2. `hiopMatrixRajaSparseTriplet::copy_rows_block_from` 
3. `hiopMatrixRajaSparseTriplet::copySubDiagonalFrom`
4. `hiopMatrixRajaSparseTriplet::setSubDiagonalTo`
5. `hiopVectorIntRaja::copy_from`

According to the discussion with Cosmin, I will address other RAJA functions in different PRs.
Try to make each PR as short as possible, so others can review the code easily.

CLOSE #103

PR #281, opened 2021-07-15
Revert "RAJA implementation for sparse matrix"

cnpetra:
Reverts LLNL/hiop#259


cnpetra:
done via a reset and forced push

Issue #284, opened 2021-07-20
PriDeccomp Solver x and coupled x dimension issue


jwang125:
Some of the vectors in pridecomp solver are not initialized correctly in the latest develop branch. Some vectors are the coupled part of all the optimization variables, thus should have a much smaller size. This will be fixed through a new PR soon.

cnpetra:
closed by PR #285

PR #285, opened 2021-07-21
fix x vector dimension error

jwang125:
Fixing issue mention here https://github.com/LLNL/hiop/issues/284.

### DIFF
--- a/src/Interface/hiopInterfacePrimalDecomp.cpp
+++ b/src/Interface/hiopInterfacePrimalDecomp.cpp
@@ -80,7 +80,8 @@ RecourseApproxEvaluator(const int nc,
                         const hiopVector& rgrad, 
                         const hiopVector& rhess, 
                         const hiopVector& x0)
-  : nc_(nc), S_(S)
+  : nc_(nc), S_(S),
+    mem_space_("DEFAULT")
 {
   //assert(S>=nc);
   rval_ = rval;
--- a/src/Optimization/hiopAlgPrimalDecomp.cpp
+++ b/src/Optimization/hiopAlgPrimalDecomp.cpp
@@ -552,12 +552,12 @@ bool hiopAlgPrimalDecomposition::stopping_criteria(const int it, const double co
 }
   
 double hiopAlgPrimalDecomposition::
-step_size_inf(const int nc, const hiopVector& x, const hiopVector& x0)
+step_size_inf(const int nc, const int* idx, const hiopVector& x, const hiopVector& x0)
 {
   double step = -1e20;
-  hiopVector* temp = x.new_copy();
-  //temp = LinearAlgebraFactory::create_vector(x.get_local_size()); 
-  //temp->copyFrom(x);   
+  hiopVector* temp;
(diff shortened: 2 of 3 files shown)


nychiang:
I saw you add more changes to the code. For example, change `local_data_host` to `local_data` in several places, and modify the function `step_size_inf`. Will HiOp still reach FR in the recourse problem after you change these?
I don't know why the previous assertion error disappears after these changes.

nychiang (commented):
Will these changes introduce new problem to the exago small examples (9 or 118 system)?

jwang125:
step_size_inf was wrong. Wanted to change a long time ago because was measuring |x-x_0|^2 which have different lengths and thus always returns inf. I'm not sure why the error disappeared. Could be updated develop. But develop has the error you saw.

jwang125:
The smaller tests seem to suffer the same error now with create_Vector function which throws an umpire exception. Testing to see if the new added lines caused this or was a problem with develop.

jwang125:
I have confirmed that if I only implement the one line, the error in create_Vector comes up as well. And it's from hiopInterfacePrimalDecomp. This means there's more than one error with the latest develop.

grad_r =LinearAlgebraFactory::create_vector(master_prob_->get_mem_space(), nc_) ;

I could go back to this one update for this branch if that's better. Then we can try to fix this next error.

cnpetra:
@jwang125 : currently, the hiopInterfacePrimalDecomp (and its classes) and hiopAlgPrimalDecomp have memory space hardcoded as "default". I think this may be the root cause for the umpire exceptions and other nasty things you're seeing with exago. 

I am working on a PR that will allow to change the memory space for PriDec solver (and master solver). 

@nychiang suggested to try to changes the hardcoded "default" memory space to "um" in PriDec interface and Alg and see if that's working with exago.

cnpetra:
@jwang125 sorry for the bugs I introduced with alloc_clone :(

jwang125:
Yes I am looking at the code and I agree that's why. It's also because a bad code I gave Shri a while ago that createsa a new RecourseApproxEvaluator each time set_recourse_approx_evaluator is called. And he has that in the exago code. This might explain why ex9 doesn't fail because I've changed that in ex9, which does call the constructor.

jwang125:
Hi guys, I think the issue is less with "DEFAULT" but more with the constructor for 

RecourseApproxEvaluator(const int nc,
                        const int S,
                        const double& rval,
                        const hiopVector& rgrad,
                        const hiopVector& rhess,
                        const hiopVector& x0)
which didn't define mem_space_ so it was not assigned any value. Once this is fixed, I'm back to previous known errors! If you're OK with this I will maintain changes on step_size_inf, modify this constructor and push and do this PR. Then we can look at the older known bugs.

cnpetra (approved):

cnpetra:
> Yes I am looking at the code and I agree that's why. It's also because a bad code I gave Shri a while ago that creates a a new RecourseApproxEvaluator each time set_recourse_approx_evaluator is called. And he has that in the exago code. This might explain why ex9 doesn't fail because I've changed that in ex9, which doesn't call the constructor.

user is not supposed to create `RecourseApproxEvaluator `, only to evaluate it ... Ex8 is still doing this.

PR #286, opened 2021-07-21
modify rowStartInfo for hiopMatrixRajaSparseTriplet

nychiang:
This PR addresses the following issues:
1. Compute `row_start` on CPU and save it for later use.
2. Add functions to synchronize the host and device data in RowStartsInfo.
3. Remove RAJA:inclusive_scan, and build row_start on CPU when FR is activated

### DIFF
--- a/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
+++ b/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
@@ -355,8 +355,8 @@ timesMatTrans(double beta, hiopMatrix& Wmat, double alpha, const hiopMatrix& M2m
   int M2nnz = M2.nnz_;   
 #endif
 
-  int* M1_idx_start = this->row_starts_->idx_start_;
-  int* M2_idx_start = M2.row_starts_->idx_start_;
+  index_type* M1_idx_start = this->row_starts_->idx_start_;
+  index_type* M2_idx_start = M2.row_starts_->idx_start_;
 
   int* M1jCol = this->jCol_;
   int* M2jCol = M2.jCol_;
@@ -370,8 +370,8 @@ timesMatTrans(double beta, hiopMatrix& Wmat, double alpha, const hiopMatrix& M2m
       {
         // dest[i,j] = weigthed_dotprod(M1_row_i,M2_row_j)
         double acc = 0.;
-        int ki = M1_idx_start[i];
-        int kj = M2_idx_start[j];
+        index_type ki = M1_idx_start[i];
+        index_type kj = M2_idx_start[j];
         
         while(ki<M1_idx_start[i+1] && kj<M2_idx_start[j+1])
         {
@@ -497,14 +497,14 @@ void hiopMatrixRajaSparseTriplet::row_max_abs_value(hiopVector& ret_vec)
   assert(row_starts_);
 
   int num_rows = this->nrows_;
-  int* idx_start = row_starts_->idx_start_;
+  index_type* idx_start = row_starts_->idx_start_;
   double* values = values_;
 
   RAJA::forall<hiop_raja_exec>(
     RAJA::RangeSegment(0, num_rows),
     RAJA_LAMBDA(RAJA::Index_type row_id)
     {
-      for(int itnz=idx_start[row_id]; itnz<idx_start[row_id+1]; itnz++) {
+      for(index_type itnz=idx_start[row_id]; itnz<idx_start[row_id+1]; itnz++) {
         double abs_val = fabs(values[itnz]);
         vd[row_id] = (vd[row_id] > abs_val) ? vd[row_id] : abs_val;
       }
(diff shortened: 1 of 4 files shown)


cnpetra (approved):

src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
```diff
@@ -1074,17 +1099,17 @@ void hiopMatrixRajaSparseTriplet::set_Jac_FR(const hiopMatrixSparse& Jac_c,
...
 
-  if(J_d.row_starts_host == nullptr) {
-    J_d.row_starts_host = J_d.allocAndBuildRowStarts();
+  if(J_d.row_starts_ == nullptr) {
+    J_d.row_starts_ = J_d.allocAndBuildRowStarts();
   }
-  assert(J_d.row_starts_host);
-  int* Jd_row_st = J_d.row_starts_host->idx_start_;
+  assert(J_d.row_starts_);
+  int* Jd_row_st = J_d.row_starts_->idx_start_;
```
pelesh, line 1113: Should we use HiOp's `index_type` here and elsewhere in the code when accessing vector/matrix indices?

pelesh:
Why are `idx_start_` and `idx_start_host_` encapsulated in the nested class `RowStartInfo`? It seems to me it would be better to have member variables of the sparse matrix as:
```c++
  int* iRow_;       ///< row indices of the nonzero entries
  int* iRowStarts_; ///< indices pointing to starts of new rows in values_
  int* jCol_;       ///< column indices of the nonzero entries
  double* values_;  ///< values of the nonzero entries
```
That way, it seems to me, it would be easier to set row start indices instead of going through a nested class. Also, this would allow you to instantiate CSR or COO matrix as you wish. This may be beyond the scope of this PR, though.

pelesh:
The code builds smoothly and all tests pass.

However, when built in developer mode where additional compiler warning flags are added, the build produces a massive amount of warnings. Most of them seem easy to fix, but there may me some hidden in there that are pointing to potential issues. Also, CMake reported a nonfatal errors when developer mode was enabled.

CC @ashermancinelli

pelesh (changes requested):
I suggest reviewing warnings in developer mode that pertain to this PR. Since these are only 3 files to check, it should not be too difficult. Some warnings may be pointing to actual bugs.

nychiang:
> Why are `idx_start_` and `idx_start_host_` encapsulated in the nested class `RowStartInfo`? It seems to me it would be better to have member variables of the sparse matrix as:
> 
> ```c++
>   int* iRow_;       ///< row indices of the nonzero entries
>   int* iRowStarts_; ///< indices pointing to starts of new rows in values_
>   int* jCol_;       ///< column indices of the nonzero entries
>   double* values_;  ///< values of the nonzero entries
> ```
> 
> That way, it seems to me, it would be easier to set row start indices instead of going through a nested class. Also, this would allow you to instantiate CSR or COO matrix as you wish. This may be beyond the scope of this PR, though.

@pelesh Oh, I thought you introduced the nested class `RowStartInfo` and I just added the host buffer into your class. For sure I can make `irow_start_` as a member variable of the sparse matrix, but this is is a question about code design and I prefer getting an approval from the author of this code.

nychiang:
@ashermancinelli @pelesh 
I received the same cmake error in the `develop` branch:
CMake Error: install(EXPORT "hiop-targets" ...) includes target "hiop_tpl" which requires target "hiop_warnings" that is not in any export set. I think we introduced this error in PR #279. Better to file another new PR to correct it.  

About the compiler warnings, I noticed that this PR doesn't introduce new warnings. 
We have 14944 warnings even in the develop branch.
In addition, I found that 14567 warnings come from unused parameters.
I think we should file a new PR to go through all these warnings.

cnpetra:
> > Why are `idx_start_` and `idx_start_host_` encapsulated in the nested class `RowStartInfo`? It seems to me it would be better to have member variables of the sparse matrix as:
> > ```c++
> >   int* iRow_;       ///< row indices of the nonzero entries
> >   int* iRowStarts_; ///< indices pointing to starts of new rows in values_
> >   int* jCol_;       ///< column indices of the nonzero entries
> >   double* values_;  ///< values of the nonzero entries
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That way, it seems to me, it would be easier to set row start indices instead of going through a nested class. Also, this would allow you to instantiate CSR or COO matrix as you wish. This may be beyond the scope of this PR, though.
> 
> @pelesh Oh, I thought you introduced the nested class `RowStartInfo` and I just added the host buffer into your class. For sure I can make `irow_start_` as a member variable of the sparse matrix, but this is is a question about code design and I prefer getting an approval from the author of this code.

I authored the classes. The rationale for using a nested class `RowStartInfo` is that in most cases the functionality of  `RowStartInfo` is not needed and, thus,  the inner class and it's many members will not be instantiated. Not sure if this is still the case. I would let it as it is since will not help with anything at this point, performance or functionality wise.

src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
nychiang: Now I use `index_type` and `size_type` in class RowStartsInfo and the places to access these member variables.

PR #290, opened 2021-07-30
Fix the feasibility problem in MDS format

nychiang:
Fix a bug in the feasibility problem in MDS format.
Switch of gradient scaling in feasibility problem to have better numerical behaviors.


jwang125 (approved):
Works with exago problems now.

cnpetra (approved):

PR #294, opened 2021-08-06
update HiOp's Magma solvers to v2.6.1 

cnpetra:
Updates the Bunch-Kaufmann HiOp's MAGMA linear solver class to perform the inertia computation on the GPU. The dense and MDS linear algebra of HiOp runs completely in the device memory space when HiOp is used under the gpu compute mode. Under the hybrid compute node, the updated class performs the matrix and rhs transfer (with the intermediary steps being kept on the device memory space). 

Has not been tested with memory space set to gpu (only with um) but should work with minimal changes. Passes all the CI tests.


scripts/ascentVariables.sh
```diff
@@ -20,7 +20,9 @@ module load exasgd-libsigsegv/2.13/gcc-7.4.0-garv4jn
 # libtool@2.4.6%gcc@7.4.0 arch=linux-rhel7-power9le
 module load exasgd-libtool/2.4.6/gcc-7.4.0-lbasl6y
 # magma@2.5.4%gcc@7.4.0+cuda+fortran~ipo+shared build_type=RelWithDebInfo cuda_arch=70 arch=linux-rhel7-power9le
-module load exasgd-magma/2.5.4/cuda-10.2.89/gcc-7.4.0-oltcefd
+#module load exasgd-magma/2.5.4/cuda-10.2.89/gcc-7.4.0-oltcefd
+#module load exasgd-magma/2.6.1/cuda-11.0.194/gcc-7.4.0-s75j4m7
```
nychiang, line 24: Should we remove these two lines?

nychiang, line 24: Should we remove these two lines?

src/LinAlg/hiopLinSolverIndefDenseMagma.cpp
```diff
@@ -57,87 +122,25 @@ namespace hiop
...
+    return negEigVal;
   }
 
-  bool hiopLinSolverIndefDenseMagmaBuKa::compute_inertia(int N, 
-                                                         int *ipiv,
-                                                         int& posEigVal, 
-                                                         int& negEigVal, 
-                                                         int& nullEigVal)
+  bool hiopLinSolverIndefDenseMagmaBuKa::
+  compute_inertia(int N, int* ipiv, int& posEigVal, int& negEigVal, int& nullEigVal)
```
nychiang, line 132: use code style guide.

src/LinAlg/hiopLinSolverIndefDenseMagma.cpp
```diff
@@ -57,87 +122,25 @@ namespace hiop
...
   {
-    double det[2];
     int inert[3];
-    int job = 100;
-    int info;
+    int info, retcode;
```
nychiang, line 135: style guide.

src/LinAlg/hiopLinSolverIndefDenseMagma.cpp
```diff
@@ -345,7 +369,9 @@ namespace hiop
     assert(M_->n() == M_->m());
     assert(x.get_size()==M_->n());
     int N=M_->n(), LDA = N, LDB=N;
```
nychiang, line 371: use style guide.

nychiang (approved):
LGTM. Just left some comments about code style.

src/LinAlg/hiopLinSolverIndefDenseMagma.cpp
cnpetra: this does not violate codestyle guide

src/LinAlg/hiopLinSolverIndefDenseMagma.cpp
cnpetra: fixed

pelesh:
Building in "developer mode" with stricter warning flags produces a large number of warnings. I will submit a separate issue on this because it is out of scope for this PR.

src/LinAlg/hiopLinSolverIndefDenseMagma.cpp
```diff
@@ -1,49 +1,116 @@
...
+    
+    const int nrhs=1;
+    lddb_ = ldda_;
+
+    magmaRet = magma_malloc((void**)&dinert_, 3*sizeof(int));
+    assert(MAGMA_SUCCESS == magmaRet);
+
+
+    magmaRet = magma_imalloc_pinned(&ipiv_, ldda_);
+    assert(MAGMA_SUCCESS == magmaRet);
```
pelesh, line 33: Do you need to use pinned memory here?

src/LinAlg/hiopLinSolverIndefDenseMagma.cpp
```diff
@@ -397,4 +424,137 @@ namespace hiop
     return true;
   }
 
+#if 0
+  hiopLinSolverIndefDenseMagmaBuKa_old2::hiopLinSolverIndefDenseMagmaBuKa_old2(int n, hiopNlpFormulation* nlp_)
+    : hiopLinSolverIndefDense(n, nlp_), work_(NULL)
+  {
+    ipiv_ = new int[n];
+  }
```
pelesh, line 432: Do we need to keep commented out code here?

src/LinAlg/hiopLinSolverIndefDenseMagma.hpp
```diff
@@ -151,6 +184,58 @@ class hiopLinSolverIndefDenseMagmaNopiv : public hiopLinSolverIndefDense
...
+ * Bunch-Kaufmann. This is a numerically stable factorization with decent peak 
+ * performance on large matrices.
+ *
+ * Superceeded by BuKa solver - code is disabled
+ */
+class hiopLinSolverIndefDenseMagmaBuKa_old2 : public hiopLinSolverIndefDense
+{
+public:
+  hiopLinSolverIndefDenseMagmaBuKa_old2(int n, hiopNlpFormulation* nlp_);
+  virtual ~hiopLinSolverIndefDenseMagmaBuKa_old2();
```
pelesh, line 200: Do we need to keep commented out code in this PR?

pelesh (approved):
Builds like a charm, all tests pass. I think we can merge.

src/LinAlg/hiopLinSolverIndefDenseMagma.cpp
cnpetra: it looks like... it's not documented by magma, but in their example uses pinned memory and magma crashes with other memory allocation types..

src/LinAlg/hiopLinSolverIndefDenseMagma.cpp
cnpetra: will be removed later , but for now will keep it just in case we need to rollback parts of it.

PR #295, opened 2021-08-08
FindHiopCudaLibraries.cmake Fix

### LINKED ISSUE #293: HiOp fails to link to CUDA when RAJA is disabled.
On at least one Power9 CI platform, HiOp fails to link to cuda when Raja is disabled.

CameronRutherford:
Closes #293. I ended up removing culibos and nvblas from being explicitly linked against, as well as using the `CUDA::` namespace as recommended in the documentation. This is passing CI on all platforms so I think it is good to go, although I am now suspicious as to how this was working before.


cnpetra:
I tested this manually on ascent and it fixes/closes #293.

ashermancinelli (approved):

PR #296, opened 2021-08-10
Add updated summit variables

ashermancinelli:
As per offline conversation, these are updated modules to allow building on Summit:

```console
$ MY_CLUSTER=summit ./BUILD.sh --build-only
```


cnpetra:
looks good to me... I am wondering why the gihub CI is failing?

ashermancinelli:
This is out of date since the rhel8 upgrade, we'll need another PR for this...

PR #300, opened 2021-08-20
Add mechanism to override launch commands

ashermancinelli:
Added a mechanism to override the launch commands used by ctest and bumped the version of HiOp in the CMake. This was really helpful when using an interactive session was not as straightforward.


cnpetra (approved):

ashermancinelli:
@cnpetra Thanks! I don't have permissions to merge this, so merge whenever you're ready.

PR #301, opened 2021-08-22
Sparse gpu kernel dev 1

nychiang:
In this PR, I addressed the following functions and their corresponding unit tests:

hiopMatrixRajaSparseTriplet::copyRowsFrom
hiopMatrixRajaSparseTriplet::copy_rows_block_from
hiopMatrixRajaSparseTriplet::copySubDiagonalFrom
hiopMatrixRajaSparseTriplet::setSubDiagonalTo
hiopVectorIntRaja::copy_from

According to the discussion with Cosmin, I will address other RAJA functions in different PRs.
Will make the following PRs as short as possible.


nychiang:
@cnpetra @pelesh we can merge this one. Since I need to change 11 files to address Slaven's comments about "It is probably a better idea to implement method to set all elements of an abstract int vector to a value.", I will file another PR for those changes.

PR #312, opened 2021-09-03
Cuda configuration fix

### LINKED ISSUE #311: build with cuSolver but without STRUMPACK
Problem: while building with HIOP_USE_STRUMPACK OFF but HIOP_USE_CUSOLVER ON I am getting an error during `make`: make fails on the first file that requires CUDA. If both are on (STRUMPACK and CUSOLVER), it compiles ok.

Why this happens: somehow when STRUMPACK is ON, cuda includes are added to the compile commands. But even though cusolver explicitly links cuda and relevant libraries, they are not added to compile.

Not sure how to fix. 

branch: cusolver-rebase-dev.
@CameronRutherford @pelesh

CameronRutherford:
Resolves #311

Some notes:
- When building with static +strumpack, I noticed that the strumpack library would report lots or complaints about incompatibility with toc optimization. I notice that this also reports searching for header files in `/scratch` for @ashermancinelli... `toc optimization is not supported for 0xe7e90002 instruction`
- When building with shared + stumpack, I get information about requiring PIC to be enabled:
```
call to `strumpack::TaskTimer::~TaskTimer()' lacks nop, can't restore toc; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
collect2: error: ld returned 1 exit status
```

That being said, if we enforce compiling with only static libs for strumpack, this seems to be a functioning commit. Further discussion may well be necessary @pelesh @kswirydo

### DIFF
--- a/cmake/FindHiopCudaLibraries.cmake
+++ b/cmake/FindHiopCudaLibraries.cmake
@@ -9,16 +9,6 @@ add_library(hiop_cuda INTERFACE)
 
 find_package(CUDAToolkit REQUIRED)
 
-# # Uncomment for old workaround
-# target_link_libraries(hiop_cuda INTERFACE 
-#   culibos
-#   nvblas
-#   cusparse
-#   cusolver
-#   cudart
-#   cublasLt
-#   )
-
 if(HIOP_BUILD_SHARED)
   target_link_libraries(hiop_cuda INTERFACE 
     nvblas # shared only
--- a/cmake/FindHiopCudaLibraries.cmake
+++ b/cmake/FindHiopCudaLibraries.cmake
@@ -21,7 +21,7 @@ find_package(CUDAToolkit REQUIRED)
 
 if(HIOP_BUILD_SHARED)
   target_link_libraries(hiop_cuda INTERFACE 
-    CUDA::nvblas # shared only
+    nvblas # shared only
     CUDA::cusparse
     CUDA::cudart
     cublas
--- a/cmake/FindHiopCudaLibraries.cmake
+++ b/cmake/FindHiopCudaLibraries.cmake
@@ -9,32 +9,35 @@ add_library(hiop_cuda INTERFACE)
 
 find_package(CUDAToolkit REQUIRED)
 
-target_link_libraries(hiop_cuda INTERFACE 
-  culibos
-  nvblas
-  cusparse
-  cusolver
-  cudart
-  cublasLt
-  )
+# # Uncomment for old workaround
+# target_link_libraries(hiop_cuda INTERFACE 
+#   culibos
+#   nvblas
+#   cusparse
+#   cusolver
+#   cudart
+#   cublasLt
+#   )
 
-# if(HIOP_BUILD_SHARED)
-#   target_link_libraries(hiop_cuda INTERFACE 
-#     CUDA::cusparse
-#     CUDA::cudart
-#     CUDA::cublasLt
-#     CUDA::cusolver
-#     )
-# endif()
-# if(HIOP_BUILD_STATIC)
-#   target_link_libraries(hiop_cuda INTERFACE
-#     CUDA::culibos
-#     CUDA::cusparse_static
-#     CUDA::cudart_static
-#     CUDA::cublasLt_static
-#     CUDA::cusolver_static
-#     )
-# endif()
+if(HIOP_BUILD_SHARED)
+  target_link_libraries(hiop_cuda INTERFACE 
+    CUDA::nvblas # shared only
+    CUDA::cusparse
+    CUDA::cudart
+    cublas
+    cublasLt
+    CUDA::cusolver
+  )
+endif()
+if(HIOP_BUILD_STATIC)
+  target_link_libraries(hiop_cuda INTERFACE
+    CUDA::cusparse_static
+    CUDA::cudart_static
+    cublas
+    cublasLt
+    CUDA::cusolver_static
+  )
+endif()
 
 install(TARGETS hiop_cuda EXPORT hiop-targets)
 


cmake/FindHiopCudaLibraries.cmake
```diff
@@ -9,32 +9,35 @@ add_library(hiop_cuda INTERFACE)
...
-#     )
-# endif()
+endif()
+if(HIOP_BUILD_STATIC)
+  target_link_libraries(hiop_cuda INTERFACE
+    CUDA::cusparse_static
+    CUDA::cudart_static
+    cublas
+    cublasLt
+    CUDA::cusolver_static
```
pelesh, line 38: @CameronRutherford, is there a reason why some of these library names are prefixed with the namespace `CUDA::` and others are not? It is a bit confusing.

pelesh:
This indeed seems to fix the issue #311. @CameronRutherford, do you understand how changing the way CUDA toolkit libraries were added as target link libraries all of a sudden made STRUMPACK the linchpin for all things CUDA in HiOp build? These seem to me as two unrelated things, yet the empirical evidence suggest otherwise.

CameronRutherford:
> This indeed seems to fix the issue #311. @CameronRutherford, do you understand how changing the way CUDA toolkit libraries were added as target link libraries all of a sudden made STRUMPACK the linchpin for all things CUDA in HiOp build? These seem to me as two unrelated things, yet the empirical evidence suggest otherwise.

@pelesh hopefully I can give some explanation as to what I found. I don't quite understand how STRUMPACK fit's into HiOp, so perhaps there are some intricacies to this I am missing.

When you enable STRUMPACK within HiOp, you use `find_package` to search for it. This in turn incorporates all of the CMake that STRUMPACK exports. It is possible another library that STRUMPACK links against works around the bug in HiOps CMake.

The errors I were seeing were todo with static CUDA libs trying to link against shared CUDA libs, and I also know that STRUMPACK is only built with static library support. I also know some CUDA libraries are only shared libraries, so there could be some best practice that I am missing here to best solve this situation.

cmake/FindHiopCudaLibraries.cmake
CameronRutherford: See my comment for more context. The workaround that we are using here is to link against either the static *or* shared version of these libraries though the `LD_LIBRARY_PATH`. This seems to resolve the issue of some libraries not finding the right shared/static library call at build time, but I am still not satisfied with it being a good final solution.

ashermancinelli: So cublas and cublasLt can be either shared or static, but we mandate cudart and cusolver be static? Seems like we should allow shared and static for all libraries in that case

CameronRutherford: If it's the case that we want to link against both static and shared variants of these CUDA libraries, then we could incorporate that logic into the CMake directly, rather than inferring it and hoping the linker does the right thing. I imagine declaring `HIOP_LINK_CUDA_SHARED` and `HIOP_LINK_CUDA_STATIC` as advanced CMake options could work.

Part of the issue in this case is the statically built STRUMPACK links against the shared cuda libs in the export CMake, and so the linker has to interpret which one to use at link time (?) to work around this issue. I imagine linking both `CUDA::cublas` and `CUDA::cublas_static` would have the equivalent affect as this fix. Note that this fix seemingly would only apply when building with STRUMPACK.

If we want to unconditionally try and link against both then we can, but perhaps we can decide on a better practice here...

kswirydo:
I tested your fix @CameronRutherford It works correctly.

pelesh (approved):

pelesh:
@CameronRutherford, thanks for your explanations.

@kswirydo, I think you can merge this PR. We can decouple completely STRUMPACK from the cuSolver interface after the merge.

ashermancinelli:
Looks like the gh actions build is failing - @CameronRutherford I'm not sure why this is manifesting in your pr in particular, but could you disable GPU in the cmake options for the github actions tests?

CameronRutherford:
> Looks like the gh actions build is failing - @CameronRutherford I'm not sure why this is manifesting in your pr in particular, but could you disable GPU in the cmake options for the github actions tests?

I'm happy to disable GPU but this PR is probably messing up because it's merging from my custom branch into @kswirydo's branch. I don't want to get too far into debugging the failing MR as I'm not confident the original branch if fully functioning, but I can disable CUDA if you still want.

cmake/FindHiopCudaLibraries.cmake
```diff
@@ -9,32 +9,35 @@ add_library(hiop_cuda INTERFACE)
...
+    cublas
+    cublasLt
+    CUDA::cusolver
   )
-
-# if(HIOP_BUILD_SHARED)
-#   target_link_libraries(hiop_cuda INTERFACE 
-#     CUDA::cusparse
-#     CUDA::cudart
-#     CUDA::cublasLt
```
ashermancinelli, line 25: Can we remove the commented code

ashermancinelli (changes requested):

ashermancinelli:
> > Looks like the gh actions build is failing - @CameronRutherford I'm not sure why this is manifesting in your pr in particular, but could you disable GPU in the cmake options for the github actions tests?
> 
> @ashermancinelli I'm happy to disable GPU but this PR is probably messing up because it's merging from my custom branch into @kswirydo's branch. I don't want to get too far into debugging the failing MR as I'm not confident the original branch if fully functioning, but I can disable CUDA if you still want.

So it's not going into develop, I see. Yes, this is fine.

pelesh:
@CameronRutherford and @ashermancinelli: Why is mirroring to CI pipelines failing? We don't care too much for github action tests, but the branch needs to pass our CI pipeline tests.

ashermancinelli:
> @CameronRutherford and @ashermancinelli: Why is mirroring to CI pipelines failing? We don't care too much for github action tests, but the branch needs to pass our CI pipeline tests.

Looks like an issue with the PNNL runners. I'm already talking wiht the PNNL gitlab folks about what's going wrong.

ashermancinelli:
@CameronRutherford Could you check on the pnnl/ornl pipelines and rerun any that were failing due to _no space left on device_ error?

CameronRutherford:
I re-ran the tests that failed and it looks like they are passing now. I think this should be good to merge unless there is any more discussion...

PR #318, opened 2021-09-09
Re-enable build matrix with Spack in GitHub Actions

ashermancinelli:
This PR re enables the build matrix we were previously using, but using github actions. We may add similar builds in the future for GPU-enabled configurations, but this should provide a basis.

We would also like to run the tests via spack as well, but as per some conversations on the spack slack server, the spack test command has some issues that need to be resolved before we incorporate it into our CI.

### DIFF
--- a/.github/workflows/spack_build.yml
+++ b/.github/workflows/spack_build.yml
@@ -1,0 +1,40 @@
+name: Spack Builds
+
+on: [push]
+
+jobs:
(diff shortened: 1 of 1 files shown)


ashermancinelli:
feel free to recommend additional builds to add to the list

cnpetra (approved):

PR #321, opened 2021-09-14
Sparse kernel development for GPU. Part 2

nychiang:
This PR includes the following changes:
1. add hiopMatrixRajaSparseTriplet::copySubMatrixFromTrans and unit test
2. add hiopMatrixRajaSparseTriplet::copySubMatrixFrom and unit test

PR #322, opened 2021-09-14
Sparse kernel development for GPU. Part 3

nychiang:
Draft!
Will change the base to `develop` after PR #321


nychiang:
Function `copyDiagMatrixToSubblock_w_pattern` tries to extract the nonzeros from a given vector and put them into a smaller vector while keeping the order of nonzeros.
For example:
```
input_vec = [1 2 3 4]
input_pattern = [0 1 0 1]
output = [2 4]
```
Without using RAJA::inclusive_scan, I use a temporary vector in CPU to compute the location in output.
Note that this CPU code is required in every iteration.

Should we leave it right now and wait for the fix of RAJA::inclusive_scan_inplace?
Otherwise we can create a map between the pattern and the index in output, and reuse it. 
See [this](https://github.com/LLNL/hiop/blob/4701dc8b682e0a006a547cbe68a6b31bb5a2fb60/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp#L1576-L1614) for more details.

@cnpetra @pelesh @ashermancinelli

ashermancinelli:
Do we know what the issue is with that raja algorithm? Is there an expected release with a fix?

nychiang:
> Do we know what the issue is with that raja algorithm? Is there an expected release with a fix?
It seems this bug will be fixed in the next RAJA release in Nov --- https://github.com/LLNL/RAJA/issues/1037

cnpetra:
> Function `copyDiagMatrixToSubblock_w_pattern` tries to extract the nonzeros from a given vector and put them into a smaller vector while keeping the order of nonzeros.
> For example:
> 
> ```
> input_vec = [1 2 3 4]
> input_pattern = [0 1 0 1]
> output = [2 4]
> ```
> 
> Without using RAJA::inclusive_scan, I use a temporary vector in CPU to compute the location in output.
> Note that this CPU code is required in every iteration.
> 
> Should we leave it right now and wait for the fix of RAJA::inclusive_scan_inplace?
> Otherwise we can create a map between the pattern and the index in output, and reuse it.
> See [this](https://github.com/LLNL/hiop/blob/4701dc8b682e0a006a547cbe68a6b31bb5a2fb60/src/LinAlg/hiopMatrixRajaSparseTriplet.cpp#L1576-L1614) for more details.
> 
> @cnpetra @pelesh @ashermancinelli

if this is not used for MDS linear algebra, then your solution with CPU preprocessing if fine (for now). If it's used for MDS, then we should implement a basic scan that keeps everything in the device mem space.

nychiang:
> 
> if this is not used for MDS linear algebra, then your solution with CPU preprocessing if fine (for now). If it's used for MDS, then we should implement a basic scan that keeps everything in the device mem space.

It is NOT used for MDS. This function is only used for sparse linear algebra.

ashermancinelli:
This seems fine for now then. We can bump to the newest raja as soon as it's released and functional on our target platforms.

PR #327, opened 2021-09-22
Preprocessor logic for gpu compute mode

### LINKED ISSUE #310: Preprocessor logic for GPU-enabled linear solvers should return error if none is available
In classes `hiopDualsLsqUpdateLinsysAugSparse` and `hiopKKTLinSysCompressedSparseXYcYd` preprocessor logic will look for GPU-enabled linear solver when user options other than "cpu" are enabled. In other words, if user selects "hybrid", "auto" or "gpu", the preprocessor directives will first try to use a GPU-enabled linear solver, but if none is available, they will fall back on a CPU-based solver. The code relies on unified memory to manage all host-device data transfers.

Consider returning error if user tries to use "hybrid" or "gpu" mode without GPU-enabled linear solver. I believe logic in HiOp user options already hides "hybrid" and "gpu" options from the user in cases when no GPU-enabled linear solver is available. Conflict between requested and available linear solvers in KKT linear system or duals updater would indicate a bug in options logic, so returning error here would be more appropriate than quietly falling back to a CPU linear solver. 

CC @kswyrido

cnpetra:
This PR changes the behavior of the sparse linear solvers for device computations. Under hybrid or gpu compute mode, HiOp will stop if no GPU-based linear solver is available. Previously, HiOp was (silently) instantiating a CPU-based solver under this scenario.

Closes #310.

### DIFF
--- a/src/Optimization/hiopDualsUpdater.cpp
+++ b/src/Optimization/hiopDualsUpdater.cpp
@@ -4,7 +4,7 @@
 //
 // This file is part of HiOp. For details, see https://github.com/LLNL/hiop. HiOp 
 // is released under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause). 
-// Please also read “Additional BSD Notice” below.
+// Please also read "Additional BSD Notice" below.
 //
 // Redistribution and use in source and binary forms, with or without modification, 
 // are permitted provided that the following conditions are met:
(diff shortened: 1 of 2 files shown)


nychiang (approved):

nychiang (commented):
Some CI failed due to no connection to the platform.
I think this PR is straightforward and hence I just approved it.

PR #340, opened 2021-10-14
Enable cuda separable compilation

ashermancinelli:
This is required to build with Umpire 6.0.0 and/or cuda >= 11, and I believe this is one of the issues @jwang125 ran into last FY.

CC @CameronRutherford @pelesh


ashermancinelli:
This took a bit to track down, but we won't be able to build shared libraries in CI anymore, only static. I found this line in Umpire's spack packge:

```yaml
    # device allocator exports device code, which requires static libs
    # currently only available for cuda.
    conflicts('+shared', when='+cuda')
```

This apparently means we as users can't build our own shared libraries with Umpire's device code either.

Issue #251, opened 2021-06-08
PriDec solver int pointers


cnpetra:
In a couple of places in the PriDec solver, the int pointers are allocated on the host. This need to be double checked to make sure this is intended (i.e., they are used only in the the scenario/contingency bookeeping and is ok to stay on host or indexes in arithmetic ops and need to be "ported")

cnpetra:
closed by PR#359

PR #359, opened 2021-11-08
replaced int* with hiopVectorInt in PriDec solver

### LINKED ISSUE #250: copyFrom and copyTo of hiopVector
the methods need to be adjusted to work with hiopVectorInt and supported mem spaces

cnpetra:
closes #250 and #251

Main target is to replace int arrays used for indexing with hiopVectorInt. Both Seq and RAJA implementations are done. 

The PR also fixes a number of memory leaks in the PriDec solver. Both single and MPI PriDec drivers are now leaks-free.

### DIFF
--- a/src/Drivers/nlpPriDec_ex8.cpp
+++ b/src/Drivers/nlpPriDec_ex8.cpp
@@ -61,7 +61,6 @@ Ex8::Ex8(int ns_, int S_, bool include, hiopInterfacePriDecProblem::RecourseAppr
 
 Ex8::~Ex8()
 {
-  //delete evaluator_;
 }
  
 bool Ex8::get_prob_sizes(size_type& n, size_type& m)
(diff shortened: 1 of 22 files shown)


nychiang:
Since we have function `copy_from_indexes` and `hiopVectorInt`, should we define the `ixl`, `ixu`, `idl` and `idu` (the pattern vector for variable/constraint bounds) as `hiopVectorInt`? 
See https://github.com/LLNL/hiop/blob/8b2468e814a03ab2acd0c3fc20d68c4868cd1350/src/LinAlg/hiopVector.hpp#L69.

Note that we may need to change other functions and files, such as the functions with name `xxxx_w_pattern` in the vector/matrix kernel. 
@cnpetra

jwang125 (approved):
Tests all work except for 16 NlpDenseCons3_50k_mpi on Lassen. Using same compiling options as 2 months ago when everything worked.

cnpetra:
> Tests all work except for 16 NlpDenseCons3_50k_mpi on Lassen. Using same compiling options as 2 months ago when everything worked.

strange, it passes all the CI tests. can you run manually and email me the log? verbosity_level 4 or higher

nychiang (approved):

PR #365, opened 2021-11-18
Evaluate functions after adjust primal variables

nychiang:
Fix the a bug which prevents the convergence of Ex3 on lassen.


jwang125 (approved):
Tests all passed!

cnpetra (approved):

Issue #349, opened 2021-10-28
Export HiOp Sparse configuration through CMake


CameronRutherford:
In ExaGO we were using `check_cxx_symbol_exists` in CMake in order to detect if `HIOP_SPARSE` was defined in `hiop_defs.hpp`. 

This worked for some time, but at some point this stopped working. We ended up needing something along the following lines in our FindHiop custom module to work around the issue:

```CMake
set(CMAKE_REQUIRED_FLAGS -I${MPI_C_INCLUDE_DIRS} ${MPI_C_COMPILE_OPTIONS})
set(CMAKE_REQUIRED_LIBRARIES ${MPI_C_LIBRARIES})
```
While this works, a better practice would be to also export `HIOP_SPARSE` as a variable in HiOp's export configuration https://github.com/LLNL/hiop/blob/develop/cmake/HiOpConfig.cmake.in to avoid the issue altogether.

We will be addressing this at PNNL and will create a PR with the necessary changes.

cc @ashermancinelli

PR #369, opened 2021-11-23
Export sparse config

jaelynlitz:
addresses issue #349 

exports the target HiOp::SPARSE so that it is accessible after install


CameronRutherford (approved):
This is much needed in order to to make exporting the HiOp sparse configuration for other projects easier.

ashermancinelli (approved):
Yes, I believe this is ready. CC @cnpetra.

PR #373, opened 2021-12-01
Update xsdk policy compliance document

ashermancinelli:
Updates to the compliance document as per our conversation with Cody and Satish this week.


cnpetra (approved):
I've added the CHANGELOG and did a couple of small edits to the compliance document.

ashermancinelli:
Even better! I think we are in very good shape.

PR #377, opened 2021-12-17
code reorg for NLP formulation classes

cnpetra:
Reorganized code so that the bounds and constraint preprocessing sit in (virtual) functions. Cleaned up the code and adopted the newer coding styleguide.


nychiang (approved):
LGTM

PR #378, opened 2021-12-18
Add HIP+Clang build to CI

ashermancinelli:
Note that this only _builds_ on Incline, the PNNL AMDGPU system. This is due to some an issue with the Slurm configuration at PNNL, which is being tracked in an internal ticket. All tests pass when I run locally. I'll enable the tests on Incline as soon as it's ready.


ashermancinelli:
CC @CameronRutherford

cnpetra (approved):

PR #379, opened 2021-12-20
Relax EQ to INEQ 

cnpetra:
This PR offers a NLP formulation class that relaxes the equality constraints to two-sided inequality constraints.

Adds an option for eq rhs relax factor and updates Ex7 to test the new class.


nychiang (approved):
LGTM

PR #390, opened 2022-01-07
maintainance of linear solvers code

cnpetra:
- changed name from IndefSparse to SymSparse

 - changed the sys matrix member to pointer (from reference) and to the more general class hiopMatrixSparse to accomodate future use of CSR sparse matrices types


nychiang (approved):
I just suggests three commits to correct the name of member functions.

PR #395, opened 2022-01-13
Iterative refinement for condensed linsys 

cnpetra:
Adds iterative refinement for the condensed linsys via BiCGStab preconditioned with solves based on Cholesky factorization of the "condensed" matrix. 

Documents the two-loop linear solve from `hiopAlgFilterIPM` 

Fixes a couple of small bugs and outputting issues in the BiCGStab class.

### DIFF
--- a/src/Optimization/hiopKKTLinSysSparseCondensed.cpp
+++ b/src/Optimization/hiopKKTLinSysSparseCondensed.cpp
@@ -69,19 +69,26 @@ hiopKKTLinSysCondensedSparse::hiopKKTLinSysCondensedSparse(hiopNlpFormulation* n
   : hiopKKTLinSysCompressedSparseXDYcYd(nlp),
     dd_pert_(nullptr),
     JtDiagJ_(nullptr),
-    M_condensed_(nullptr)
+    M_condensed_(nullptr),
+    delta_wx_(0.),
+    krylov_mat_opr_(nullptr),
+    krylov_prec_opr_(nullptr),
+    krylov_rhs_xdycyd_(nullptr),
+    bicgstab_(nullptr)
 {
 }
 
 hiopKKTLinSysCondensedSparse::~hiopKKTLinSysCondensedSparse()
 {
+  delete bicgstab_;
+  delete krylov_rhs_xdycyd_;
+  delete krylov_prec_opr_;
+  delete krylov_mat_opr_;
   delete M_condensed_;
   delete JtDiagJ_;
   delete dd_pert_;
 }
 
-   
-  
 bool hiopKKTLinSysCondensedSparse::build_kkt_matrix(const double& delta_wx_in,
                                                     const double& delta_wd_in,
                                                     const double& dcc,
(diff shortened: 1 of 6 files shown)


src/Optimization/hiopKKTLinSysSparseCondensed.cpp
```diff
@@ -562,16 +559,84 @@ bool hiopKKTLinSysCondensedSparse::solveCompressed(hiopVector& rx,
...
+  nlp_->runStats.kkt.tmSolveRhsManip.start();
+
+  // this is rhs used by the direct solve
+  if(rhs_ == NULL) {
+    rhs_ = LinearAlgebraFactory::create_vector(nlp_->options->GetString("mem_space"), nx);
+  }
+  assert(rhs_->get_size() == nx);
+
+  nlp_->log->write("RHS KKT_SPARSE_Condensed rx: ", rx,  hovIteration);
+  nlp_->log->write("RHS KKT_SPARSE_Condensed rx: ", rd,  hovIteration);
```
nychiang, line 599: "RHS KKT_SPARSE_Condensed rd: ", it's `rd`, not `rx`

src/Optimization/hiopKKTLinSysSparseCondensed.cpp
```diff
@@ -562,16 +559,84 @@ bool hiopKKTLinSysCondensedSparse::solveCompressed(hiopVector& rx,
...
+  nlp_->log->write("RHS KKT_SPARSE_Condensed ryd:", ryd, hovIteration);
+#if 0
+  bret = solve_compressed_direct(rx, rd, ryc, ryd, dx, dd, dyc, dyd);
+#else
+  
+  if(nullptr == krylov_mat_opr_) {
+    krylov_mat_opr_ = new hiopKKTMatVecOpr(this);
+    krylov_prec_opr_ = new hiopKKTPrecondOpr(this);
+    krylov_rhs_xdycyd_ = new hiopVectorPar(nx+nd+nyd);
+    bicgstab_ = new hiopBiCGStabSolver(nx+nd+nyd, "DEFAULT", krylov_mat_opr_, krylov_prec_opr_);
```
nychiang, line 610: should we add a note here to check/replace the hardcode "DEFAULT" in the future?

src/Optimization/hiopKKTLinSysSparseCondensed.hpp
```diff
@@ -208,26 +216,39 @@ class hiopKKTLinSysCondensedSparse : public hiopKKTLinSysCompressedSparseXDYcYd
...
+                               hiopVector& rd,
+                               hiopVector& ryc,
+                               hiopVector& ryd,
+                               hiopVector& dx,
+                               hiopVector& dd,
+                               hiopVector& dyc,
+                               hiopVector& dyd);
 protected:
-
+  virtual bool solve_compressed_direct(hiopVector& rx,
```
nychiang, line 228: maybe it's a good idea to add a brief description to this function and the above one (even though I know the difference :-))

src/Optimization/hiopKKTLinSysSparseCondensed.cpp
```diff
@@ -562,16 +559,84 @@ bool hiopKKTLinSysCondensedSparse::solveCompressed(hiopVector& rx,
...
+  dx.startingAtCopyFromStartingAt(0, *krylov_rhs_xdycyd_, 0);
+  dd.startingAtCopyFromStartingAt(0, *krylov_rhs_xdycyd_, nx);
+  dyd.startingAtCopyFromStartingAt(0, *krylov_rhs_xdycyd_, nx+nd);
+  
+#endif
+
   nlp_->log->write("SOL KKT_SPARSE_Condensed dx: ", dx,  hovMatrices);
   nlp_->log->write("SOL KKT_SPARSE_Condensed dd: ", dd,  hovMatrices);
   nlp_->log->write("SOL KKT_SPARSE_Condensed dyc:", dyc, hovMatrices);
   nlp_->log->write("SOL KKT_SPARSE_Condensed dyd:", dyd, hovMatrices);
```
nychiang, line 637: I see know we only record the RHS and SOL for this method. Should we also record the matrix, e.g., the KKT matrix which is used in solve_compressed_direct?

nychiang (approved):
LGTM. Just some minor issues.

src/Optimization/hiopKKTLinSysSparseCondensed.cpp
cnpetra: it's on the TODO, waiting proper resolution of CSR and COO formats...

PR #397, opened 2022-01-21
Adjust var/con bounds after adjusting small slacks

nychiang:
After adjusting small slacks, reset the bounds according to the current variables and slacks.

### DIFF
--- a/src/Optimization/hiopIterate.cpp
+++ b/src/Optimization/hiopIterate.cpp
@@ -82,24 +82,24 @@ hiopIterate::hiopIterate(const hiopNlpFormulation* nlp_)
 
 hiopIterate::~hiopIterate()
 {
-  if(x) delete x;
-  if(d) delete d;
-  if(sxl) delete sxl;
-  if(sxu) delete sxu;
-  if(sdl) delete sdl;
-  if(sdu) delete sdu;
+  delete x;
+  delete d;
+  delete sxl;
+  delete sxu;
+  delete sdl;
+  delete sdu;
   delete sx_arg1_;
   delete sx_arg2_;
   delete sx_arg3_;
   delete sd_arg1_;
   delete sd_arg2_;
   delete sd_arg3_;
-  if(yc) delete yc;
-  if(yd) delete yd;
-  if(zl) delete zl;
-  if(zu) delete zu;
-  if(vl) delete vl;
-  if(vu) delete vu;
+  delete yc;
+  delete yd;
+  delete zl;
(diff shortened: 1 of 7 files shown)


src/Optimization/hiopIterate.cpp
```diff
@@ -396,67 +408,122 @@ bool hiopIterate::updateDualsIneq(const hiopIterate& iter, const hiopIterate& di
 }
 */
 
-int hiopIterate::adjust_small_slacks(hiopVector& slack, 
-                                     const hiopVector& bound, 
-                                     const hiopVector& slack_dual, 
-                                     const hiopVector& select,
-                                     const double& mu)
+int hiopIterate::adjust_small_var_slacks(hiopVector& slack,
```
cnpetra, line 411: I liked it better before, when was only one `adjust_small_slacks_vars` or similarly named method. The functionality of the two news methods is identical. I suggest to go back to one function and pass the cached pointers as arguments to the method, would that be too much hassle?

src/Optimization/hiopIterate.cpp
```diff
@@ -65,6 +65,12 @@ hiopIterate::hiopIterate(const hiopNlpFormulation* nlp_)
   sxu = x->alloc_clone();
   sdl = d->alloc_clone();
   sdu = d->alloc_clone();
+  sx_arg1_ = x->alloc_clone();
```
cnpetra, line 68: I would allocate on demand, in the `adjust_small_slacks` method. I think bound pushing occurs rather infrequently ...

src/Optimization/hiopIterate.hpp
```diff
@@ -87,13 +87,20 @@ class hiopIterate
...
-  /// @brief adjust small slack variables
+  /// @brief adjust slack variables if they are negative, or if they are positive but too small
   virtual int adjust_small_slacks(const hiopIterate& iter_curr, const double& mu);
-  virtual int adjust_small_slacks(hiopVector& slack, 
-                                  const hiopVector& bound, 
-                                  const hiopVector& slack_dual, 
-                                  const hiopVector& select,
-                                  const double& mu);
+  /// @brief adjust slack variables sxl or sxu, i.e., introduced by variable bounds
+  virtual int adjust_small_var_slacks(hiopVector& slack,
```
cnpetra, line 93: It would help to document the math a little bit.

cnpetra (changes requested):

cnpetra (approved):

PR #406, opened 2022-02-07
Update linear solver safe mode

cnpetra:
The outer optimization loop switches between "safe" and "fast" KKT formulations when various convergence metrics (or lack of them) occur in the Krylov-based iterative refinement for the GPU Cholesky solves and in the filter line-search algorithm. Builds on the "safe mode" idea used for MDS linear algebra but now it allows switching between KKT formulations on the fly during the optimization.

### DIFF
--- a/src/Optimization/hiopAlgFilterIPM.cpp
+++ b/src/Optimization/hiopAlgFilterIPM.cpp
@@ -877,7 +877,7 @@ void hiopAlgFilterIPMBase::displayTerminationMsg()
     {
       nlp->log->printf(hovSummary, "Couldn't solve the problem.\n");
       nlp->log->printf(hovSummary, "Linesearch returned unsuccessfully (small step). Probable cause: "
-		       "inaccurate gradients/Jacobians or infeasible problem.\n");
+		       "inaccurate gradients/Jacobians or locally infeasible problem.\n");
       nlp->log->printf(hovSummary, "%s\n", strStatsReport.c_str());
       break;
     }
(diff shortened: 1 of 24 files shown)


nychiang:
Rebase to or merge with `develop` to clear the CI failure on PNNL gitlab

src/Optimization/hiopAlgFilterIPM.cpp
```diff
@@ -1055,7 +1055,7 @@ hiopSolveStatus hiopAlgFilterIPMQuasiNewton::run()
     /************************************************
      * update mu and other parameters
      ************************************************/
-    while(_err_log<=kappa_eps * _mu) {
+    if(_err_log<=kappa_eps * _mu) {
       //update mu and tau (fraction-to-boundary)
       bret = updateLogBarrierParameters(*it_curr, _mu, _tau, _mu, _tau);
       if(!bret) break; //no update is necessary
```
nychiang, line 1061: why do we change this `while` loop to `if`? I think `while` is designed to reduce `mu` faster.
In addition, if we change to `if`,  the `break` command on line 1061 may break the outer loop and give us unpredictable result.

cnpetra:
merged this PR in `csr-interface-dev` and we will review and discuss it there.

src/Optimization/hiopAlgFilterIPM.cpp
cnpetra: under `elastic_mode` using `while` tighten up bounds too quickly and enters FR quite often. Also, aggressive reduction puts a lot of stress on the linear solver for some reason. Not sure using `while` for the non`elastic_mode` outweigh the, but I can leave the `while` on for nonelastic mode.

Will remove fix the `break`.

PR #407, opened 2022-02-07
CSR interface

cnpetra:
This PR proposes a common interface for CPU and GPU CSR matrix and provides an efficient CPU implementation.


src/Optimization/hiopAlgFilterIPM.cpp
```diff
@@ -1055,7 +1055,7 @@ hiopSolveStatus hiopAlgFilterIPMQuasiNewton::run()
     /************************************************
      * update mu and other parameters
      ************************************************/
-    while(_err_log<=kappa_eps * _mu) {
+    if(_err_log<=kappa_eps * _mu) {
       //update mu and tau (fraction-to-boundary)
       bret = updateLogBarrierParameters(*it_curr, _mu, _tau, _mu, _tau);
       if(!bret) break; //no update is necessary
```
nychiang, line 1061: why do we change this while loop to if? I think while is designed to reduce mu faster.
In addition, if we change to if, the break command on line 1061 may break the outer loop and give us unpredictable result.

cnpetra: I switched back the `while` loop. It behaves like before with `elastic_mode` off, but when it's off, it only does one reduction of log barrier params. see commit #[55d8333](https://github.com/LLNL/hiop/pull/407/commits/55d83330df4024dfd40dcd4dfe3a1c25433ccf4d)

pelesh:
### Suggestion:

This may be beyond the scope of this PR, but it would be worthwhile considering to have sequential sparse matrix implementation using _unordered_ triplet format and separate thread parallel linear algebra (including RAJA) implementations using CSR format. Triplet sparse matrices would retain CPU-only implementation. The inheritance structure would look something like this:
```
Matrix -+--- Dense
        |
        +--- Sparse -+--- Triplet (sequential only)
                     |
                     +--- CSR -+--- Sequential
                               |
                               +--- RAJA
                               |
                               +--- CUDA hand optimized
                               |
                               +--- OpenMP
                               |
                              ...
```

### Pros:

1. Currently HiOp is not quite compatible with Ipopt models because HiOp requires triplet matrix format to be ordered. Having dedicated linear algebra using triplet format (albeit sequential) would provide a way to maintain compatibility with Ipopt models. The `hybrid` mode with only linear solver on GPU would still work with triplet fromat.
2. CSR sparse matrices are more suitable for thread-parallel implementation. Using CSR would simplify RAJA kernels and allow us to remove a lot of clutter and hacks from the existing code. Also, whoever wants to use thread-parallel implementation would prefer to write his models using CSR rather than triplet matrix format for similar reasons.

### Cons:

1. Biggest concern would be the maintenance cost for HiOp, but it seems to me the decision to support triplet and CSR formats in a long run has been already made.
2. Would require more intrusive refactor of the matrix linear algebra code.

cnpetra:
> ### Suggestion:
> This may be beyond the scope of this PR, but it would be worthwhile considering to have sequential sparse matrix implementation using _unordered_ triplet format and separate thread parallel linear algebra (including RAJA) implementations using CSR format. Triplet sparse matrices would retain CPU-only implementation. The inheritance structure would look something like this:
> 
> ```
> Matrix -+--- Dense
>         |
>         +--- Sparse -+--- Triplet (sequential only)
>                      |
>                      +--- CSR -+--- Sequential
>                                |
>                                +--- RAJA
>                                |
>                                +--- CUDA hand optimized
>                                |
>                                +--- OpenMP
>                                |
>                               ...
> ```
> 
> ### Pros:
> 1. Currently HiOp is not quite compatible with Ipopt models because HiOp requires triplet matrix format to be ordered. Having dedicated linear algebra using triplet format (albeit sequential) would provide a way to maintain compatibility with Ipopt models. The `hybrid` mode with only linear solver on GPU would still work with triplet fromat.
> 2. CSR sparse matrices are more suitable for thread-parallel implementation. Using CSR would simplify RAJA kernels and allow us to remove a lot of clutter and hacks from the existing code. Also, whoever wants to use thread-parallel implementation would prefer to write his models using CSR rather than triplet matrix format for similar reasons.
> 
> ### Cons:
> 1. Biggest concern would be the maintenance cost for HiOp, but it seems to me the decision to support triplet and CSR formats in a long run has been already made.
> 2. Would require more intrusive refactor of the matrix linear algebra code.

good summary and points, we have open issues for most of them. Besides pros and cons, there are also a couple of "unknowns", related to CUDA and ROCm (changing) APIs as well as to limited experience and profiling with compute_mode GPU and memory_space device. I am already spotting potential performance bottlenecks. But the near term plan is to push CSR as a replacement for triplet, from the lower tiers toward the user interface, and when we have have a good feeling of the potential bottlenecks and solutions for them, to refactor the user interface and NLP formulation(s) glue code.

PR #412, opened 2022-02-18
Inertia free fix for condensed system

nychiang:
With condensed linear system, current implementation add dual regularization if a Cholesky solver fails, e.g. matrix is not p.d. This PR adds primal regularizations in stead of the dual ones.

In some test cases, the un-regularized matrix can be Cholesky factorized successfully, while adding a small primal regularization fails the Cholesky solver. This PR also addressed this issue, by increasing the primal regularization till the Cholesky solver succeeds.


cnpetra:
note: fix typo https://github.com/LLNL/hiop/blob/2fd0edf19bfae48418b08fc463c3d997989ceed9/src/Optimization/hiopKKTLinSys.cpp#L395

PR #413, opened 2022-02-18
Add option to redirect ctest output directory.

pelesh:
HiOp CMake tests are hardwired to write output to the build directory. On some clusters schedulers may not have permission to write in user's build directory. This minor change to CMake will add an option to configure where CMake tests will write output files. The option is marked as advanced, so most users won't notice any difference in HiOp configuration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,6 +40,8 @@ option(HIOP_WITH_VALGRIND_TESTS "Run valgrind on certain integration tests" OFF)
 option(HIOP_BUILD_DOCUMENTATION "Build HiOp documentation via Doxygen" ON)
 
 set(HIOP_CTEST_LAUNCH_COMMAND "" CACHE STRING "Manually override launch commands used when running CTest")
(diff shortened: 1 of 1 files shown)


cnpetra (approved):

CMakeLists.txt
```diff
@@ -40,6 +40,8 @@ option(HIOP_WITH_VALGRIND_TESTS "Run valgrind on certain integration tests" OFF)
 option(HIOP_BUILD_DOCUMENTATION "Build HiOp documentation via Doxygen" ON)
 
 set(HIOP_CTEST_LAUNCH_COMMAND "" CACHE STRING "Manually override launch commands used when running CTest")
+set(HIOP_CTEST_OUTPUT_DIR ${PROJECT_BINARY_DIR} CACHE PATH "Directory where CTest outputs are saved")
+mark_as_advanced(HIOP_CTEST_OUTPUT_DIR) # Hide this var as the default will be overriden rarely
```
ashermancinelli, line 44: [nitpick] I'm not sure this needs to be advanced. IMO it just hurts discoverability, and if users are going to mess with cmake variables they don't understand, marking this var as advanced won't prevent them from doing so.

ashermancinelli (approved):

Issue #244, opened 2021-06-01
Replace standard library function with Hiop ones


jwang125:
Currently using std::max/min in the code. Hoping to replace them with HiopMax and HiopMin so that all the linear algebra can be encapsulated in Hiop. Applies to other commonly used functions as well.

ashermancinelli:
Have `HiopMax`/`HiopMin` been implemented yet?

cnpetra:
@nychiang : random functions also fit here.

cnpetra:
we also need 


- random uniform 
- random normal
- absolute value
- min
- max

PR #414, opened 2022-02-19
Fixes for magma inertia fix

ashermancinelli:
* Resolves nearly all warnings from clang
* Adds `scripts/spockVariables.sh` script for setting up development environment on spock
* Updates to `BUILD.sh` script so it runs cleanly on Spock with `MY_CLUSTER=spock ./BUILD.sh` just like on all other CI clusters
* Uses modules I just built there, including magma from master branch


src/Drivers/CMakeLists.txt
```diff
@@ -71,11 +71,6 @@ if(HIOP_USE_MPI AND HIOP_SPARSE)
         nlpPriDec_ex9_sparse_raja.cpp 
         nlpPriDec_ex9_sparse_raja_driver.cpp 
         PROPERTIES LANGUAGE CUDA)
-    elseif(HIOP_USE_HIP)
```
nychiang, line 74: Why should we remove this `elseif` section for HIP?

src/LinAlg/hiopMatrixDenseRowMajor.cpp
```diff
@@ -237,15 +237,15 @@ void hiopMatrixDenseRowMajor::copyFromMatrixBlock(const hiopMatrixDense& srcmat,
 void hiopMatrixDenseRowMajor::shiftRows(size_type shift)
 {
   if(shift==0) return;
-  if(fabs(shift)==m_local_) return; //nothing to shift
+  if(std::fabs(shift)==m_local_) return; //nothing to shift
```
pelesh, line 240: Why is `fabs()` used here? Did you intend to use `std::abs()`? There are several other places in the code where `fabs` is used with integer arguments and where the return value is also expected to be an integer. Building in developer mode or with `-Wconversion` flag should catch these instances.

src/LinAlg/hiopMatrixRajaDense.cpp
```diff
@@ -1365,7 +1365,7 @@ double hiopMatrixRajaDense::max_abs_value()
   RAJA::forall<hiop_raja_exec>(RAJA::RangeSegment(0, n_local_ * m_local_),
     RAJA_LAMBDA(RAJA::Index_type i)
     {
-      norm.max(fabs(dd[i]));
+      norm.max(std::fabs(dd[i]));
```
pelesh, line 1368: I am not sure if this is appropriate implementation. I am not sure if CUDA, for example, supports `std` math functions.

src/LinAlg/hiopMatrixRajaDense.cpp
```diff
@@ -1403,7 +1403,7 @@ void hiopMatrixRajaDense::row_max_abs_value(hiopVector &ret_vec)
     RAJA_LAMBDA(RAJA::Index_type i)
     {
       for(int j = 0; j < n_local; j++) {
-        double abs_val = fabs(Mview(i, j));
+        double abs_val = std::fabs(Mview(i, j));
```
pelesh, line 1406: Same here: Use of `std::fabs` is questionable here because this code is supposed to be portable.

src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
```diff
@@ -1163,10 +1163,10 @@ void hiopMatrixRajaSparseTriplet::print(FILE* file,
 
     if(NULL==msg) {
       if(numranks>1)
-        fprintf(file, "matrix of size %lld %lld and nonzeros %lld, printing %d elems (on rank=%d)\n", 
+        fprintf(file, "matrix of size %d %d and nonzeros %d, printing %d elems (on rank=%d)\n",
```
pelesh, line 1166: I suggest replacing this with streaming operators or casting all integer arguments of `fprintf` function to `int`s. The former is probably a better long term solution. Nothing really guarantees you that e.g. `n()` will return an `int` instead of `long long`.

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -1918,9 +1918,9 @@ void hiopVectorRajaPar::print(FILE* file, const char* msg/*=NULL*/, int max_elem
     if(NULL==msg)
     {
       if(numranks>1)
-	      fprintf(file, "vector of size %lld, printing %d elems (on rank=%d)\n", n_, max_elems, myrank);
+	      fprintf(file, "vector of size %d, printing %d elems (on rank=%d)\n", n_, max_elems, myrank);
       else
-	      fprintf(file, "vector of size %lld, printing %d elems (serial)\n", n_, max_elems);
+	      fprintf(file, "vector of size %d, printing %d elems (serial)\n", n_, max_elems);
```
pelesh, line 1923: I would replace this with streaming operators, which are overloaded, so we get correct output regardless of what is set for HiOp's `index_type` or `size_type`

pelesh (changes requested):
Need to ensure functions from `std` library are used correctly before merging.

src/Drivers/CMakeLists.txt
ashermancinelli: Because we already link against HIP device libraries in `cmake/FindHiopHipLibraries.cmake`, which is all that's actually needed for compiling HIP. It's unclear what the CMake best practice will be for this sort of thing, but at least from discussions on Spack and AMD slack channels, it isn't needed.

src/LinAlg/hiopMatrixDenseRowMajor.cpp
ashermancinelli: I kept the function the same and just added the `std` namespace qualifier, but it does seem like a few of these should have originally been `abs` instead of `fabs`. I can changes these.

src/Drivers/CMakeLists.txt
nychiang: Similarly, can we remove the previous section about `HIOP_USE_CUDA` since we have already link against CUDA via `cmake/FindHiopCUDALibraries.cmake`?

ashermancinelli: No, we probably want to use the CUDA CMake language since there are additional steps with CUDA C++. For example, we want to link with Umpire's device code using the nvidia compiler, and that will only happen if the source language is CUDA.

src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
ashermancinelli: We should really be using `std::format` from [fmtlib](https://github.com/fmtlib/fmt), but I'll concede to use streams for now :)

src/LinAlg/hiopMatrixRajaSparseTriplet.cpp
```diff
@@ -1162,12 +1163,14 @@ void hiopMatrixRajaSparseTriplet::print(FILE* file,
...
+      ss << "matrix of size " << m() << " " << n() << " and nonzeros "
+        << numberOfNonzeros() << ", printing " << max_elems
+        << " elems";
       if(numranks>1)
-        fprintf(file, "matrix of size %lld %lld and nonzeros %lld, printing %d elems (on rank=%d)\n", 
-		m(), n(), numberOfNonzeros(), max_elems, myrank_);
-      else
-        fprintf(file, "matrix of size %lld %lld and nonzeros %lld, printing %d elems\n", 
-		m(), n(), numberOfNonzeros(), max_elems);
+        ss << " (on rank=" << myrank_ << ")";
```
nychiang, line 1171: add brackets to this `if` statement, to follow the code-style guide

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -1917,10 +1918,14 @@ void hiopVectorRajaPar::print(FILE* file, const char* msg/*=NULL*/, int max_elem
...
     if(NULL==msg)
     {
+      std::stringstream ss;
+      ss << "vector of size " << n_ << ", printing " << max_elems << " elems ";
       if(numranks>1)
-	      fprintf(file, "vector of size %lld, printing %d elems (on rank=%d)\n", n_, max_elems, myrank);
+        ss << "(on rank=" << myrank << ")";
       else
-	      fprintf(file, "vector of size %lld, printing %d elems (serial)\n", n_, max_elems);
+        ss << "(serial)";
```
nychiang, line 1926: add brackets in this section

nychiang (commented):
LGTM. Besides the minor changes I listed in the review, I also wonder if we should change `fabs` to `abs` in the other files as well.
For example, `fabs` is also used in `hiopVectorRajaPar.cpp`, `hiopVectorPar.cpp`, `hiopMatrixRajaSparseTriplet.cpp` and `hiopMatrixSparseTriplet.cpp`.

nychiang:
@cnpetra @pelesh I think we are ready to merge this to `develop` and hence I suggest that we should change the base class of this PR to `develop`

src/LinAlg/hiopMatrixDenseRowMajor.cpp
```diff
@@ -237,15 +237,15 @@ void hiopMatrixDenseRowMajor::copyFromMatrixBlock(const hiopMatrixDense& srcmat,
 void hiopMatrixDenseRowMajor::shiftRows(size_type shift)
 {
   if(shift==0) return;
-  if(fabs(shift)==m_local_) return; //nothing to shift
+  if(abs(shift)==m_local_) return; //nothing to shift
```
pelesh, line 240: I believe everything that is outside RAJA loops can use `std::abs`. For inside RAJA loops, it is better to use `fabs` or similar as CUDA (and I believe HIP, as well) stick to C interface for math functions.

pelesh: For inside RAJA loops, see #244.

pelesh (approved):
I think this is ready to merge. Optionally, you may want to use C++ std math library functions (which are overloaded) _outside_ RAJA loops.

scripts/ascentVariables.sh
```diff
@@ -3,82 +3,57 @@ module use /gpfs/wolf/proj-shared/csc359/src/spack/share/spack/modules/linux-rhe
...
-module load cuda/11.0.1
-module load gcc/7.4.0
-module load spectrum-mpi/10.3.1.2-20200121
-module load cmake/3.18.2
-
-#These are not picked up by the module for some reason
-export CC=/sw/ascent/gcc/7.4.0/bin/gcc
-export CXX=/sw/ascent/gcc/7.4.0/bin/g++
-export FC=/sw/ascent/gcc/7.4.0/bin/gfortran
+module load gcc/9.1.0
```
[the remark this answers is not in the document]
ashermancinelli: It should not matter in this case. But feel free to try it out locally.

src/LinAlg/hiopMatrixDenseRowMajor.cpp
ashermancinelli: Since it seems like #244 has not seen much movement, I'll use naked `(f)abs` for RAJA and `std::` qualified `(f)abs` elsewhere. I'd like to resolve this in this PR.

ashermancinelli:
This PR has been updated to target develop branch. Once it's merged, I will update spack packages accordingly.

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -852,7 +853,7 @@ double hiopVectorRajaPar::infnorm_local() const
   RAJA::forall< hiop_raja_exec >( RAJA::RangeSegment(0, n_local_),
     RAJA_LAMBDA(RAJA::Index_type i)
     {
-      norm.max(std::abs(data[i]));
+      norm.max(abs(data[i]));
```
pelesh, line 856:
```suggestion
      norm.max(fabs(data[i]));
```
When using math functions from C library, it is `fabs`, otherwise it is absolute value of an integer; C++ std::abs is overloaded so you can pass floating point or integer argument to it.

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -888,7 +889,7 @@ double hiopVectorRajaPar::onenorm_local() const
   RAJA::forall< hiop_raja_exec >( RAJA::RangeSegment(0, n_local_),
     RAJA_LAMBDA(RAJA::Index_type i)
     {
-      sum += std::abs(data[i]);
+      sum += abs(data[i]);
```
pelesh, line 892:
```suggestion
      sum += fabs(data[i]);
```
Same here.

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -1365,7 +1366,7 @@ void hiopVectorRajaPar::invert()
     RAJA_LAMBDA(RAJA::Index_type i)
     {
 #ifdef HIOP_DEEPCHECKS
-      assert(std::abs(data[i]) > small_real);
+      assert(abs(data[i]) > small_real);
```
pelesh, line 1369:
```suggestion
      assert(fabs(data[i]) > small_real);
```
and here.

ashermancinelli:
I found a couple more bugs I'd like to address before this is merged.

ashermancinelli:
@cnpetra I think this is ready for merge, and we can then cut v0.5.4.

cnpetra (approved):

PR #421, opened 2022-03-04
Fix FR by using Example 10

nychiang:
By using example 10 from PR #419, I detected some mismatches in the FR implementation. 
This PR is a fix to those mismatches, and it adds an infeasible problem to CI test.

### DIFF
--- a/src/Optimization/hiopDualsUpdater.hpp
+++ b/src/Optimization/hiopDualsUpdater.hpp
@@ -117,6 +117,19 @@ public:
                   const hiopIterate& search_dir, const double& alpha_primal, const double& alpha_dual,
                   const double& mu, const double& kappa_sigma, const double& infeas_nrm_trial);
 
+  /** LSQ update of the constraints duals (yc and yd). Source file describe the math. */
+  virtual bool go(hiopIterate& it_ini,
+                  const hiopVector& grad_f,
+                  const hiopMatrix& jac_c,
+                  const hiopMatrix& jac_d)
+  {
+    //nlp_->log->printf(hovSummary,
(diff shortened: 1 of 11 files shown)


src/Optimization/hiopDualsUpdater.hpp
```diff
@@ -117,6 +117,19 @@ class hiopDualsLsqUpdate : public hiopDualsUpdater
...
+  virtual bool go(hiopIterate& it_ini,
+                  const hiopVector& grad_f,
+                  const hiopMatrix& jac_c,
+                  const hiopMatrix& jac_d)
+  {
+    //nlp_->log->printf(hovSummary,
+    //                  "LSQ Dual Initialization --- Dense linsys: size %d (%d eq-cons)\n",
+    //                  nlp_->m_eq()+nlp_->m_ineq(), nlp_->m_eq());  
+    bool bret = do_lsq_update(it_ini,grad_f,jac_c,jac_d);
+    return bret;
```
pelesh, line 130: Why is log output commented out?

pelesh:
There is large number of warnings when rebuilding only ex. 10 driver in developer's mode, but most of them seem to be inherited from HiOp.

There is a fall through warning for the case statement in the ex. 10 driver, but it seems the fall through cases is by design.

pelesh (commented):
Builds cleanly (w/o developer's mode warning flags) and passes all tests.

pelesh (approved):

src/Optimization/hiopDualsUpdater.hpp
nychiang: I should remove this part, since this dual updater can be used in any IPM iteration instead of the initialization.

Issue #432, opened 2022-03-17
CSR matrix classes organization


cnpetra:
### Suggestion:

It would be worthwhile to consider having sequential sparse matrix implementation using _unordered_ triplet format and separate thread parallel linear algebra (including RAJA) implementations using CSR format. Triplet sparse matrices would retain CPU-only implementation. The inheritance structure would look something like this:
```
Matrix -+--- Dense
        |
        +--- Sparse -+--- Triplet (sequential only)
                     |
                     +--- CSR -+--- Sequential
                               |
                               +--- RAJA
                               |
                               +--- CUDA hand optimized
                               |
                               +--- OpenMP
                               |
                              ...
```

### Pros:

1. Currently HiOp is not quite compatible with Ipopt models because HiOp requires triplet matrix format to be ordered. Having dedicated linear algebra using triplet format (albeit sequential) would provide a way to maintain compatibility with Ipopt models. The `hybrid` mode with only linear solver on GPU would still work with triplet fromat.
2. CSR sparse matrices are more suitable for thread-parallel implementation. Using CSR would simplify RAJA kernels and allow us to remove a lot of clutter and hacks from the existing code. Also, whoever wants to use thread-parallel implementation would prefer to write his models using CSR rather than triplet matrix format for similar reasons.

### Cons:

1. Biggest concern would be the maintenance cost for HiOp, but it seems to me the decision to support triplet and CSR formats in a long run has been already made.
2. Would require more intrusive refactor of the matrix linear algebra code.

cnpetra:
closed by PR #436

PR #436, opened 2022-03-22
Csr interface organization

cnpetra:
Reorganizes CSR Matrix class by creating a base virtual CSR Matrix interface. Existing Host code is moved to a "Seq" class. Skeleton for the CUDA CSR (that will be using cuSPARSE API and potentially some hand written code) is also created.

Closes issue #432

### DIFF
--- a/src/Drivers/nlpSparse_ex6_driver.cpp
+++ b/src/Drivers/nlpSparse_ex6_driver.cpp
@@ -167,7 +167,7 @@ int main(int argc, char **argv)
   
   nlp.options->SetStringValue("compute_mode", "cpu");
   nlp.options->SetStringValue("KKTLinsys", "xdycyd");
-  nlp.options->SetStringValue("write_kkt", "yes");
+  //nlp.options->SetStringValue("write_kkt", "yes");
 
   nlp.options->SetNumericValue("mu0", 0.1);
   nlp.options->SetStringValue("options_file_fr_prob", "hiop_fr_ci.options");
@@ -176,6 +176,7 @@ int main(int argc, char **argv)
     nlp.options->SetStringValue("linear_solver_sparse", "pardiso");
   }
   if(use_cusolver) {
+    nlp.options->SetStringValue("linsol_mode", "speculative");
     nlp.options->SetStringValue("linear_solver_sparse", "cusolver");
     nlp.options->SetStringValue("compute_mode", "hybrid");
   }
--- a/src/Drivers/nlpSparse_ex6_driver.cpp
+++ b/src/Drivers/nlpSparse_ex6_driver.cpp
@@ -179,6 +179,7 @@ int main(int argc, char **argv)
     nlp.options->SetStringValue("linsol_mode", "speculative");
     nlp.options->SetStringValue("linear_solver_sparse", "cusolver");
     nlp.options->SetStringValue("compute_mode", "hybrid");
+    nlp.options->SetStringValue("fact_acceptor", "inertia_free");
   }
   if(force_fr) {
     nlp.options->SetStringValue("force_resto", "yes");
--- a/src/LinAlg/hiopMatrixSparseCSRCUDA.hpp
+++ b/src/LinAlg/hiopMatrixSparseCSRCUDA.hpp
@@ -54,7 +54,7 @@
 #ifdef HIOP_SPARSE_MATRIX_CSRCUDA
 #define HIOP_SPARSE_MATRIX_CSRCUDA
 
-#ifndef HIOP_USE_CUDAaaa
+#ifdef HIOP_USE_CUDA
 
 #include "hiopVector.hpp"
 #include "hiopMatrixDense.hpp"
@@ -68,7 +68,7 @@ namespace hiop
 {
 
 /**
- * @brief Sparse matrix of doubles in compressed row format for use on CPU/host. Data
+ * @brief Sparse matrix of doubles in compressed row format for use on CUDA GPUs. Data
  * is not (memory, MPI) distributed.
  *
  * @note The methods of this class expect and maintains unique and ordered column indexes 
--- a/src/LinAlg/hiopMatrixSparseCSRCUDA.hpp
+++ b/src/LinAlg/hiopMatrixSparseCSRCUDA.hpp
@@ -582,16 +582,7 @@ protected:
   index_type* row_starts_;
   
 private:
-  hiopMatrixSparseCSRCUDA(const hiopMatrixSparseCSRCUDA&)
-    : hiopMatrixSparseCSR(0, 0, 0),
-      irowptr_(nullptr),
-      jcolind_(nullptr),
-      values_(nullptr),
-      buf_col_(nullptr),
-      row_starts_(nullptr)
-  {
-    assert(false && "copy constructor disabled");
-  }
+  hiopMatrixSparseCSRCUDA(const hiopMatrixSparseCSRCUDA&) = delete;
 };
 
 
--- a/src/LinAlg/hiopMatrixSparseCSRCUDA.hpp
+++ b/src/LinAlg/hiopMatrixSparseCSRCUDA.hpp
@@ -1,0 +1,601 @@
+// Copyright (c) 2022, Lawrence Livermore National Security, LLC.
+// Produced at the Lawrence Livermore National Laboratory (LLNL).
+// LLNL-CODE-742473. All rights reserved.
+//
+// This file is part of HiOp. For details, see https://github.com/LLNL/hiop. HiOp
+// is released under the BSD 3-clause license (https://opensource.org/licenses/BSD-3-Clause).
+// Please also read "Additional BSD Notice" below.
+//
+// Redistribution and use in source and binary forms, with or without modification,
+// are permitted provided that the following conditions are met:
+// i. Redistributions of source code must retain the above copyright notice, this list
+// of conditions and the disclaimer below.
+// ii. Redistributions in binary form must reproduce the above copyright notice,
+// this list of conditions and the disclaimer (as noted below) in the documentation and/or
+// other materials provided with the distribution.
+// iii. Neither the name of the LLNS/LLNL nor the names of its contributors may be used to
+// endorse or promote products derived from this software without specific prior written
+// permission.
+//
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY
+// EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES
+// OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT
+// SHALL LAWRENCE LIVERMORE NATIONAL SECURITY, LLC, THE U.S. DEPARTMENT OF ENERGY OR
+// CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+// CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
+// OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED
+// AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
+// EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// Additional BSD Notice
+// 1. This notice is required to be provided under our contract with the U.S. Department
+// of Energy (DOE). This work was produced at Lawrence Livermore National Laboratory under
+// Contract No. DE-AC52-07NA27344 with the DOE.
+// 2. Neither the United States Government nor Lawrence Livermore National Security, LLC
+// nor any of their employees, makes any warranty, express or implied, or assumes any
+// liability or responsibility for the accuracy, completeness, or usefulness of any
+// information, apparatus, product, or process disclosed, or represents that its use would
+// not infringe privately-owned rights.
+// 3. Also, reference herein to any specific commercial products, process, or services by
+// trade name, trademark, manufacturer or otherwise does not necessarily constitute or
+// imply its endorsement, recommendation, or favoring by the United States Government or
+// Lawrence Livermore National Security, LLC. The views and opinions of authors expressed
+// herein do not necessarily state or reflect those of the United States Government or
+// Lawrence Livermore National Security, LLC, and shall not be used for advertising or
+// product endorsement purposes.
+
+/**
+ * @file hiopMatrixSparseCSRCUDA.hpp
+ *
+ * @author Cosmin G. Petra <petra1@llnl.gov>, LNNL
+ *
+ */
+#ifdef HIOP_SPARSE_MATRIX_CSRCUDA
+#define HIOP_SPARSE_MATRIX_CSRCUDA
+
+#ifndef HIOP_USE_CUDAaaa
+
+#include "hiopVector.hpp"
+#include "hiopMatrixDense.hpp"
+#include "hiopMatrixSparseCSR.hpp"
+#include "hiopMatrixSparseTriplet.hpp"
+
+#include <cassert>
+#include <unordered_map>
+
+namespace hiop
+{
+
+/**
+ * @brief Sparse matrix of doubles in compressed row format for use on CPU/host. Data
+ * is not (memory, MPI) distributed.
+ *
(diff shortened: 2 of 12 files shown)


nychiang (approved):
LGTM. Looking for the follow-up PRs.

src/LinAlg/hiopMatrixSparseCSRCUDA.hpp
```diff
@@ -0,0 +1,601 @@
...
+  hiopMatrixSparseCSRCUDA(const hiopMatrixSparseCSRCUDA&)
+    : hiopMatrixSparseCSR(0, 0, 0),
+      irowptr_(nullptr),
+      jcolind_(nullptr),
+      values_(nullptr),
+      buf_col_(nullptr),
+      row_starts_(nullptr)
+  {
+    assert(false && "copy constructor disabled");
+  }
```
ashermancinelli, line 594:
```suggestion
  hiopMatrixSparseCSRCUDA(const hiopMatrixSparseCSRCUDA&) = delete;
```

src/LinAlg/hiopMatrixSparseCSRSeq.hpp
```diff
@@ -0,0 +1,598 @@
...
+  hiopMatrixSparseCSRSeq(const hiopMatrixSparseCSRSeq&)
+    : hiopMatrixSparseCSR(0, 0, 0),
+      irowptr_(nullptr),
+      jcolind_(nullptr),
+      values_(nullptr),
+      buf_col_(nullptr),
+      row_starts_(nullptr)
+  {
+    assert(false && "copy constructor disabled");
+  }
```
ashermancinelli, line 592:
```suggestion
  hiopMatrixSparseCSRSeq(const hiopMatrixSparseCSRSeq&) = delete;
```

ashermancinelli (approved):

pelesh (changes requested):
Since we are starting fresh, I would suggest using class templates to specify execution policies for CSR matrices. That would eliminate need for separate `CSRseq` class and make the whole code more compact and easier to maintain.

src/LinAlg/hiopMatrixSparseCSRSeq.hpp
```diff
@@ -0,0 +1,598 @@
...
+
+  /// @brief Performs a quick check and returns false if the CSR indexes are not ordered
+  bool check_csr_is_ordered();
+  /////////////////////////////////////////////////////////////////////
+  // end of new CSR-specific methods
+  /////////////////////////////////////////////////////////////////////
+
+private:
+  void alloc();
+  void dealloc();
```
pelesh, line 559: This is great!

src/LinAlg/hiopMatrixSparseCSRSeq.hpp
pelesh: Why is copy constructor disabled? You could make it private and then use it to implement `clone` and `dep_copy` methods, which can be handy.

cnpetra:
> Since we are starting fresh, I would suggest using class templates to specify execution policies for CSR matrices. That would eliminate need for separate `CSRseq` class and make the whole code more compact and easier to maintain.

I am not following... CSRSeq is the reference CPU class. CUDA will be vendor specific. Will have one for AMD. You mean exec policy templates for a portable implementation, likely based on RAJA ?

pelesh:
Consider more compact design using class templates like this:
```c++
template <typename indexT = int, typename scalarT = double, class exec_policy = hiop::Seq>
class hiopMatrixSparseCSR
{
public:
...
 
private:
  indexT rowptrs_;
  indexT colidx_;
  scalarT values_;
}; 

Classes passed as `exec_policy` would contain bare bones (BLAS style) kernels. For each added policy one would implement only those kernels (or substitute third party ones), while using the same class template.

**Comment by pelesh** [comment]
> I am not following... CSRSeq is the reference CPU class. CUDA will be vendor specific. Will have one for AMD. You mean exec policy templates for a portable implementation, likely based on RAJA ?

I need to prototype that in a simple example to communicate it better. It does not need to be based on RAJA.

**Comment by ashermancinelli** [comment]
> > I am not following... CSRSeq is the reference CPU class. CUDA will be vendor specific. Will have one for AMD. You mean exec policy templates for a portable implementation, likely based on RAJA ?
> 
> I need to prototype that in a simple example to communicate it better. It does not need to be based on RAJA.

@cnpetra We may specialize certain methods for the exec policy template parameter, such that the kernels will by default just use sequential or CUDA exec policies in RAJA kernels, but then provide some specialization like so:

```c++
template<typename IndexT, typename ScalarT>
hiop::hiopMatrixSparseCSR<IndexT, ScalarT, hiop::Cuda>::some_specialized_method(...) {
  // some vendor-specific code here
}
```

Apologies if this is not what you were getting at @pelesh. I can provide a small example to show how this might work if that's helpful.

ashermancinelli:
However, if we're comfortable with C++17 (which is supported by every compiler we test with) we can use `if constexpr` which removes the need for as many template shenanigans. [Here's a full example of a matrix class with compile-time specialization](https://godbolt.org/z/s1oWGjKv6).

nychiang:
Thanks @ashermancinelli and @pelesh for the suggestion and example.
I understand how this approach works, but my experience is that it ends up a long file including all the specialized implementations. (For example, all the functions need to have this `if-else statement` and  all the specialized code must be implemented in the `if-else statement`. 
Is there a way to avoid that?

pelesh:
> I understand how this approach works, but my experience is that it ends up a long file including all the specialized implementations. (For example, all the functions need to have this `if-else statement` and all the specialized code must be implemented in the `if-else statement`.
> Is there a way to avoid that?

I believe what I propose in #437 would avoid using conditionals to select specialized source code and if I understand it correctly, @ashermancinelli proposal would do the same. The difference is I am (surprise, surprise) more conservative in using C++ newer features.

src/Optimization/hiopDualsUpdater.cpp
```diff
@@ -413,39 +413,63 @@ bool hiopDualsLsqUpdateLinsysAugSparse::do_lsq_update(hiopIterate& iter,
...
+
+      // This is our first choice on the device.
+      if(linear_solver == "cusolver" || linear_solver == "auto") {
+        hiopLinSolverIndefSparseCUSOLVER *p = new hiopLinSolverIndefSparseCUSOLVER(n, nnz, nlp_);
         nlp_->log->printf(hovSummary,
                           "LSQ Dual Initialization --- KKT_SPARSE_XDYcYd linsys: using CUSOLVER on device as an "
                           "indefinite solver, size %d (%d cons)\n",
                           n,
                           neq+nineq);
         p->setFakeInertia(neq + nineq);
```
pelesh, line 440: Remove fake inertia from here. It is not used anyways.
```suggestion
      if(linear_solver == "cusolver" || linear_solver == "auto") {
        hiopLinSolverIndefSparseCUSOLVER *p = new hiopLinSolverIndefSparseCUSOLVER(n, nnz, nlp_);
        nlp_->log->printf(hovSummary,
                          "LSQ Dual Initialization --- KKT_SPARSE_XDYcYd linsys: using CUSOLVER on device as an "
                          "indefinite solver, size %d (%d cons)\n",
                          n,
                          neq+nineq);
```

src/Optimization/hiopKKTLinSysSparse.cpp
```diff
@@ -572,70 +602,106 @@ namespace hiop
...
+        assert(nullptr == linSys_);
+        if(linear_solver == "cusolver" || linear_solver == "auto") {
+#if defined(HIOP_USE_CUSOLVER)        
+          hiopLinSolverIndefSparseCUSOLVER *p = new hiopLinSolverIndefSparseCUSOLVER(n, nnz, nlp_);
+          nlp_->log->printf(hovScalars,
+                            "KKT_SPARSE_XDYcYd linsys: alloc CUSOLVER size %d (%d cons) (hybrid)\n",
+                            n,
+                            neq+nineq);
+          p->setFakeInertia(neq + nineq);
+          linSys_ = p;
```
pelesh, line 635: Remove fake inertia:
```suggestion
          hiopLinSolverIndefSparseCUSOLVER *p = new hiopLinSolverIndefSparseCUSOLVER(n, nnz, nlp_);
          nlp_->log->printf(hovScalars,
                            "KKT_SPARSE_XDYcYd linsys: alloc CUSOLVER size %d (%d cons) (hybrid)\n",
                            n,
                            neq+nineq);
          linSys_ = p;
```

src/Optimization/hiopKKTLinSysSparse.cpp
```diff
@@ -572,70 +602,106 @@ namespace hiop
...
+        assert(nullptr == linSys_);
+        if(linear_solver == "cusolver" || linear_solver == "auto") {
+#if defined(HIOP_USE_CUSOLVER)        
+          hiopLinSolverIndefSparseCUSOLVER *p = new hiopLinSolverIndefSparseCUSOLVER(n, nnz, nlp_);
+          nlp_->log->printf(hovScalars,
+                            "KKT_SPARSE_XDYcYd linsys: alloc CUSOLVER size %d (%d cons) (gpu)\n",
+                            n,
+                            neq+nineq);
+          p->setFakeInertia(neq + nineq);
+          linSys_ = p;
```
pelesh, line 698: Remove fake inertia:
```suggestion
          hiopLinSolverIndefSparseCUSOLVER *p = new hiopLinSolverIndefSparseCUSOLVER(n, nnz, nlp_);
          nlp_->log->printf(hovScalars,
                            "KKT_SPARSE_XDYcYd linsys: alloc CUSOLVER size %d (%d cons) (gpu)\n",
                            n,
                            neq+nineq);
          linSys_ = p;
```

pelesh:
When running example 7:
```shell
jsrun -n 1 -g 1 ./src/Drivers/nlpSparse_ex7.exe 500 -cusolver -inertiafree -selfcheck
```
for condensed model I get following output:
```shell
LSQ Dual Updater --- KKT_SPARSE_XDYcYd linsys: using MA57 on CPU(!!!) size 1502 (501 cons)
Input Error: Incorrect objective type.
```
After that computation seems to run fine using Cholesky solver module:
```shell
...
[Warning] hiopLinSolverCholCuSparse: the 499th pivot is <=1.00000e-24
[Warning] solve compressed high absolute resid norm (= 8.45063e-02)
   7  2.1799950e+02 4.050e-13  2.726e-01  -1.40  9.321e-01  1.000e+00  1(s)
...
```
It is unclear to me what solver is invoked when, but it looks like first solve is done with MA57 and subsequent with Cholesky. Is that the idea here?

nychiang:
@pelesh In the first iteration, the dual variables are computed by a least-square subproblem. It is passed to MA57, and failed Not sure why the error msg is "Incorrect objective type", but it happens if the matrix is singular (Ex7 has rank deficient constraints). When the least-square initialization fails, HiOp set all the duals to zero and continue the iterates.
Therefore, you may want to set `duals_init = zero` to ignore the lsq initialization, and just use cuSolvers in HiOp.

pelesh:
> When the least-square initialization fails, HiOp sets all the duals to zero and continue the iterates.
> Therefore, you may want to set `duals_init = zero` to ignore the lsq initialization, and just use cuSolvers in HiOp.

@nychiang, you are right on the money. Setting `duals_init` to `zero` seems to fix the logic of selecting Cholesky solver, as well. It seems that using `lsq` for duals initialization breaks the solver selection logic when cuSolver (Cholesky or LU) is enabled. 

CC @kswirido

cnpetra:
shoot, @pelesh , I deleted you reply, please repost

cnpetra:
> @nychiang, you are right on the money. Setting duals_init to zero seems to fix the logic of selecting Cholesky solver, as well. It seems that using lsq for duals initialization breaks the solver selection logic when cuSolver (Cholesky or LU) is enabled.
> 
> CC @kswirido[kswirydo](https://github.com/kswirydo)

Just to clarify, the Cholesky solver should be selected in the second test in ex7 driver, independent of the other options (assuming CUDA is present and HIOP_SPARSE is on)...

pelesh:
> Just to clarify, the Cholesky solver should be selected in the second test in ex7 driver, when cusolver is required. the LU solver should solve the first test in the driver

Agreed!

cnpetra:
> > Just to clarify, the Cholesky solver should be selected in the second test in ex7 driver, when cusolver is required. the LU solver should solve the first test in the driver
> 
> Agreed!

does it work like that for both when duals initialization set to lsq and to zero? or you have seen otherwise?

pelesh:
> does it work like that for both when duals initialization set to lsq and to zero? or you have seen otherwise?

In #427 I find that @nychiang's suggestion to initialize duals with zero fixes issues with both cuSolver LU (first test) and Cholesky (secod test). In particular, in second test, it seems that setting duals initialization to zero also fixes the issue where LU factorization is called instead of Cholesky. It looks to me that requesting `lsq` initialization of duals messes up the solver selection logic.

cnpetra:
> > does it work like that for both when duals initialization set to lsq and to zero? or you have seen otherwise?
> 
> In #427 I find that @nychiang's suggestion to initialize duals with zero fixes issues with both cuSolver LU (first test) and Cholesky (secod test). In particular, in second test, it seems that setting duals initialization to zero also fixes the issue where LU factorization is called instead of Cholesky. It looks to me that requesting `lsq` initialization of duals messes up the solver selection logic.

I just checked and this PR fixes that.

src/Optimization/hiopDualsUpdater.cpp
cnpetra: it turns out that cuSOLVER LU still relies on the fake inertia in this branch and stops working if I remove it.

src/Drivers/nlpSparse_ex6_driver.cpp
```diff
@@ -176,6 +176,7 @@ int main(int argc, char **argv)
     nlp.options->SetStringValue("linear_solver_sparse", "pardiso");
   }
   if(use_cusolver) {
+    nlp.options->SetStringValue("linsol_mode", "speculative");
     nlp.options->SetStringValue("linear_solver_sparse", "cusolver");
     nlp.options->SetStringValue("compute_mode", "hybrid");
```
pelesh, line 181: Need to enable `inertia_free` for cuSolver:
```suggestion
    nlp.options->SetStringValue("linsol_mode", "speculative");
    nlp.options->SetStringValue("linear_solver_sparse", "cusolver");
    nlp.options->SetStringValue("compute_mode", "hybrid");
    nlp.options->SetStringValue("fact_acceptor", "inertia_free");
```

pelesh: Or merge with `cusolver-inertia-free-dev` :)

cnpetra: > Or merge with `cusolver-inertia-free-dev` :)

I hacked it for now
https://github.com/LLNL/hiop/blob/22ad66217e79c3d4086cc2c79e0cf908e722c681/src/Optimization/hiopKKTLinSysSparse.cpp#L638

pelesh: No, the issue is with sparse ex6, not "fake inertia".

cnpetra: it does not work with 
`
nlp.options->SetStringValue("fact_acceptor", "inertia_free");
`

src/Optimization/hiopDualsUpdater.cpp
```diff
@@ -413,39 +413,63 @@ bool hiopDualsLsqUpdateLinsysAugSparse::do_lsq_update(hiopIterate& iter,
...
+      //under compute mode gpu, at this point we don't have a sparse linear solver 
+      if(compute_mode == "gpu") {
+        if(linear_solver == "auto") {
+          assert("HiOp was not built with a sparse GPU and cannot fullfil the requirement of the option "
+                 "'duals_init_linear_solver_sparse'. Either build with a supported GPU sparse solver or "
+                 "change compute mode to hybrid, which will allow using a CPU sparse solver.");
+        } else {
+          assert("Impossible to deploy the (CPU?) sparse linear solver specified by option "
+                 "'duals_init_linear_solver_sparse' under gpu compute mode. Either build with a supported "
+                 "GPU sparse solver or change compute mode to hybrid, which will allow using CPU sparse solvers.");
```
nychiang, line 453: Two comments:
1. when we use this function, `do_lsq_update`, to initialize dual variables, we can return `false` whenever assertion fails. Then the outer loop ([here](https://github.com/LLNL/hiop/blob/e29104a18e380f9a537ddc2a960a4e34a95398cf/src/Optimization/hiopDualsUpdater.hpp#L139-L159)) will set duals to `zero`
2. this function can also be used in standard ipm iterations, e.g., solve the duals for quasi-newton method, and hence the error message may need to be modified. That is, the problem is not about `duals_init_linear_solver_sparse`, but `linear_solver_sparse` when `duals_update_type = lsq`.

cnpetra: good point, will revisit

src/Drivers/nlpSparse_ex6_driver.cpp
pelesh: I can run it on my machines without calls to fake inertia, see #440.

pelesh (approved):
The code builds correctly and all tests pass. The approach to add CSR matrices is consistent with the rest of HiOp architecture.

There is a large number of warnings when built in "developer's mode", which need to be addressed eventually, but that's beyond the scope of this PR.

PR #440, opened 2022-03-29
Remove fake inertia and fix example 6 accordingly

pelesh:
This PR is to demonstrate cuSOLVER working without fake inertia. I've been able to run sparse examples 6 and 7 from `csr-interface-hpp-dev` branch after removing fake inertia from cuSOLVER.

### DIFF
--- a/src/Optimization/hiopDualsUpdater.cpp
+++ b/src/Optimization/hiopDualsUpdater.cpp
@@ -439,7 +439,7 @@ bool hiopDualsLsqUpdateLinsysAugSparse::do_lsq_update(hiopIterate& iter,
                           "indefinite solver, size %d (%d cons)\n",
                           n,
                           neq+nineq);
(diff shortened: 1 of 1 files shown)


src/Optimization/hiopDualsUpdater.cpp
```diff
@@ -439,7 +439,7 @@ bool hiopDualsLsqUpdateLinsysAugSparse::do_lsq_update(hiopIterate& iter,
                           "indefinite solver, size %d (%d cons)\n",
                           n,
                           neq+nineq);
-        p->setFakeInertia(neq + nineq);
+        // p->setFakeInertia(neq + nineq);
```
cnpetra, line 442: you should have removed the `p` pointer and `dynamic_cast` above

cnpetra (approved):

Issue #439, opened 2022-03-29
linear solver selection in LSQ updater


cnpetra:
the (sparse) LSQ updater seems to instantiate linear solver based on the `duals_init_linear_solver_sparse` option only. It should instantiate based on `linear_solver_sparse` (when `duals_update_type` = `lsq`).

@nychiang

cnpetra:
closed by PR #441

Issue #417, opened 2022-02-22
Consider using conditionals instead of preprocessor directives in function `do_lsq_update`


pelesh:
The logic for selecting linear solver in function `hiopDualsLsqUpdateLinsysAugSparse::do_lsq_update` in file `hiopDualsUpdater.cpp` is implemented as combination of preprocessor directives and conditional statements. Consider replacing preprocessor directives with conditionals where arguments are constant expressions from structure `hiopVersion`. For example, replace
```c++
#ifdef HIOP_USE_GPU
  ...
#endif
```
with
```c++
if(hiopVersion::useGPU) {
  ...
}
```
Eliminating preprocessor directives could make code more readable and easier to debug/maintain.

CC @ashermancinelli

cnpetra:
logic should be more clear now that the "logic" code is much compact (PR #441).

PR #441, opened 2022-03-29
support for different LSQ linear solvers.

cnpetra:
addresses issues #439, #417

### DIFF
--- a/src/Optimization/hiopAlgFilterIPM.cpp
+++ b/src/Optimization/hiopAlgFilterIPM.cpp
@@ -76,7 +76,8 @@ hiopAlgFilterIPMBase::hiopAlgFilterIPMBase(hiopNlpFormulation* nlp_in, const boo
   //force completion of the nlp's initialization
   nlp->finalizeInitialization();
 }
-void hiopAlgFilterIPMBase::destructorPart()
+
+void hiopAlgFilterIPMBase::dealloc_alg_objects()
 {
   if(it_curr)  delete it_curr;
   if(it_trial) delete it_trial;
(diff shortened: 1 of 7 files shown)


src/Optimization/hiopDualsUpdater.hpp
```diff
@@ -148,18 +178,36 @@ class hiopDualsLsqUpdate : public hiopDualsUpdater
...
-                          ynrm, lsq_dual_init_max);
+                          ynrm,
+                          lsq_dual_init_max);
       }
     }
     return bret;
   }
 protected:
-  //method called by both 'go' and 'computeInitialDualsEq'
+  /// Helper method doing the work for both `go` and `computeInitialDualsEq`
```
nychiang, line 188: maybe it's a good time for us to rename `computeInitialDualsEq` as `compute_initial_duals`

src/Optimization/hiopDualsUpdater.cpp
```diff
@@ -336,64 +344,53 @@ hiopDualsLsqUpdateLinsysAugSparse::~hiopDualsLsqUpdateLinsysAugSparse()
   delete lin_sys_;
 }
 
-bool hiopDualsLsqUpdateLinsysAugSparse::do_lsq_update(hiopIterate& iter,
-                                                      const hiopVector& grad_f,
-                                                      const hiopMatrix& jac_c,
-                                                      const hiopMatrix& jac_d)
+bool hiopDualsLsqUpdateLinsysAugSparse::
+instantiate_linear_solver(const char* linsol_opt,
```
nychiang, line 348: To be honest, I'm thinking if we should move some logics into`hiopOptionsNLP::ensure_consistence()`.
Then we can have a much concise code in here and `hiopKKTLinSysXXX::determineAndCreateLinsys`.
For example, we have already downselect the combination of `KKTLinsys`, `linear_solver_sparse`, `compute_mode` and others in `ensure_consistence()`. Then we can implement `determineAndCreateLinsys` in the base class without worry about host or device, e.g.:
```
  // auto has been replaced by others in the preprocess
  if(linear_solver == "ma57") {
#ifdef HIOP_USE_COINHSL
    nlp_->log->printf( xxxx );
    linSys_ = new hiopLinSolverIndefSparseMA57(n, nnz, nlp_);
#endif // HIOP_USE_COINHSL
  } else  if(linear_solver == "cuslover") {
#ifdef HIOP_USE_CUSOLVER
    nlp_->log->printf( xxxx );
    hiopLinSolverIndefSparseCUSOLVER *p = new hiopLinSolverIndefSparseCUSOLVER(n, nnz, nlp_);
#endif // HIOP_USE_CUSOLVER
  } else if ...
```

nychiang: I think the proposed change can minimize our future work when some new linear solvers are added.

cnpetra: we should definitely update `ensure_consistence` with any new logic . But I think we should keep redundant logic, at least for some time, just to be safe.

cnpetra: also, some of the consistency checks for options need to be delayed and done later (in the Alg classes). This occurs because the Alg cannot always be determined based on the options and there are options that depend on/are applicable with specific Alg implementations. See `reload_options` in the Newton IPM Alg class.

nychiang: Got it!

src/Optimization/hiopAlgFilterIPM.cpp
```diff
@@ -928,12 +886,56 @@ void hiopAlgFilterIPMBase::displayTerminationMsg()
...
+  if(duals_update_type==0) {
+    dualsUpdate_ = nlp->alloc_duals_lsq_updater();
+  } else if(duals_update_type==1) {
+    dualsUpdate_ = new hiopDualsNewtonLinearUpdate(nlp);
+  } else {
+    assert(false && "duals_update_type has an unrecognized value");
+  }
+
+  resetSolverStatus();
 }
```
nychiang, line 939: Many variables are initialized in the derived classes now, not in the base class `hiopAlgFilterIPMBase` anymore.
Any particular reason to duplicate these codes?

nychiang (approved):
lgtm. 
Cannot figure out what happen if both `lsq_dual_updater` and `lsq_dual_initialization` are used, but I think we don't have such plan for the near future.

src/Optimization/hiopAlgFilterIPM.cpp
cnpetra: yes, because `reload_options` called by the constructor is virtual and needs to be called in the child classes. But I will add a couple of methods to hold the common code.

cnpetra:
> lgtm. Cannot figure out what happen if both `lsq_dual_updater` and `lsq_dual_initialization` are used, but I think we don't have such plan for the near future.

they are not used for sparse solver(s), but I realized that only after I started to reorganize the code

PR #454, opened 2022-04-16
Copy an empty vector from the end of another vector

nychiang:
Fix an incorrect assert conditions that occurs copying an empty vector from the end of another vector.
For example, the assert happens in the following sequence:
```
vec1.get_local_size()=0
vec2.get_local_size()=15

vec1.startingAtCopyFromStartingAt(0,vec2,15)
```

### DIFF
--- a/src/LinAlg/hiopVectorRajaPar.cpp
+++ b/src/LinAlg/hiopVectorRajaPar.cpp
@@ -452,9 +452,9 @@ void hiopVectorRajaPar::startingAtCopyFromStartingAt(
 #ifdef HIOP_DEEPCHECKS
   assert(n_local_ == n_ && "are you sure you want to call this?");
 #endif
-  assert(start_idx_dest >= 0 && start_idx_dest < this->n_local_);
+  assert(start_idx_dest >= 0 && start_idx_dest < this->n_local_ || this->n_local_==0);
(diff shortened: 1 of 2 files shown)


nychiang:
This problem can be reproduced in the current `develop` branch, when using BiCGStab IR on unconstrained problems.
For example, run`./src/Drivers/Sparse/NlpSparseEx3.exe`

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -452,9 +452,9 @@ void hiopVectorRajaPar::startingAtCopyFromStartingAt(
 #ifdef HIOP_DEEPCHECKS
   assert(n_local_ == n_ && "are you sure you want to call this?");
 #endif
-  assert(start_idx_dest >= 0 && start_idx_dest < this->n_local_);
+  assert(start_idx_dest >= 0 && start_idx_dest < this->n_local_ || this->n_local_==0);
   const hiopVectorRajaPar& v = dynamic_cast<const hiopVectorRajaPar&>(vec_src);
-  assert(start_idx_src >=0 && start_idx_src < v.n_local_);
+  assert(start_idx_src >=0 && start_idx_src < v.n_local_ || v.n_local_==0 || v.n_local_==start_idx_src);
```
cnpetra, line 457: please revisit this...

cnpetra (approved):

PR #456, opened 2022-04-16
Add back cusolver example

nychiang:
cusolver examples were removed by accident. 
This PR adds those two examples back to the CI tests.

@pelesh


pelesh:
Some tests still fail, but it looks unrelated to this PR. @ashermancinelli

pelesh (approved):
I suggest we merge this. It seems that CI on ascent is malfunctioning, but this is unrelated to this PR. Other CI pipelines pass.

ashermancinelli:
Yes, it does seem there was a fluke with the testing and this PR should be good to go.

cnpetra (approved):

PR #460, opened 2022-04-26
Install drivers for example 4

ashermancinelli:
Installing drivers will simplify testing procedure for collaborators and will simplify the process of adding tests to spack so users can test a hiop installation directly through the package manager.


ashermancinelli:
Related to the following spack pr: https://github.com/spack/spack/pull/30313

pelesh (approved):
It is a minor and straightforward change. Should be OK.

PR #463, opened 2022-05-04
Add support to sparse linear system in the formulation of normal equation

nychiang:
Add support to normal equation.
Right now we assume the input Hessian is diagonal.

Follow-up PRs will 
1. add unit tests for matrix function `is_diagonal` and `extract_diagonal`
2. add CI tests
3. optimize numerical efficiency for LP and QP


cnpetra (approved):
lgtm

PR #464, opened 2022-05-10
User options for condensed linear algebra

### LINKED ISSUE #425: Consider providing explicit options for using cuSOLVER-Cholesky
As far as I can see, there is no option to explicitly set HiOp to use cuSOLVER-Cholesky linear solver. It seems that HiOp will implicitly use cuSOLVER-Cholesky when model is set to `condensed`, CUDA is enabled, and linear solver is set to `auto`. Considering that Cholesky interface and condensed model are reasonably well tested and stable, it might be a good idea to provide user option to use Cholesky solver in HiOp.

@cnpetra @kswirydo

cnpetra:
Adds explicit user option for using Cholesky cuSolver when this is applicable. 

Closes #425

### DIFF
--- a/src/Drivers/Sparse/NlpSparseEx2Driver.cpp
+++ b/src/Drivers/Sparse/NlpSparseEx2Driver.cpp
@@ -21,7 +21,7 @@ static bool parse_arguments(int argc,
   n = 3;
   inertia_free = false;
   use_cusolver = false;
-  use_ginkgo = true;
+  use_ginkgo = false;
   switch(argc) {
   case 1:
     //no arguments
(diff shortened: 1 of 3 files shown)


src/Drivers/Sparse/NlpSparseEx2Driver.cpp
```diff
@@ -21,7 +21,7 @@ static bool parse_arguments(int argc,
   n = 3;
   inertia_free = false;
   use_cusolver = false;
-  use_ginkgo = true;
```
cnpetra, line 24: @pelesh : does `use_ginkgo` need to be true by default? It breaks some of existing use cases for sparse ex2.

pelesh: @ashermancinelli and I just came across this bug a moment ago. No, everything should be initialized to `false` and enabled only per user request. If user does not specify the linear solver, HiOp should set its default.

pelesh (approved):

PR #467, opened 2022-05-11
Update CI scripts to run Ginkgo tests

### LINKED ISSUE #466: Add Ginkgo tests to CI pipeline
Ginkgo library is not available on HiOp CI test pipelines. If Ginkgo is selected as linear solver, HiOp will notice it is not available in CI build and will default to available solver (MA57 in this case). This gives false positives for Ginkgo tests.

CC @cnpetra

pelesh:
Add Ginkgo libraries to test platforms and update HiOp build to enable Ginkgo support and run tests.

Closes #466 

CC @cnpetra @fritzgoebel


pelesh:
Sparse example 2 (formerly ex7) fails to converge. @fritzgoebel, were you able to run it to convergence on your machine?

pelesh:
The good news is that the convex example (Sparse ex1) passes with Ginkgo as the linear solver. It seems that CI pipeline is working correctly.

ashermancinelli (approved):
approving even though ci fails since the failing code is in develop anyways. imo we should either merge this or revert the ginkgo PR and create another PR which we merge after it's passing tests.

pelesh:
@cnpetra, I think this can be merged. Ginkgo example 2 fails, so I disabled it and I filed separate issue to address that. This PR is only about setting up CI pipelines to run Ginkgo tests. Ginkgo example 1 (convex problem) runs successfully, so we can verify Ginkgo solver runs and is configured correctly.

PR #470, opened 2022-05-12
User option for using (sym)AMD from EIGEN

### LINKED ISSUE #434: Enable use of Eigen as ordering package
Should elevate Eigen from experimental to supported since it offers the fastest and best symmetric orderings for Cholesky factorization.  

Currently the headers includes are hardcoded [here](https://github.com/LLNL/hiop/blob/342eb99ec16d45f57a492be1bf1e39cce73995a5/src/LinAlg/hiopLinSolverCholCuSparse.cpp#L68)

Will need to 
- [x] add `HIOP_USE_EIGEN` in the build system and remove the hardcoded includes
- [x] prepare git submodules for target platforms and CI
- [x] HiOp option for Eigen (new value under `linear_solver_sparse_ordering` option)

cnpetra:
closes #434

### DIFF
--- a/src/LinAlg/hiopLinSolverCholCuSparse.cpp
+++ b/src/LinAlg/hiopLinSolverCholCuSparse.cpp
@@ -165,7 +165,7 @@ bool hiopLinSolverCholCuSparse::do_symb_analysis(const size_type n,
     ret = cusolverSpXcsrsymamdHost(h_cusolver_, n, nnz, mat_descr_, rowptr, colind, perm);
     assert(ret == CUSOLVER_STATUS_SUCCESS);
   } else if("symamd-eigen" == ordering) {
-
+#ifdef HIOP_USE_EIGEN
     Eigen::Map<SparseMatrixCSR> M(n,
                                   n,
                                   nnz,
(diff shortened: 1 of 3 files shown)


src/LinAlg/hiopLinSolverCholCuSparse.cpp
```diff
@@ -161,9 +161,23 @@ bool hiopLinSolverCholCuSparse::do_symb_analysis(const size_type n,
...
+                                  nnz,
+                                  const_cast<int*>(rowptr),
+                                  const_cast<int*>(colind),
+                                  const_cast<double*>(value));
+
+    PermutationMatrix P;
+    Ordering ordering;
+    ordering(M.selfadjointView<Eigen::Upper>(), P);
+    memcpy(perm, P.indices().data(), n*sizeof(int));
+
```
nychiang, line 180: Do we need to have `#ifdef HIOP_USE_EIGEN` in this section? Can `Eigen` be undefined?

nychiang (approved):

src/LinAlg/hiopLinSolverCholCuSparse.cpp
cnpetra: indeed, it can be

PR #471, opened 2022-05-12
CSR `add_matrix` mods

cnpetra:
Supposed to provide CUDA implementation for `add_matrix_xxx`, but it also changes the signature of these methods in the base CSR class and modifies the CPU implementation as well as the Condensed Cholesky-based linear algebra.

Affects PR #463, and this PR will be merged after and take care of the changes in the NormalEquations PR.


nychiang (approved):
lgtm. Do we have any examples/drivers to test the cuda functions?

cnpetra:
> lgtm. Do we have any examples/drivers to test the cuda functions?

Not in place right now, but once all the kernels will be developed testing will be done via NlpSparseEx2 (second test) that runs condensed linear algebra. We will add unit tests shortly after.

PR #478, opened 2022-05-20
CUDA CSR diagonal kernels

cnpetra:
Provides a couple of kernels for updating diagonal of a CSR matrix

### DIFF
--- a/src/LinAlg/MatrixSparseCsrCudaKernels.cu
+++ b/src/LinAlg/MatrixSparseCsrCudaKernels.cu
@@ -63,6 +63,81 @@
 #include <device_launch_parameters.h>
 
 __global__
+void csr_add_vec_to_diag(int n, int nnz, int* irowptr, int* jcolind, double* values, double alpha, const double* diag_values)
+{
+  int tid = blockIdx.x * blockDim.x + threadIdx.x;
+  int stride = blockDim.x * gridDim.x;
+  for(int row=tid; row<n; row+=stride) {
+    int L=irowptr[row];
+    int R=irowptr[row+1]-1;
+    assert(L<nnz);
+    assert(R<nnz);
+    assert(L<=R);
+
+#ifndef NDEBUG
+    int idx_found = -1;
+#endif
+    do { //binary search
+      const int midpoint = (R+L)/2;
+      if(jcolind[midpoint]>row) {
+        R = midpoint-1;
+      } else if(jcolind[midpoint]<row) {
+        L = midpoint+1;
+      } else {
+        assert(idx_found<nnz);
+        values[midpoint] += alpha*diag_values[row];
+#ifndef NDEBUG
+        idx_found = midpoint;
+#endif        
+        break;
+      }
+    } while(L<=R);
+#ifndef NDEBUG
+    assert(idx_found>=0 &&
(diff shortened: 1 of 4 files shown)


src/LinAlg/MatrixSparseCsrCudaKernels.cu
```diff
@@ -63,7 +63,7 @@
 #include <device_launch_parameters.h>
 
 __global__
-void csr_set_diag_to_val(int n, int nnz, int* irowptr, int* jcolind, double* values, double val)
+void csr_add_vec_to_diag(int n, int nnz, int* irowptr, int* jcolind, double* values, double alpha, const double* diag_values)
```
nychiang, line 66: I think it's better to add some descriptions to each function in this file.

src/LinAlg/MatrixSparseCsrCudaKernels.hpp
```diff
@@ -62,9 +62,47 @@ namespace cuda
...
+ * for each row. Assumes pointers are on the device and parallelizes over rows.
+ * 
+ * @pre CSR matrix must be square.
+ * @pre Diagonal entries must appear explicitly among the nonzeros.
+ * @pre Column indexes must be sorted for any given row.
+ */   
+void csr_add_diag_kernel(int n, int nnz, int* irowptr, int* jcoldind, double* values, double Dval);
+
+/**
+ * Add entries of the array `values`  the diagonal of the CSR matrix. Performs a binary search on the column indexes
```
nychiang, line 83: Add entries of the array `values` "to" the diagonal of the CSR matrix.

src/LinAlg/hiopMatrixSparseCsrCuda.cpp
```diff
@@ -773,7 +779,7 @@ void hiopMatrixSparseCSRCUDA::form_transpose_from_numeric(const hiopMatrixSparse
 
 void hiopMatrixSparseCSRCUDA::form_diag_from_symbolic(const hiopVector& D)
 {
-  int m = D.get_size();
+  const int m = D.get_size();
```
nychiang, line 782: Is this function finished? seems that the sparsity patten hasn't been assigned yet.

nychiang (approved):
@cnpetra  by using the new functions "csr_add_diag_kernel" and "csr_set_diag_kernel", can we delete the diagonal matrix i[n this section?](https://github.com/LLNL/hiop/blob/54fba5c1d11fd2f5c56d46bdb039d2583fa6bca3/src/Optimization/hiopKKTLinSysSparseNormalEqn.cpp#L241-L257)

cnpetra:
> @cnpetra by using the new functions "csr_add_diag_kernel" and "csr_set_diag_kernel", can we delete the diagonal matrix i[n this section?](https://github.com/LLNL/hiop/blob/54fba5c1d11fd2f5c56d46bdb039d2583fa6bca3/src/Optimization/hiopKKTLinSysSparseNormalEqn.cpp#L241-L257)

Probably yes, but not sure which one is faster. I will follow up later on once I can benchmark the code a bit more.

src/LinAlg/hiopMatrixSparseCsrCuda.cpp
cnpetra: it is not, that code got lost somewhere somehow ...

nychiang (approved):

PR #484, opened 2022-05-31
C/Fortran interface

nychiang:
A C/Fortran interface for sparse linear system.


cnpetra (approved):

nychiang:
@cnpetra All tests on `newell` and `marianas` are passed, while `build_on_incline` is allowed to fail.

cnpetra:
> @cnpetra All tests on `newell` and `marianas` are passed, while `build_on_incline` is allowed to fail.

yeap, they passed on restart.

PR #485, opened 2022-06-01
Use HiOp options to configure cuSOLVER-LU

pelesh:
Added options for ordering, factorization and re-factorization. Removed accessors method for cuSOLVER configuration as everything should be managed through HiOp options.

### DIFF
--- a/src/LinAlg/hiopLinSolverSparseCUSOLVER.cpp
+++ b/src/LinAlg/hiopLinSolverSparseCUSOLVER.cpp
@@ -76,7 +76,8 @@ namespace hiop
       index_covert_extra_Diag2CSR_{ nullptr }, 
       n_{ n }, 
       nnz_{ 0 },
-      fact_{ "KLU" },   // default
+      ordering_{ 1 },
+      fact_{ "klu" },   // default
       refact_{ "glu" }, // default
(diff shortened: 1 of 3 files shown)


src/LinAlg/hiopLinSolverSparseCUSOLVER.cpp
```diff
@@ -236,15 +268,20 @@ namespace hiop
                                        d_work_);
       if(sp_status_ == 0) {
         checkCudaErrors(cudaMemcpy(dx, devx_, sizeof(double) * n_, cudaMemcpyDeviceToHost));
+      } else {
+        nlp_->log->printf(hovError,  // catastrophic failure
+                          "Solve failed with starus: %d\n",
```
kswirydo: status -- typo

src/LinAlg/hiopLinSolverSparseCUSOLVER.cpp
```diff
@@ -1052,7 +1105,7 @@ namespace hiop
             nlp_->log->printf(hovError,  // catastrophic failure
                               "Solve failed with starus: %d\n",
```
kswirydo: status -- typo

PR #486, opened 2022-06-01
Regularization normaleqn dev

nychiang:
Always update dual regularizations first if normal equation is used. 
Note that by default linear system `xdycyd`, we update primal first.

The current implementation has some numerical issue when randomized regularizations is used.
That is, the BICG iterative method and some other functions still use the unified regularization in computation. 
Will have a follow-up PR to fix this issue.


src/LinAlg/hiopVectorPar.cpp
```diff
@@ -124,6 +125,22 @@ void hiopVectorPar::setToConstant(double c)
...
+
+#ifdef NDEBUG
+  re.seed(std::chrono::system_clock::now().time_since_epoch().count());
+#endif
+
+  for(index_type i=0; i<n_local_; ++i) {
+    data_[i] = unif(re);
+  }
+}
+
```
nychiang, line 142: Any good way to implement this function in hiopVectorRAJAPar?
@cnpetra @ashermancinelli

ashermancinelli: cuRand is shipped in cuda toolkit, but we might have to add some linker flags.

ashermancinelli: [We could do almost exactly what this Thrust example is doing with RAJA](https://docs.nvidia.com/cuda/curand/device-api-overview.html#thrust-and-curand-example).

nychiang: @ashermancinelli I know we have cuRAND and rocRAND, but the problem now is that in this case I cannot see a portable solution to both cuda and hip. It seems to me that we need to use `#ifdef HIOP_USE_CUDA` with `cuRAND`, and `#ifdef HIOP_USE_HIP` with `rocRAND`.

ashermancinelli: I see now. Grepping through raja's docs doesn't find anything for random, so I think ugly macros are the only way to go.

cnpetra: not much to add -- it appears we need to rely on roc and cuda api (with `ifdef`s). We will add an abstraction layer for device math functions call later on.

src/LinAlg/hiopVectorPar.hpp
```diff
@@ -70,6 +70,7 @@ class hiopVectorPar : public hiopVector
 
   virtual void setToZero();
   virtual void setToConstant( double c );
+  virtual void set_to_random_constant(double minv, double maxv);
```
cnpetra, line 73: "constant" is a bit confusing maybe? name need to reflect that the uniform distribution is sampled. Did you mean `set_to_random_uniform` ?

src/LinAlg/hiopVectorRajaPar.hpp
```diff
@@ -77,6 +77,7 @@ class hiopVectorRajaPar : public hiopVector
 
   virtual void setToZero();
   virtual void setToConstant( double c );
+  virtual void set_to_random_constant(double minv, double maxv) {assert(false&&"not implemented yet!");}
```
cnpetra, line 80: we need this implementation rather sooner than later. A quick follow-up PR would be a good idea

cnpetra (approved):
will merge as-is ... name of vector's random function and the device implementation can be done via a follow-up PR

PR #487, opened 2022-06-03
C/F interface for dense linear algebra with quasi-Newton's method

nychiang:
C/F interface for dense linear algebra with quasi-Newton's method.


cnpetra (approved):

cnpetra:
tests will be added by a follow-up PR?

PR #494, opened 2022-06-14
Fix three corner cases

nychiang:
This PR fixes the following issues:
1. Fix the initial point for the FR problem when the given point is out of boundary.
2. When small slacks are detected, do not adjust variable bounds for free variables
3. change the lower bound of ex3 to reduce numerical issues.

### DIFF
--- a/src/LinAlg/hiopVectorRajaPar.cpp
+++ b/src/LinAlg/hiopVectorRajaPar.cpp
@@ -296,7 +296,7 @@ void hiopVectorRajaPar::copy_from(const hiopVector& vec, const hiopVectorInt& in
 }
 
 /// @brief Copy from vec the elements specified by the indices in index_in_src. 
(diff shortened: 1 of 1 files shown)


src/Optimization/hiopLogBarProblem.hpp
```diff
@@ -139,7 +139,7 @@ class hiopLogBarProblem
   {
     nlp->runStats.tmSolverInternal.start();
 
-    double retval_f_logbar_trial = f - mu * iter_trial.evalLogBarrier(xref);
+    double retval_f_logbar_trial = f - mu * iter_trial.evalLogBarrier(xref,mu);
```
cnpetra, line 142: space needed

cnpetra (changes requested):
looks good -- needs the small adjustments we discussed in private

cnpetra (approved):

PR #500, opened 2022-06-20
Scale rows implementation CSR Cuda matrix

cnpetra:


nychiang (approved):
It seems this is just a temporary code to test building condensed linear system on device. The follow up PRs may be more interesting.

PR #501, opened 2022-06-21
Use cuRand/hipRand to generate random numbers

nychiang:
Create random variables from cuRand or hipRand

### DIFF
--- a/src/LinAlg/hiopVectorRajaPar.cpp
+++ b/src/LinAlg/hiopVectorRajaPar.cpp
@@ -57,6 +57,9 @@
  */
 #include "hiopVectorRajaPar.hpp"
 #include "hiopVectorIntRaja.hpp"
+#ifdef HIOP_USE_CUDA
+#include "hiopUtilsCUDA.hpp"
+#endif
 
 #include <cmath>
 #include <cstring> //for memcpy
@@ -216,6 +219,16 @@ void hiopVectorRajaPar::setToConstant(double c)
     {
       data[i] = c;
     });
+}
+
+void hiopVectorRajaPar::set_to_random_uniform(double minv, double maxv)
+{
+  double* data = data_dev_;
+#ifdef HIOP_USE_CUDA
+  hiop::cuda::array_random_uniform_kernel(n_local_, data);
+#else
+  assert(0&&"not yet");
+#endif
 }
 
 /// Set selected elements to constant, zero otherwise
--- a/src/LinAlg/hiopVectorRajaPar.cpp
+++ b/src/LinAlg/hiopVectorRajaPar.cpp
@@ -225,7 +225,7 @@ void hiopVectorRajaPar::set_to_random_uniform(double minv, double maxv)
 {
   double* data = data_dev_;
 #ifdef HIOP_USE_CUDA
-  hiop::cuda::array_random_uniform_kernel(n_local_, data);
+  hiop::cuda::array_random_uniform_kernel(n_local_, data, minv, maxv);
 #else
   assert(0&&"not yet");
 #endif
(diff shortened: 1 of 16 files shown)


src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -218,6 +221,18 @@ void hiopVectorRajaPar::setToConstant(double c)
...
 
+void hiopVectorRajaPar::set_to_random_uniform(double minv, double maxv)
+{
+  double* data = data_dev_;
+#ifdef HIOP_USE_GPU
+  hiop::device::array_random_uniform_kernel(n_local_, data);
+  scale(maxv-minv);
+  addConstant(minv);
+#else
+  assert(0&&"not yet");
```
CameronRutherford, line 232: Should we just use c++ standard features to set the array to be random here?

CameronRutherford (approved):
Assuming the CPU RAJA kernel will be implemented in another PR, LGTM. I am not as familiar with HIP CMake best practices, but a cursory look didn't reveal anything.

src/LinAlg/hiopVectorRajaPar.cpp
nychiang: You are right. My bad. I should change this PR to draft mode since I would like to test hip implementation as well.

src/Utils/MathCudaKernels.cu
```diff
@@ -0,0 +1,115 @@
...
+  curandGenerator_t generator;
+  curandCreateGenerator(&generator, CURAND_RNG_PSEUDO_DEFAULT);
+  curandSetPseudoRandomGeneratorSeed(generator, seed);
+
+  // generate random val from 0 to 1
+  curandGenerateUniformDouble(generator, d_array, n);
+
+  curandDestroyGenerator(generator);
+  return 1;
+}
```
nychiang, line 111: Not sure which implementation is faster. The second method will adjust the values in the outer layer [(see here)](https://github.com/LLNL/hiop/blob/be9c984ed928b3a29f809f85f37fc9fabcfbf1b5/src/LinAlg/hiopVectorRajaPar.cpp#L230-L236)
@cnpetra @CameronRutherford any experience?

cnpetra: use the one with 4 arguments -- it "fuses" kernels

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -218,6 +225,29 @@ void hiopVectorRajaPar::setToConstant(double c)
...
+
+    // this 3-line method from cuda api also works
+//  hiop::device::array_random_uniform_kernel(n_local_, data);
+//  scale(maxv-minv);
+//  addConstant(minv);
+  }
+#else
+  // TODO: add function for openmp polocy
+  hiop::host::array_random_uniform_kernel(n_local_, data, minv, maxv);
+#endif
```
nychiang, line 248: Unit test fails when we use unified memory. Not sure why. move to offline discussion.

src/Utils/CMakeLists.txt
```diff
@@ -27,5 +30,23 @@ install(
...
+  set(hiopUtils_CUDA_SRC
+    MathCudaKernels.cu
+  )
+  set_source_files_properties(${hiopUtils_CUDA_SRC} PROPERTIES LANGUAGE CUDA)
+  list(APPEND hiopUtils_SRC ${hiopUtils_CUDA_SRC})
+elseif(HIOP_USE_HIP)
+  set(hiopUtils_HIP_SRC
+    MathHipKernels.cpp
+  )
+# Mark source files as HIP. I guess in the future just a
```
cnpetra, line 43: indentation, but I will fix myself when fixing expected merge conflicts for PR#510

cnpetra (approved):

src/LinAlg/hiopVectorRajaPar.cpp
cnpetra: for the record, lack of device synchronization was causing this.

PR #506, opened 2022-06-24
fix gingko integration

sayefsakin:
Added find package from ginkgo directory


CameronRutherford:
We found this necessary to build ExaGO with GInkgo, HiOp@develop and ExaGO@1.4.1.

CameronRutherford (approved):

PR #507, opened 2022-06-28
conversions between CSR and Triplet matrices

cnpetra:
Documents the specialized Triplet2CSR convert method and renames it accordingly to the codestyle guidelines


cnpetra:
@nychiang While I was documenting the existing method for converting to CSR, I observed the use of `std::unordered_map`. Can we use a `vector<int>` instead ?

nychiang (approved):
Have got an offline discussion with Cosmin for the future plan about the usage of CSR format.

PR #509, opened 2022-06-29
MA57 linear solver class with CSR input

cnpetra:
MA57 wrapper class is enriched to support CSR input matrices. With this development, the condensed KKT class works with generic linear solvers.


src/LinAlg/hiopLinSolverSymSparseMA57.hpp
```diff
@@ -156,14 +207,115 @@ class hiopLinSolverSymSparseMA57: public hiopLinSolverSymSparse
...
+  };
+
+/** 
+ * MA57 solver class that takes CSR sparse input and offers the boilerplate to copy this into the internal 
+ * triplet matrix used with MA57 API. 
+ * 
+ * The CSR matrix is understood to be symmetric. The underlying CSR storage can contain all the nonzero entries 
+ * or only the lower triangular part. In both cases, this class will copy ONLY the lower triangular entries to 
+ * the underlying triplet storage.
+ */
```
nychiang, line 228: Do we have a way/function to check if the input CSR matrix is `upper` triangular, which will fail all the methods below?

nychiang (approved):
I think this PR does what it claims and LGTM.
Honestly, I think it is better to have Triplet2CSR and CSR2Triplet converters outside of a particular linear solver class, in order to reuse them in other solver classes. I think CSR2Triplet may only be used when a HSL solver is given, and probably we can use CholMod as the default solver for condensed linear system instead of MA57.

cnpetra:
> them in other solver classes. I think CSR2Triplet may only be used when a HSL solver is given, and proba

I agree with CSR2triplet and Cholmod. This PR was an "preview" for when we have CSR KKT linear systems (if it's going to happen) instead of triplet.

PR #511, opened 2022-07-01
Fix compilation issues with latest RAJA/Umpire

### LINKED ISSUE #510: Consider keeping hiopNlpFormulation class hardware agnostic
In [`hiopNlpFormulation.cpp`](https://github.com/LLNL/hiop/blob/develop/src/Optimization/hiopNlpFormulation.cpp#L69) file a RAJA header is included, which pulls in vendor specific code into `hiopNlpFormulation` class. This is departure from HiOp design approach to keep optimization algorithms generic and implements vendor-specific code in linear algebra kernels.

This issue causes HiOp build failures with newer releases of RAJA. 

The obvious workaround is to compile `hiopNlpFormulation.cpp` with vendor specific compiler. Probably a better solution would be to abstract away the offending code in `hiopNlpFormulation` class.

CC @CameronRutherford

cnpetra:
closes #510

### DIFF
--- a/src/Utils/CMakeLists.txt
+++ b/src/Utils/CMakeLists.txt
@@ -23,7 +23,7 @@ set(hiopUtils_INTERFACE_HEADERS
 )
 
 if(HIOP_USE_RAJA)
-  list(APPEND hiopUtils_INTERFACE_HEADERS hiopRajaUmpireUtils.hpp)
+  list(APPEND hiopUtils_INTERFACE_HEADERS RajaUmpireUtils.hpp)
   list(APPEND hiopUtils_SRC ${hiopUtils_RAJA_SRC})
 endif()
 
--- a/src/Utils/CMakeLists.txt
+++ b/src/Utils/CMakeLists.txt
@@ -1,11 +1,15 @@
 set(hiopUtils_SRC
   hiopLogger.cpp
   hiopOptions.cpp
-  )
+)
+
+set(hiopUtils_RAJA_SRC
+  RajaUmpireUtils.cpp
+)
 
 set(hiopUtils_KRON_REDUCTION_SRC
   hiopKronReduction.cpp
-  )
+)
 
 set(hiopUtils_INTERFACE_HEADERS
   hiopCSR_IO.hpp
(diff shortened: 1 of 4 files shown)


pelesh (approved):
This addresses issue reported in #510. All enabled tests pass.

src/Utils/CMakeLists.txt
```diff
@@ -16,16 +20,27 @@ set(hiopUtils_INTERFACE_HEADERS
...
+  list(APPEND hiopUtils_SRC ${hiopUtils_RAJA_SRC})
+endif()
+
+if(HIOP_USE_CUDA)
+  set_source_files_properties(${hiopUtils_RAJA_SRC} PROPERTIES LANGUAGE CUDA)
+endif()
+if(HIOP_USE_HIP)
+  #this needs to be checked with hip/rocm
+  set_source_files_properties(${hiopUtils_RAJA_SRC} PROPERTIES LANGUAGE CXX)
+  set_source_files_properties(${hiopUtils_RAJA_SRC} PROPERTIES COMPILE_FLAGS "-x hip")
```
nychiang, line 36: This is the recommended way by the ORNL training for that particular platform.
It might be outdated since `cmake` may have already added support to `set_source_files_properties(xxxx PROPERTIES LANGUAGE HIP)

src/Utils/RajaUmpireUtils.hpp
```diff
@@ -56,42 +56,18 @@
...
-      auto found_allocator = rm.getAllocator(address);
-      std::stringstream ss;
-      ss << "Allocated on '" << found_allocator.getName() << "'   ";
-      ss << "Platform " << static_cast<int>(found_allocator.getPlatform()) << "   "; 
-      ss << "Size [" << found_allocator.getSize(address) << "]";
-      return ss.str();
-    } else {
-      return std::string("__not_managed_by_Umpire__");
-    }
+  std::string get_umpire_mem_address_info(void* address);
```
nychiang, line 70: Where is this function called in HiOp?

nychiang (approved):

src/Utils/RajaUmpireUtils.hpp
cnpetra: it's not, it's for debugging only, but I tested it manually.

src/Utils/CMakeLists.txt
cnpetra: I guess, both of us ran into https://www.olcf.ornl.gov/wp-content/uploads/2021/01/2021UM-Day-3-Joo-Developing-for-Frontier-using-HIP-on-Spock.pdf

nychiang: yes! Good to know that this doc is on public domain.

CameronRutherford (approved):

PR #514, opened 2022-07-07
Fix some compilation errors when AMD GPUs are used

### LINKED ISSUE #7: C/Fortran interface
A C/Fortran interface to this would be great. I think it could be added in a similar way that Ipopt has done it.

nychiang:
Flang is required when compiling HiOp with Hip toolkit, since ROCm bitcode library is only accessible by Clang or Flang.


cnpetra:
LGTM. I removed the (close) tag of Issue #7 since the issue would benefit from additional clarification prior to closing.

cnpetra (approved):

PR #518, opened 2022-07-12
Fix bugs in CUDA kernels for FGMRES

pelesh:
- Fixed a couple of trivial bugs in CUDA kernels
- Changed code style at several places to match HiOp guidelines


pelesh:
@kswirydo, I would suggest dropping support for nonsymmetric case for now.

kswirydo (approved):

src/LinAlg/hiopLinSolverSparseCUSOLVER.cpp
```diff
@@ -1582,22 +718,21 @@ namespace hiop
 
     /* setup GLU */
     sp_status_ = cusolverSpDgluSetup(handle_cusolver_, 
-                                     n_, 
-                                     nnz_, 
+                                     n_, nnz_,
```
kswirydo, line 721: This does not look right !!!

kswirydo (commented):
All changes are ok, except in one place you have two variables in function call in one line, this is not allowed by hiop standards/.

PR #508, opened 2022-06-28
Update Marianas variables and disable Newell CI, upgrade to C++14 and CUDA 14

CameronRutherford:
Updating the Marainas build with cuda 11.4. One of the unit tests fails.

<summary>
NlpSparse2_4 failure:
<details>
    Start 25: NlpSparse2_4

25: Test command: /share/apps/openmpi/3.1.3/gcc/7.3.0/bin/mpirun "-n" "1" "/qfs/people/ruth521/projects/hiop/hiop-git/build/src/Drivers/Sparse/NlpSparseEx2.exe" "500" "-ginkgo" "-inertiafree" "-selfcheck"
25: Test timeout computed to be: 10000000
25: ===============
25: Hiop SOLVER
25: ===============
25: Using 1 MPI ranks.
25: ---------------
25: Problem Summary
25: ---------------
25: Total number of variables: 500
25:      lower/upper/lower_and_upper bounds: 499 / 1 / 1
25: Total number of equality constraints: 2
25: Total number of inequality constraints: 499
25:      lower/upper/lower_and_upper bounds: 498 / 498 / 497
25: iter    objective     inf_pr     inf_du   lg(mu)  alpha_du   alpha_pr linesrch
25:    0  6.4379656e+01 9.980e+00  1.010e+00   0.00  0.000e+00  0.000e+00  -(-)
25: Setting up Ginkgo solver ...
25: cudaErrorNoKernelImageForDevice
25: no kernel image is available for execution on the device
25: [Warning] solve compressed high absolute resid norm (= 7.26408e+04)
25: [Warning] solve compressed high absolute resid norm (= 1.25761e+09)
25: [Warning] solve compressed high absolute resid norm (= 1.01166e+09)
25: [Warning] solve compressed high absolute resid norm (= 8.51145e+08)
25: [Warning] solve compressed high absolute resid norm (= 4.84826e+07)
25: [Warning] solve compressed high absolute resid norm (= 1.78755e+07)
25: [Warning] solve compressed high absolute resid norm (= 1.40315e+08)
25: [Warning] solve compressed high absolute resid norm (= 7.80505e+09)
25: [Warning] solve compressed high absolute resid norm (= 8.62754e+08)
25: [Warning] solve compressed high absolute resid norm (= 5.77189e+08)
25: [Warning] solve compressed high absolute resid norm (= 1.08956e+09)
25: [Warning] solve compressed high absolute resid norm (= 1.34512e+09)
[...]

PR #522, opened 2022-07-27
Update newell variables to cuda11.4

jaelynlitz:
These updated modules now use cuda 11.4 on our upgraded newell node (now centos8 from rhel7) newell02

### DIFF
--- a/scripts/newellVariables.sh
+++ b/scripts/newellVariables.sh
@@ -5,46 +5,80 @@ module use -a /share/apps/modules/tools
 module use -a /share/apps/modules/compilers
 module use -a /share/apps/modules/mpi
 module use -a /etc/modulefiles
-module use -a /qfs/projects/exasgd/src/cameron-spack/share/spack/modules/linux-rhel7-power9le
+module use -a /qfs/projects/exasgd/src/jaelyn-spack/spack/share/spack/modules/linux-centos8-power9le
 
 # Load spack-built modules
-# blt@0.4.1%gcc@7.4.0 arch=linux-rhel7-power9le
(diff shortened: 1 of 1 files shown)


CameronRutherford:
This uses a modification to the RAJA spack package that isn't in spack develop yet. Can you make a spack PR to RAJA and HiOp that includes that change, along with the new version constraint for RAJA and Umpire when using HiOp@develop?

I think once this is merged, and once we get the newer RAJA and Umpire versions installed on Marianas, we can re-enable the GitHub action to test at PNNL.

I have also been talking to E4S folks about the spack pipeline stage issues that we were seeing, and it seems like everything should be working again. @cnpetra can we test the spack build pipeline in this PR?

CameronRutherford:
> I think once this is merged, and once we get the newer RAJA and Umpire versions installed on Marianas, we can re-enable the GitHub action to test at PNNL.

Perhaps we should just re-enable the PNNL testing action in this PR as well, and just push all updates here. I remember seeing ginkgo errors in #508, and so merging broken changes would probably be ill-advised

CameronRutherford:
@pelesh with this PR, pipelines will be passing on Newell and Marianas if the GitHub action is re-enabled. I would also suggest trying to re-enable the spack pipeline, as that should be fully functional.

CameronRutherford:
I would love to re-enable the GitHub action for PNNL on this CI, and the spack action, and get this merged. cc @pelesh @cnpetra. We can modify to allow PNNL jobs to fail for the moment if there are errors in the pipeline, but we really need the action enabled in order to start getting things back u[ and running.

CameronRutherford:
Will be closing in favor of new PR updating variables in #521

PR #525, opened 2022-08-03
Cmake fortran example fix

nychiang:
Update the cmake file, to check if we have flang when 1) hip is used on AMD machine; and 2) user would like to build fortran examples.


CameronRutherford:
It seems like this should be added as an option to HiOp's spack package - would that be reasonable? I don't know how to implement this sort of logic (requiring flang for +fortran), but I would hope that is doable.

This CMake logic seems much more robust though. Thank you!

CameronRutherford (approved):

cnpetra (approved):

cnpetra:
lgtm, will merge once the conflict is fixed

nychiang:
@cnpetra no conflicts anymore.

PR #526, opened 2022-08-04
Fix PNNL CI

CameronRutherford:
PR to test out newly enabled pipelines and to push environment updates.

PR #527, opened 2022-08-04
use batch node

nychiang:
Use batch node instead of debug node.


CameronRutherford:
I am working on fixing GitLab CI in #526. I think you can ignore that failing test for now. I am also going to file a spack issue for the spack action that I have disabled. If it triggers on a PR you can just ignore it for the moment

cnpetra (approved):

PR #529, opened 2022-08-17
temporary allow failure on newell

cnpetra:


nychiang (approved):

CameronRutherford:
LGTM. The pipeline that was triggered by me has Newell tests running and failing. To be clear this is a temporary measure, and this should be re-enforced once we have everything installed successfully.

ORNL CI for Ascent is passing, and so that will suffice as a replacement for now.

PR #530, opened 2022-08-20
Refine the computation in normal equation system

nychiang:
1. Re-engineer the code to use object `hiopPDPerturbation` in `build_kkt_matrix`, to have a better OO design.
2. Remove unnecessary copies of `delta(s)`
3. if normal equation is used and primal regularization is unchanged within a step, skip building the expensive condensed matrix `J^T(H+Diag)^{-1}J` for the reason of efficiency.


cnpetra (approved):

cnpetra:
@nychiang can you fix the conflict? I will merge this PR shortly after

nychiang:
> @nychiang can you fix the conflict? I will merge this PR shortly after
all set

PR #531, opened 2022-08-23
Fix CI testing on Newell

jaelynlitz:
Rebuild the spack modules on Newell to support our OS upgrade and move to testing w/ CUDA 11 and ginkgo@glu_experimental


jaelynlitz:
Should I not allow Newell to fail anymore in this PR? Or leave as is while we're still sorting out the Newell upgrade?

cnpetra:
lgtm, will merge as soon as @CameronRutherford  approves it

CameronRutherford (approved):
LGTM. Glad Newell is back online!!

PR #534, opened 2022-08-24
Callback function that returns host array

nychiang:
umm...


nychiang:
use #535 instead of this one

PR #538, opened 2022-08-31
Update document

nychiang:
Update user manual for the latest changes in interface.


cnpetra (approved):
please copy the user manual pdf from src/ to doc/hiop_usermanual.pdf and commit the pdf. Also it would be nice to update the date.

PR #539, opened 2022-08-31
fixing small numerical issues in CUSOLVER iterative refinement

kswirydo:
What changed:
-- Gram-Schmidt function inside hiopSolverLinSolverCUSOLVER class have been updated.
-- Now one case use any restart value (even very large) with FGMRES IR with low synch Gram-Schmidt
-- Fixed two small numerical issues for the low-synch variants of Gram-Schmidt. They do not influence stability unless the space is extremely large.
-- Cleaned some repeated comments.
-- In "KrylovSolverKernels" increased the constant that regulates the size of allocated shared memory (to make sure it works for larger Krylov spaces).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -132,11 +132,6 @@
       done
       echo "finished batch job: $res"
       exit $res
-  after_script:
-    - |
-      set -xv
-      export WORKDIR="$HOME/gitlab/$CI_JOB_ID/"
-      rm -rf "$WORKDIR"
 
 .pnnl_tags_template: &pnnl_tags_definition
   tags:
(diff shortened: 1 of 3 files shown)


CameronRutherford:
I pushed a quick fix to fix the PNNL mirroring (I should have gotten around to that much sooner, but better late than never...)

CameronRutherford (approved):
Not sure I fully understand the numerical fixes applied here, but LGTM - tests are passing at PNNL and are reported correctly now

pelesh (approved):
Meaningful improvements to already well functioning solver. PNNL pipelines are fixed and reporting correctly. This is ready to merge.

PR #444, opened 2022-04-01
A temporary solution to the magma-hip bug

nychiang:
There is a strange bug which prevent magma returning the correct solution from `magma_dsytrs_gpu`, when HIP is used.
See https://bitbucket.org/icl/magma/issues/57/magma_dsytrs_gpu-returns-wrong-solution-on for more details about this issue. A temporary solution suggested by Dr. Stanimire Tomov is to use `MagmaUpper` when calling the corresponding magma routines.

This PR addresses this issue, and uses MC47 as the default ordering for MA57.

@pelesh @ashermancinelli @cnpetra

PR #541, opened 2022-09-01
Fix the bug that magma returns wrong inertia

nychiang:
We also need to use `MagmaUpper` before calling `magmablas_dsiinertia`, to return the correct inertia on hipblas.
See PR #444 for more detail about this change.

### DIFF
--- a/src/LinAlg/hiopLinSolverSymDenseMagma.cpp
+++ b/src/LinAlg/hiopLinSolverSymDenseMagma.cpp
@@ -144,7 +144,9 @@ namespace hiop
     int info;
     int retcode;
     magma_uplo_t uplo=MagmaLower; // M is upper in C++ so it's lower in fortran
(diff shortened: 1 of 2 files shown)


cnpetra:
> 

I think the initial PR was tested against OPF without line constraints. The issue is not a corner case. It passed all of our "corner" cases, which were devised to test this functionality. Only solving real word problems (and many of them) can catch this sort of issue.

cnpetra (approved):
@nychiang hope you had fun fishing, this is a great catch.

nychiang:
@CameronRutherford Same comments as Cosmin. We used to test the problems without line constraints and hence we didn't catch this problem. Honestly, I think this is a corner cases in Magma and the root cause comes from hipblas.

CameronRutherford (approved):

PR #543, opened 2022-09-07
Execution spaces: support for memory backends and execution policies

cnpetra:
Abstraction needed to support multiple hardware backends. The goal is to allow multiple "backends" to run concurrently, e.g., Cuda/Hip linear algebra with Raja linear algebra. This proposed implementation differentiates between memory backend and execution policy to allow using Raja with Umpire memory, Raja with Cuda/Hip memory, or Raja with std memory.

Passes tests on Crusher (with Umpire memory backend and RAJA exec policies), in addition to all the CIs. Tested with CUDA memory backend and RAJA exec policies manually on Ascent.


src/Drivers/PriDec/NlpPriDecEx2SparseRaja.cpp
```diff
@@ -4,8 +4,24 @@
...
+using ex9_raja_exec = hiop::ExecRajaPoliciesBackend<hiop::ExecPolicyRajaCuda>::hiop_raja_exec;
+using ex9_raja_reduce = hiop::ExecRajaPoliciesBackend<hiop::ExecPolicyRajaCuda>::hiop_raja_reduce;
+#endif
+
+#ifdef HIOP_USE_HIP
+#include <ExecPoliciesRajaHipImpl.hpp>
+using ex9_raja_exec = hiop::ExecRajaPoliciesBackend<hiop::ExecPolicyRajaHip>::hiop_raja_exec;
+using ex9_raja_reduce = hiop::ExecRajaPoliciesBackend<hiop::ExecPolicyRajaHip>::hiop_raja_reduce;
+#endif
+
```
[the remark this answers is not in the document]
CameronRutherford: Noting that we require raja to be built with at least one backend (OpenMP, CUDA of HIP), and we don't support RAJA with sequential execution policies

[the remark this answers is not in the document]
CameronRutherford: I would consider re-structuring these preprocessor macros to use if, elsif and else as it might be cleaner.

src/Utils/hiopOptions.cpp
```diff
@@ -1152,6 +1144,51 @@ void hiopOptionsNLP::register_options()
...
+    // auto - solver decides based on its capabilities and values of mem_space and mem_backend
+    // seq  - cpu sequential (compatible only with HOST mem_space and stdcpp mem_backend
+    // raja - RAJA backend
+    //          - raja-openmp execution; requires mem_space = 'HOST'   mem_backend = 'stdcpp' or 'umpire'
+    //          - raja-cuda execution;   requires mem_space = 'DEVICE' mem_backend = 'cuda' or 'umpire'
+    //          - raja-hip execution;    requires mem_space = 'DEVICE' mem_backend = 'hip' or 'umpire'
+    // cuda - only cuda kernels;         requires mem_space = 'DEVICE' mem_backend = 'cuda' or 'umpire'
+    // hip  - only hip kernels;          requires mem_space = 'DEVICE' mem_backend = 'hip' or 'umpire'
+
+    vector<string> range = {"auto", "seq", "raja", "cuda", "hip"};
```
CameronRutherford, line 1185: Consider using pre-processor guards to only compile in options that are supported.

src/LinAlg/hiopVectorRajaPar.hpp
```diff
@@ -54,9 +54,32 @@
...
+{
+#ifdef HIOP_USE_CUDA
+  typedef hiopVectorRaja<MemBackendUmpire, ExecPolicyRajaCuda> hiopVectorRajaPar;
+#endif
+
+#ifdef HIOP_USE_HIP
+  typedef hiopVectorRaja<MemBackendUmpire, ExecPolicyRajaHip> hiopVectorRajaPar;
+#endif
+
+//todo: we need a macro for openmp raja
```
CameronRutherford, line 74: If a user compiles with RAJA~cuda~hip~OpenMP, then they could shoot themselves in the foot. I think we should support sequential policy out of the box, but from discussion perhaps this is fine

pelesh (commented):
This looks very good. Still some work to do, but the overall direction is sound.

src/LinAlg/hiopVectorRajaPar.hpp
pelesh: Next version of RAJA will fix the issue when HIP and OpenMP backends are enabled at the same time, so this issue will become moot. However, it is a good idea to have option to build RAJA with sequential policy, as well.

CC @davidbeckingsale

src/Utils/hiopOptions.cpp
cnpetra: done

src/LinAlg/hiopVectorRajaPar.hpp
cnpetra: I think I understand what you guys are alluding to. But RAJA with sequential policy does not make sense. You mean with OpenMP policy?

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -0,0 +1,2233 @@
...
+      dd[i] = vd[id[i]];
+    });
+}
+
+/// @brief Copy from vec the elements specified by the indices in index_in_src
+template<class MEM, class POL> 
+void hiopVectorRaja<MEM,POL>::copy_from_indexes(const hiopVector& vv, const hiopVectorInt& index_in_src)
+{
+  const hiopVectorIntRaja& indexes = dynamic_cast<const hiopVectorIntRaja&>(index_in_src);
+  const hiopVectorRaja& v = dynamic_cast<const hiopVectorRaja&>(vv);
```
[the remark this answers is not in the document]
cnpetra: I missed this, thanks for pointing it out.

src/LinAlg/hiopVectorRajaPar.hpp
CameronRutherford: RAJA sequential policy makes sense for users that are experimenting, without needing to compile OpenMP or another RAJA backend.

That being said, it would make sense for users just to use "plain" HiOp without RAJA in that case, as you shouldn't see any difference between CPU based HiOp, and RAJA with sequential policies.

src/Drivers/PriDec/NlpPriDecEx2SparseRaja.cpp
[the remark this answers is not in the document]
cnpetra: done

src/LinAlg/hiopVectorRajaPar.hpp
pelesh: @cnpetra, using RAJA without OpenMP was a temporary workaround because of an issue on AMD architectures. There is no need to build RAJA without OpenMP backend anymore afaik.

src/LinAlg/hiopLinAlgFactory.cpp
```diff
@@ -85,22 +89,45 @@ using namespace hiop;
...
+#ifdef HIOP_USE_CUDA
+        return new hiop::hiopVectorRaja<hiop::MemBackendUmpire,hiop::ExecPolicyRajaCuda>(glob_n, mem_space_upper, col_part, comm);
+#endif        
+#ifdef HIOP_USE_HIP
+        return new hiop::hiopVectorRaja<hiop::MemBackendUmpire,hiop::ExecPolicyRajaHip>(glob_n, mem_space_upper, col_part, comm);
+#endif        
+#if !defined(HIOP_USE_CUDA) && !defined(HIOP_USE_HIP)
+        return new hiop::hiopVectorRaja<hiop::MemBackendUmpire,hiop::ExecPolicyRajaOmp>(glob_n, mem_space_upper, col_part, comm);
+#endif
+
```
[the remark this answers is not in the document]
cnpetra: Indeed, if `LinAlgFactory::create_vector` is going to be a template. However, doing so will still require the calling code to process the runtime options to decide the vector's template parameters. Or am I missing something?

src/LinAlg/hiopVectorRajaPar.hpp
cnpetra: ok, good. Commit  https://github.com/LLNL/hiop/pull/543/commits/59466953fd65b5e233da789eabfcf7d9a5e355c7 removes `hiopVectorRajaPar`. Propagation of new template class went smoothly. But I think it is instructive to take a look at the commit's diffs, especially in the tests  since it points to some additional work that may be needed.

cnpetra: > ok, good. Commit [5946695](https://github.com/LLNL/hiop/commit/59466953fd65b5e233da789eabfcf7d9a5e355c7) removes `hiopVectorRajaPar`. Propagation of new template class went smoothly. But I think it is instructive to take a look at the commit's diffs, especially in the tests since it points to some additional work that may be needed.

the commit also takes out the faulty preprocessor logic...

src/ExecBackends/ExecSpace.hpp
```diff
@@ -0,0 +1,429 @@
...
+};
+#endif //HIOP_USE_RAJA //HIOP_USE_UMPIRE
+
+#ifdef HIOP_USE_CUDA
+/// Cuda memory backend for device memory space that is implemented using Cuda API
+struct MemBackendCuda
+{
+  /**
+   * Constructor taking a memory space as input; provided for exchangeability with
+   * other memory backends.
```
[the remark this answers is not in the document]
cnpetra: we could, but would like this header to contain all the memory backends and policy types available.

src/LinAlg/hiopLinAlgFactory.cpp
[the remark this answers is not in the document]
pelesh: You are right, in that case the calling function (or function above in the hierarchy) would need to decide the template parameter.

src/ExecBackends/ExecPoliciesRajaHipImpl.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+namespace hiop
+{
+
+#define RAJA_LAMBDA [=] __device__
+  
+template<>
+struct ExecRajaPoliciesBackend<ExecPolicyRajaHip>
+{
+  static constexpr unsigned short int HIOP_RAJA_GPU_BLOCK_SIZE = 128;
```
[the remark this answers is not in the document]
cnpetra: correct. hip and cuda cannot sit in the same compilation unit, for example.

src/ExecBackends/MemBackendUmpireImpl.hpp
```diff
@@ -0,0 +1,215 @@
...
+ * This file contains Umpire implementation memory backend.
+ */
+
+#ifndef HIOP_MEM_BCK_UMPIRE
+#define HIOP_MEM_BCK_UMPIRE
+
+#include <ExecSpace.hpp>
+
+#ifdef HIOP_USE_RAJA // can/should be HIOP_USE_UMPIRE
+
```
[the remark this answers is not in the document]
cnpetra: yes, the idea is to be able to drop Umpire and reduce dependencies down the road.

src/LinAlg/hiopVectorRajaImpl.hpp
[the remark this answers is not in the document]
cnpetra: take a another look and see if I addressed this correctly.

src/ExecBackends/ExecSpace.hpp
```diff
@@ -0,0 +1,429 @@
...
+  }
+
+  const EXECPOLICIES& exec_policies() const
+  {
+    return ep_;
+  }
+  
+  template<typename T>
+  inline T* alloc_array(const size_t& n)
+  {
```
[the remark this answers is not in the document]
cnpetra: done

src/ExecBackends/ExecSpace.hpp
```diff
@@ -0,0 +1,450 @@
...
+
+#ifdef HIOP_USE_HIP
+struct ExecPolicyRajaHip
+{
+  //empty since no runtime info is stored
+};
+#endif
+
+//RAJA OMP execution policies backend
+#if !defined(HIOP_USE_CUDA) && !defined(HIOP_USE_HIP)
```
CameronRutherford, line 281:
```suggestion
#ifdef HIOP_USE_CUDA
struct ExecPolicyRajaCuda
{
  //empty since no runtime info is stored
};
#elifdef HIOP_USE_HIP
struct ExecPolicyRajaHip
{
  //empty since no runtime info is stored
};
#else
```
`#elifdef` is C++23, so please don't apply this, but I was curious if the syntax could be cleaner.

src/ExecBackends/ExecSpace.hpp
```diff
@@ -0,0 +1,450 @@
...
+};
+
+/**
+ * Transfers between memory backends and memory spaces should be provided by specializations
+ * of `TransferImpl` class.
+ */
+template<class MEMDEST, class EXEPOLDEST, class MEMSRC, class EXEPOLSRC, typename T, typename I>
+struct TransferImpl
+{
+  inline static bool do_it(T* p_dest,
```
CameronRutherford, line 438: Maybe I am missing some conetxt, but `do_it` seems like a vague function name imo

src/ExecBackends/ExecSpace.hpp
```diff
@@ -0,0 +1,450 @@
...
+                           const ExecSpace<MEMSRC, EXEPOLSRC>& hwb_src,
+                           const I& n)
+  {
+    return false;
+  }
+};
+
+} // end namespace
+
+#endif
```
CameronRutherford, line 450: I agree with @pelesh that this should maybe be broken down into separate compilation units for each backend.

I also think that there should perhaps be a `.cpp` file to go along with this one, but that is just prefernce.

src/ExecBackends/MemBackendUmpireImpl.hpp
```diff
@@ -0,0 +1,261 @@
...
+                           ExecSpace<MemBackendCpp, EXECPOLDEST>& hwb_dest,
+                           const T* p_src,
+                           const ExecSpace<MemBackendUmpire, MemBackendUmpire>& hwb_src,
+                           const I& n)
+  {
+    if(hwb_src.mem_backend().is_host()) {
+      std::memcpy(p_dest, p_src, n*sizeof(T));
+    } else {
+      assert(false && "Transfer BACKENDS(TO:Cpp-host,FROM:umpire) only supported with Umpire mem space host"); 
+      return false;
```
CameronRutherford, line 135: I know it would invoke a memory copy, but couldn't we support this with `mem_space=DEVICE` by copying from the device first?

src/ExecBackends/MemBackendUmpireImpl.hpp
```diff
@@ -0,0 +1,261 @@
...
+  inline static bool do_it(T* p_dest,
+                           ExecSpace<MemBackendUmpire, EXECPOLDEST>& hwb_dest,
+                           const T* p_src,
+                           const ExecSpace<MemBackendCpp, EXECPOLSRC>& hwb_src,
+                           const I& n)
+  {
+    if(hwb_dest.mem_backend().is_host()) {
+      std::memcpy(p_dest, p_src, n*sizeof(T));
+    } else {
+      assert(false && "Transfer BACKENDS(TO:Umpire,FROM:Cpp-host) only supported with Umpire mem space host");
```
CameronRutherford, line 153: Same here, although it would be a host -> device transfer instead.

src/ExecBackends/MemBackendUmpireImpl.hpp
```diff
@@ -0,0 +1,261 @@
...
+                           const I& n)
+  {
+    if(hwb_src.mem_backend().is_device()) {
+      return cudaSuccess == cudaMemcpy(p_dest, p_src, n*sizeof(T), cudaMemcpyDeviceToDevice);
+    } else {
+      if(hwb_src.mem_backend().is_host()) {
+        return cudaSuccess == cudaMemcpy(p_dest, p_src, n*sizeof(T), cudaMemcpyHostToDevice);
+      } else {
+        assert(false && "Transfer BACKENDS(TO:Cuda,FROM:umpire) not supported with Umpire mem space 'um'");
+        return false;
```
CameronRutherford, line 180: Might be missing something here... I would expect that this can be supported, but on the other hand, `um` memory is automatically transferred between device/host as needed, so users would be doing something quite weird to invoke this part of HiOp :)

src/ExecBackends/MemBackendUmpireImpl.hpp
```diff
@@ -0,0 +1,261 @@
...
+                           const ExecSpace<MemBackendHip, EXECPOLSRC>& hwb_src,
+                           const I& n)
+  {
+    if(hwb_dest.mem_backend().is_device()) {
+      return hipSuccess == hipMemcpy(p_dest, p_src, n*sizeof(T), hipMemcpyDeviceToDevice);
+    } else {
+      if(hwb_dest.mem_backend().is_host()) {
+        return hipSuccess == hipMemcpy(p_dest, p_src, n*sizeof(T), hipMemcpyDeviceToHost);
+      } else {
+        assert(false && "Transfer BACKENDS(TO:Umpire,FROM:Hip) not supported with Umpire mem space 'um'");
```
CameronRutherford, line 250: I suppose we haven't needed it, but cool that HIP also supports UM

src/LinAlg/hiopMatrixSparseCsrCuda.cpp
```diff
@@ -902,9 +919,11 @@ void hiopMatrixSparseCSRCUDA::form_from_symbolic(const hiopMatrixSparseTriplet&
 
   //transfer coo/triplet to device
   int* d_rowind=nullptr;
-  cudaMalloc(&d_rowind, nnz_*sizeof(index_type));
+  //cudaMalloc(&d_rowind, nnz_*sizeof(index_type));
+  d_rowind = exec_space_.alloc_array<index_type>(nnz_);
   assert(d_rowind);
-  cudaMemcpy(d_rowind, M.i_row(), nnz_*sizeof(index_type), cudaMemcpyHostToDevice);
+  //cudaMemcpy(d_rowind, M.i_row(), nnz_*sizeof(index_type), cudaMemcpyHostToDevice);
```
CameronRutherford, line 925: Should these commented out `cudaMemcpy` calls just be deleted?

src/LinAlg/hiopVectorRaja.hpp
```diff
@@ -275,7 +279,8 @@ class hiopVectorRajaPar : public hiopVector
   virtual bool isinf_local() const;
   virtual bool isfinite_local() const;
   
-  virtual void print(FILE* file=nullptr, const char* message=nullptr,int max_elems=-1, int rank=-1) const;
+  virtual void print(FILE*, const char* withMessage=NULL, int max_elems=-1, int rank=-1) const;
```
CameronRutherford, line 282: I assume the default values are initialized in the `.cpp` file instead...

src/LinAlg/hiopVectorRaja.hpp
```diff
@@ -306,18 +309,28 @@ class hiopVectorRajaPar : public hiopVector
   virtual bool is_equal(const hiopVector& vec) const;
   
 private:
+  ExecSpace<MEMBACKEND, EXECPOLICYRAJA> exec_space_;
+  using MEMBACKENDHOST = typename MEMBACKEND::MemBackendHost;
+
+  //EXECPOLICYRAJA is used internally as a execution policy. EXECPOLICYHOST is not used internally
+  //in this class. EXECPOLICYHOST can be any host policy as long as memory allocations and
+  //and transfers within and from `exec_space_host_` work with EXECPOLICYHOST (currently all such
+  //combinations work).
```
CameronRutherford, line 318: This comment is very helpful, and makes me wonder if we should be focusing more on HiOp documentation/user manual in order to communicate these sorts of things. I admit I am not as up to date on the sate of HiOp docs as others however.

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -130,28 +132,27 @@ hiopVectorRajaPar::hiopVectorRajaPar(
   n_local_ = glob_iu_ - glob_il_;
 
 #ifndef HIOP_USE_GPU
-  mem_space_ = "HOST"; // If no GPU support, fall back to host!
+  assert(mem_space_ == "HOST");
```
CameronRutherford, line 135: #577 applies here - perhaps in the meantime we should give an error message with the assert.

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -130,28 +132,27 @@ hiopVectorRajaPar::hiopVectorRajaPar(
...
 #ifndef HIOP_USE_GPU
-  mem_space_ = "HOST"; // If no GPU support, fall back to host!
+  assert(mem_space_ == "HOST"); 
 #endif
 
-  auto& resmgr = umpire::ResourceManager::getInstance();
-  umpire::Allocator devalloc  = resmgr.getAllocator(mem_space_);
-  data_dev_ = static_cast<double*>(devalloc.allocate(n_local_*sizeof(double)));
-  if(mem_space_ == "DEVICE")
+  data_dev_ = exec_space_.template alloc_array<double>(n_local_);
```
CameronRutherford, line 138: I'm getting very thrown off by this syntax, and don't know what to google to learn more. Is this an artifact of the templatized execution space class that we are using now?

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -161,48 +162,45 @@ hiopVectorRajaPar::hiopVectorRajaPar(const hiopVectorRajaPar& v)
   mem_space_ = v.mem_space_;
 
 #ifndef HIOP_USE_GPU
-  mem_space_ = "HOST"; // If no GPU support, fall back to host!
+  assert(mem_space_ == "HOST");
```
CameronRutherford, line 165: Same here w.r.t. a helpful error message.

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -161,48 +162,45 @@ hiopVectorRajaPar::hiopVectorRajaPar(const hiopVectorRajaPar& v)
...
   delete idx_cumsum_;
 }
 
-hiopVector* hiopVectorRajaPar::alloc_clone() const
+template<class MEM, class POL>
+hiopVector* hiopVectorRaja<MEM,POL>::alloc_clone() const
 {
-  hiopVector* v = new hiopVectorRajaPar(*this); assert(v);
+  hiopVector* v = new hiopVectorRaja<MEM,POL>(*this);
+  assert(v);
```
CameronRutherford, line 196: Is this necessary? I would hope that `hiopVectorRaja`'s constructor would handle error checking

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -413,20 +403,21 @@ void hiopVectorRajaPar::copy_from_indexes(const double* vv, const hiopVectorInt&
...
 
   // If nothing to copy, return.  
   if(nv == 0)
     return;
 
-  auto& rm = umpire::ResourceManager::getInstance();
-  double* vv = const_cast<double*>(v); // <- cast away const
-  rm.copy(data_dev_ + start_index_in_this, vv, nv*sizeof(double));
+  //TODO: data_dev_+start_index_in_this   -> is not portable, may not work on the device. RAJA loop should be used
+  exec_space_.copy(data_dev_+start_index_in_this, v, nv);
```
CameronRutherford, line 420: This function hasn't been an issue before AFAIK - does the new implementation introduce this bug? Is there some pre/post conditions that we need to check that would be helpful?

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -439,21 +430,21 @@ void hiopVectorRajaPar::copyFromStarting(int start_index_in_this, const double*
...
 
   // If there is nothing to copy, return.
   if(v.n_local_ == 0)
     return;
 
-  auto& rm = umpire::ResourceManager::getInstance();
-  double* vdata = const_cast<double*>(v.data_dev_); // scary:
-  rm.copy(this->data_dev_ + start_index, vdata, v.n_local_*sizeof(double));
+  //TODO: data_dev_+start_index   -> is not portable, may not work on the device. RAJA loop should be used
+  exec_space_.copy(data_dev_+start_index, v.data_dev_, v.n_local_, v.exec_space_);
```
CameronRutherford, line 447: Same here w.r.t. assertions or pre/post condition checking.

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -466,18 +457,19 @@ void hiopVectorRajaPar::copyFromStarting(int start_index, const hiopVector& src)
...
   // If nothing to copy, return.
   if(nv == 0)
     return;
-
-  auto& rm = umpire::ResourceManager::getInstance();
-  double* vv = const_cast<double*>(v); // <- cast away const
-  rm.copy(data_dev_, vv + start_index_in_v, nv*sizeof(double));
+  
+  //TODO: v+start_index_in_v   -> is not portable, may not work on the device. RAJA loop should be used
+  exec_space_.copy(data_dev_, v+start_index_in_v, nv);
```
CameronRutherford, line 472: Same here.

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -514,10 +506,11 @@ void hiopVectorRajaPar::startingAtCopyFromStartingAt(
 
   assert(howManyToCopyDest <= howManyToCopySrc);
 
-  auto& rm = umpire::ResourceManager::getInstance();
-  rm.copy(this->data_dev_ + start_idx_dest, 
-          v.data_dev_ + start_idx_src, 
-          howManyToCopyDest*sizeof(double));
+  //TODO: this also looks like is not portable
```
CameronRutherford, line 509: Same here

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -542,8 +536,8 @@ void hiopVectorRajaPar::copyToStarting(int start_index, hiopVector& dst) const
   if(v.n_local_ == 0)
     return;
 
-  auto& rm = umpire::ResourceManager::getInstance();
-  rm.copy(v.data_dev_, this->data_dev_ + start_index, v.n_local_*sizeof(double));
+  //TODO: pointer arithmetic on host should be avoided
+  v.exec_space_.copy(v.data_dev_, this->data_dev_ + start_index, v.n_local_, exec_space_);
```
CameronRutherford, line 540: Again, this hasn't been an issue before, and assume pre/post conditions are validated this should be fine?

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -555,29 +549,30 @@ void hiopVectorRajaPar::copyToStarting(int start_index, hiopVector& dst) const
...
+  auto& v = dynamic_cast<hiopVectorRaja<MEM, POL>&>(vec);
   assert(start_index_in_dest+n_local_ <= v.n_local_);
 
   // If there is nothing to copy, return.
   if(n_local_ == 0)
     return;
 
-  auto& rm = umpire::ResourceManager::getInstance();
-  rm.copy(v.data_dev_ + start_index_in_dest, this->data_dev_, this->n_local_*sizeof(double));
+  //TODO: pointer arithmetic on host should be avoided
```
CameronRutherford, line 562: Same here.

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -774,19 +772,23 @@ void hiopVectorRajaPar::startingAtCopyToStartingAt(
 
   if(num_elems == 0)
     return;
-  auto& rm = umpire::ResourceManager::getInstance();
-  rm.copy(dest.data_dev_ + start_idx_dest, this->data_dev_ + start_idx_in_src, num_elems*sizeof(double));
+
+  //rm.copy(dest.data_dev_ + start_idx_dest, this->data_dev_ + start_idx_in_src, num_elems*sizeof(double));
+  //TODO: fix pointer arithmetic on host
+  dest.exec_space_.copy(dest.data_dev_+start_idx_dest, data_dev_+start_idx_in_src, num_elems, exec_space_);
```
CameronRutherford, line 778: Same w.r.t. pre/post consitions.

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -774,19 +772,23 @@ void hiopVectorRajaPar::startingAtCopyToStartingAt(
...
-                                                             size_type num_elems/*=-1*/) const
+template<class MEM, class POL>
+void hiopVectorRaja<MEM,POL>::
+startingAtCopyToStartingAt_w_pattern(index_type start_idx_in_src,
+                                     hiopVector& destination,
+                                     index_type start_idx_dest,
+                                     const hiopVector& selec_dest,
+                                     size_type num_elems/*=-1*/) const
 {
 #if 0
```
CameronRutherford, line 789: It isn't a part of the PR, but I really dislike these dangling `#if 0` (or `#if 1`) pre-processor directives. We should either remove them or control things via a real variable that is CMake toggleable.

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -827,12 +829,16 @@ void hiopVectorRajaPar::startingAtCopyToStartingAt_w_pattern(index_type start_id
...
  */
-void hiopVectorRajaPar::copyTo(double* dest) const
+template<class MEM, class POL>
+void hiopVectorRaja<MEM,POL>::copyTo(double* dest) const
 {
-  auto& rm = umpire::ResourceManager::getInstance();
-  rm.copy(dest, this->data_dev_, n_local_*sizeof(double));
+  auto* This_const = const_cast<hiopVectorRaja*>(this);
+  assert(nullptr != This_const);
+  This_const->exec_space_.copy(dest, data_dev_, n_local_);
```
CameronRutherford, line 841:
```suggestion
  auto* this_const = const_cast<hiopVectorRaja*>(this);
  assert(nullptr != this_const);
  this_const->exec_space_.copy(dest, data_dev_, n_local_);
```
This would ideally be caught by a linter, but this is inconsistent with other variables.

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -2024,14 +2073,11 @@ bool hiopVectorRajaPar::isfinite_local() const
  * 
  * @pre Vector data was moved from the memory space to the host mirror.
  */
-void hiopVectorRajaPar::print(FILE* file/*=nullptr*/, const char* msg/*=nullptr*/, int max_elems/*=-1*/, int rank/*=-1*/) const
+template<class MEM, class POL>
+void hiopVectorRaja<MEM,POL>::
+print(FILE* file, const char* msg/*=NULL*/, int max_elems/*=-1*/, int rank/*=-1*/) const
```
CameronRutherford, line 2078: These default values should be enforced either in header file or here... I commented on another function that had default values removed, so wanted to make sure that functionality wasn't lost.

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -2024,14 +2073,11 @@ bool hiopVectorRajaPar::isfinite_local() const
...
 {
-  int myrank=0, numranks=1;
-
-  if(nullptr == file) {
-    file = stdout;
-  }
```
CameronRutherford, line 2033: I also think keeping this logic could be helpful, if not just defaulting the `file` variable to `stdout`.

src/LinAlg/hiopVectorRajaImpl.hpp
```diff
@@ -2062,48 +2108,43 @@ void hiopVectorRajaPar::print(FILE* file/*=nullptr*/, const char* msg/*=nullptr*
...
-    return;
-  auto& resmgr = umpire::ResourceManager::getInstance();
-  resmgr.copy(data_host_, data_dev_);
+  auto* This = const_cast<hiopVectorRaja<MEM, POL>* >(this);
+  assert(nullptr != This);
+  This->copyFromDev();
+  for(index_type it=0; it<n_local_; it++) {
+    printf("vec [%d] = %1.16e\n",it+1,data_host_[it]);
+  }
+  printf("\n");
```
CameronRutherford, line 2126:
```suggestion
  auto* temp_vec = const_cast<hiopVectorRaja<MEM, POL>* >(this);
  assert(nullptr != temp_vec);
  temp_vec->copyFromDev();
  for(index_type it=0; it<n_local_; it++) {
    printf("vec [%d] = %1.16e\n",it+1,data_host_[it]);
  }
  printf("\n");
```
Perhaps this is too pedantic...

CameronRutherford (changes requested):
Perhaps the biggest PR I have ever reviewed. Amazing work @cnpetra, and while I made 43 comments, they are all relatively minor.

tests/LinAlg/vectorTestsRajaPar.hpp
```diff
@@ -76,7 +76,7 @@ class VectorTestsRajaPar : public VectorTests
   virtual ~VectorTestsRajaPar(){}
 
 private:
-  virtual const real_type* getLocalDataConst(const hiop::hiopVector* x);
+  virtual const real_type* getLocalDataConst(hiop::hiopVector* x);
```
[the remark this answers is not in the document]
cnpetra: (some) vector implementations can change, and yes, it because of the host/device memory transfer needed. still returning a const, which I think is ok for the purpose of testing...

src/LinAlg/hiopMatrixRajaDense.cpp
```diff
@@ -67,15 +67,39 @@
...
 
-#include "hiopVectorPar.hpp"
-#include "hiopVectorRajaPar.hpp"
+//TODO:
+// - introduce hip and cuda .cpp
+// - Use vector deduced from this class' template parameters
+#ifdef HIOP_USE_CUDA
+#include <ExecPoliciesRajaCudaImpl.hpp>
+using ExecPolicyRajaType = hiop::ExecPolicyRajaCuda;
+using hiopVectorRajaT = hiop::hiopVectorRaja<hiop::MemBackendUmpire, hiop::ExecPolicyRajaCuda>;
```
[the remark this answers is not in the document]
cnpetra: it's going to be revisited when the raja dense matrix class is going to be split in implementation header and source files for different backends/compilers, similarly to what this PR does to vector raja class. At that point `hiopVectorRajaT` is going be set accordingly to the execution backend implementation file.

src/LinAlg/MatrixSparseCsrCudaKernels.hpp
```diff
@@ -102,20 +123,28 @@ void csr_get_diag_kernel(int n,
...
  */
-void csr_scalerows_kernel(int nrows, int ncols, int nnz, int* irowptr, int* jcoldind, double* values, const double* D); 
+void csr_scalerows_kernel(int nrows,
+                          int ncols,
+                          int nnz,
+                          int* irowptr,
+                          int* jcoldind,
+                          double* values,
+                          const double* D,
+                          int block_size);
```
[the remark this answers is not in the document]
cnpetra: it would be great if you set up the linter for HiOp as well.

src/ExecBackends/ExecSpace.hpp
cnpetra: currently cuda and hip is exclusive, but not necessarily the case in the future

src/ExecBackends/MemBackendUmpireImpl.hpp
cnpetra: we could, but right now I prefer to avoid "under the hood" device/host transfers

cnpetra: since we don't have a real need for it rn

src/LinAlg/hiopVectorRaja.hpp
cnpetra: for sure will be included in the docs for the next release

src/LinAlg/hiopVectorRajaImpl.hpp
cnpetra: yes, compiler would not confuse the brackets for operators

src/LinAlg/hiopVectorRajaImpl.hpp
cnpetra: it's just that I find it unsafe to do device pointer arithmetic on host. Works fine on both amd and cuda right now.

src/LinAlg/hiopVectorRajaCuda.cpp
```diff
@@ -0,0 +1,33 @@
+#include "hiopVectorRaja.hpp"
+
+#include "MemBackendUmpireImpl.hpp"
+#include "ExecPoliciesRajaCudaImpl.hpp"
```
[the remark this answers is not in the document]
pelesh: Here is a [prototype](https://github.com/pelesh/templates/blob/csr-proto/MatrixSparseCSRCuda.cpp) how to address this, which I created a while ago. Might be helpful in this case.

src/LinAlg/hiopVectorRajaImpl.hpp
pelesh: +1 Variable name `This` is too close to keyword `this`.

src/LinAlg/hiopVectorRajaImpl.hpp
pelesh: Google for "dot-template": https://stackoverflow.com/questions/8463368/template-dot-template-construction-usage.

I have to agree with @CameronRutherford, this syntax looks awful. :)

src/LinAlg/hiopVectorRajaCuda.cpp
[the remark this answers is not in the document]
cnpetra: I am not following anymore, please feel free to submit a PR with a concrete solution.

src/LinAlg/hiopVectorRajaImpl.hpp
cnpetra: that code is commented out and not removed in case it is going to be needed  later on.

src/LinAlg/hiopVectorRajaCuda.cpp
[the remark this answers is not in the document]
pelesh: Let me take a closer look at the code and make more specific suggestions.

src/LinAlg/hiopMatrixSparseCsrCuda.cpp
cnpetra: done

src/LinAlg/CMakeLists.txt
```diff
@@ -140,7 +150,7 @@ endif(HIOP_USE_CUDA)
 
 # Add RAJA/Umpire sources when enabled
 if(HIOP_USE_RAJA)
-  list(APPEND hiopLinAlg_INTERFACE_HEADERS hiop_raja_defs.hpp)
+  #list(APPEND hiopLinAlg_INTERFACE_HEADERS hiop_raja_defs.hpp)
```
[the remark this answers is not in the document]
cnpetra: done

src/LinAlg/hiopVectorRaja.hpp
cnpetra: updated to code to default to `stdout` when the output file is null.

src/LinAlg/hiopVectorRajaImpl.hpp
cnpetra: fixed

src/LinAlg/hiopVectorRajaImpl.hpp
cnpetra: fixed

src/LinAlg/hiopVectorRajaImpl.hpp
cnpetra: fixed

Issue #542, opened 2022-09-02
Ginkgo+cuda fails on Ascent if build type is not Debug


maksud:
Ginkgo with CUDA fails when HiOp is not built with build type `Debug`. By default, the spack builds with `RelWithDebInfo` therefore the default spack installation fails with the following message:

```
bash-4.4$ jsrun -n1 -g1 ./src/Drivers/Sparse/NlpSparseEx2.exe 500 -ginkgo_cuda
[h49n01:1092778] *** Process received signal ***
[h49n01:1092778] Signal: Segmentation fault (11)
[h49n01:1092778] Signal code: Invalid permissions (2)
[h49n01:1092778] Failing at address: 0x200112e20400
[h49n01:1092778] [ 0] linux-vdso64.so.1(__kernel_sigtramp_rt64+0x0)[0x2000000504d8]
[h49n01:1092778] [ 1] ./src/Drivers/Sparse/NlpSparseEx2.exe[0x101c236c]
[h49n01:1092778] [ 2] ./src/Drivers/Sparse/NlpSparseEx2.exe[0x101c370c]
[h49n01:1092778] [ 3] ./src/Drivers/Sparse/NlpSparseEx2.exe[0x10172e6c]
[h49n01:1092778] [ 4] ./src/Drivers/Sparse/NlpSparseEx2.exe[0x10176bc8]
[h49n01:1092778] [ 5] ./src/Drivers/Sparse/NlpSparseEx2.exe[0x1017530c]
[h49n01:1092778] [ 6] ./src/Drivers/Sparse/NlpSparseEx2.exe[0x1016bb64]
[h49n01:1092778] [ 7] ./src/Drivers/Sparse/NlpSparseEx2.exe[0x10048a50]
[h49n01:1092778] [ 8] /lib64/power9/libc.so.6(+0x24078)[0x20000ef54078]
[h49n01:1092778] [ 9] /lib64/power9/libc.so.6(__libc_start_main+0xb4)[0x20000ef54264]
[h49n01:1092778] *** End of error message ***
ERROR:  One or more process (first noticed rank 0) terminated with signal 11 (core dumped)
```

However, setting the build type to Debug still doesn't solve the other Ginkgo issue on Summit, Marianas, etc., that is the following error:

```
Setting up Ginkgo solver ...
terminate called after throwing an instance of 'gko::CudaError'
[...]

Issue #540, opened 2022-09-01
Ginkgo + CUDA Tests Fail on Marianas


CameronRutherford:
The error message is here:
```shell
27: Setting up Ginkgo solver ... 
27: terminate called after throwing an instance of 'gko::CudaError'
27:   what():  /tmp/ruth521/spack-stage/spack-stage-ginkgo-glu_experimental-dbmokiqc3tlyvnwehe546lb25lrnuaod/spack-src/cuda/base/executor.cpp:192: raw_copy_to: cudaErrorInvalidValue: invalid argument
27: [dlt03:33417] *** Process received signal ***
```

Marianas is CentOS 7 with a max compute capability of 60, and so the current assumption is that there is a bug with that specific build combination (see [here](https://github.com/LLNL/hiop/pull/521#issuecomment-1206800344) in #521 ).

The current test config to `ctest` disables relevant tests, and so it may appear CI is passing: https://github.com/LLNL/hiop/blob/develop/.gitlab-ci.yml#L164

CameronRutherford:
Relevant failing test logs in PNNL CI:

<details>
<summary>
NlpSparse1_6 (Click to show logs)

[PNNL Pipeline](https://gitlab.pnnl.gov/exasgd/solvers/hiop-mirror/-/jobs/97388#L3187)

</summary>

```shell
test 22
      Start 22: NlpSparse1_6
22: Test command: /share/apps/openmpi/4.1.0/gcc/10.2.0/bin/mpirun "-n" "1" "/people/svcexasgd/gitlab/97388/build/src/Drivers/Sparse/NlpSparseEx1.exe" "500" "-ginkgo_cuda" "-selfcheck"
22: Test timeout computed to be: 1800
22: [1659713290.594617] [dlt03:33270:0]    ucp_context.c:1470 UCX  WARN  UCP version is incompatible, required: 1.10, actual: 1.8 (release 0 /usr/lib64/libucp.so.0)
22: [1659713290.621398] [dlt03:33270:0]       ib_iface.c:665  UCX  ERROR ibv_create_cq(cqe=4096) failed: Cannot allocate memory
22: [dlt03.local:33270] pml_ucx.c:273  Error: Failed to create UCP worker
22: ===============
22: Hiop SOLVER
22: ===============
22: Using 1 MPI ranks.
22: ---------------
22: Problem Summary
22: ---------------
22: Total number of variables: 500
22:      lower/upper/lower_and_upper bounds: 499 / 1 / 1
[...]

PR #548, opened 2022-09-22
Fix segfault, remove nonsymmetric ginkgo solver

fritzgoebel:
This PR should fix issues #540 and #542. It also removes the nonsymmetric ginkgo solver as previously suggested by @pelesh.

The problem were a missing host side copy of the rhs vector for #540 and directly accessing matrix and vector values located in GPU memory in `update_matrix` for 
#542. This is fine as long as they are in unified memory, which the apparently are when built with Debug. I added a host side copy of the matrix that gets updated and moved back to the GPU instead.

### DIFF
--- a/scripts/gcc-cuda.cmake
+++ b/scripts/gcc-cuda.cmake
@@ -25,7 +25,7 @@ set(HIOP_SPARSE ON CACHE BOOL "")
 message(STATUS "Enabling HiOp's deepchecking")
 set(HIOP_DEEPCHECKS ON CACHE BOOL "")
 
-message(STATUS "Setting default cuda architecture to 60")
-set(CMAKE_CUDA_ARCHITECTURES 60 CACHE STRING "")
+set(CMAKE_CUDA_ARCHITECTURES 60 70 75 80 CACHE STRING "")
+message(STATUS "Setting default cuda architecture to ${CMAKE_CUDA_ARCHITECTURES}")
 
 message(STATUS "Done preloading CMake cache with values for continuous integration")
--- a/scripts/marianasVariables.sh
+++ b/scripts/marianasVariables.sh
@@ -91,5 +91,4 @@ EOD
 export NVBLAS_CONFIG_FILE=$PWD/nvblas.conf
 echo "Generated $PWD/nvblas.conf"
 
-export EXTRA_CMAKE_ARGS="$EXTRA_CMAKE_ARGS -DCMAKE_CUDA_ARCHITECTURES=60,70,75,80"
 export CMAKE_CACHE_SCRIPT=gcc-cuda.cmake
--- a/scripts/marianasVariables.sh
+++ b/scripts/marianasVariables.sh
@@ -1,68 +1,80 @@
 #  NOTE: The following is required when running from Gitlab CI via slurm job
 source /etc/profile.d/modules.sh
-module use -a /qfs/projects/exasgd/src/cameron/spack/share/spack/modules/linux-centos7-zen2
+module use -a /qfs/projects/exasgd/src/ci-deception/ci-modules/linux-centos7-zen2
 
 # Load spack-built modules
-# autoconf@2.69%gcc@10.2.0 patches=35c4492,7793209,a49dd5b arch=linux-centos7-zen2
-module load autoconf-2.69-gcc-10.2.0-r677m42
-# autoconf-archive@2022.02.11%gcc@10.2.0 patches=130cd48 arch=linux-centos7-zen2
-module load autoconf-archive-2022.02.11-gcc-10.2.0-pbrbzut
-# automake@1.16.5%gcc@10.2.0 arch=linux-centos7-zen2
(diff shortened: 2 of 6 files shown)


pelesh:
I think this is relevant for other GPU solvers in HiOp. 

@kswirydo @nychiang

fritzgoebel:
> I think this is relevant for other GPU solvers in HiOp.
> 
> @kswirydo @nychiang

Not sure it actually is, the GPU memory was coming from inside a ginkgo matrix

CameronRutherford:
PNNL Pipelines seem to be giving a false negative here. Tests failed initially, but are passing upon a retry. I would suggest adding a new commit with https://github.com/LLNL/hiop/blob/develop/.gitlab-ci.yml#L159 changed to not include `-E NlpSparse...`, and we can see if the failing tests are now passing!

Thankfully this doesn't require a rebuild, so this should be a quick check.

nkoukpaizan:
Tests are passing on `ascent` and `summit` with HiOp built in `Release` mode. I think this fixes #542.

CameronRutherford:
Seems like CUDA_ARCHITECTURES cannot readily be passed from the CLI in the platform variables script into the CMake config. As a workaround I have specified all cuda architectures for Marianas in `gcc-cuda.cmake`. Only downside is that now all CI platforms/developers that use this config will by default now build with all the architectures, rather than only on marianas as would be desireable.

CameronRutherford:
I was able to get tests passing when running manually on previously failing platform, and I have also re-built with same modules and incorrect cuda arch and was able to reproduce the issue.

I am not able to re-produce old issue with updated module set. I assume that since ginkgo is built with the fully array of cuda architectures, that the CUDA runtime is smart enough to pick up the right spec, despite HiOp being built with the "wrong" cuda arch.

Unless people have qualms with how I hacked this together, I think this should be good to merge.

As a side note, ExaGO now has updated modules ready to go with a fresh build of latest hiop@develop.

scripts/gcc-cuda.cmake
```diff
@@ -25,7 +25,7 @@ set(HIOP_SPARSE ON CACHE BOOL "")
 message(STATUS "Enabling HiOp's deepchecking")
 set(HIOP_DEEPCHECKS ON CACHE BOOL "")
 
-message(STATUS "Setting default cuda architecture to 60")
-set(CMAKE_CUDA_ARCHITECTURES 60 CACHE STRING "")
+set(CMAKE_CUDA_ARCHITECTURES 60 70 75 80 CACHE STRING "")
+message(STATUS "Setting default cuda architecture to ${CMAKE_CUDA_ARCHITECTURES}")
```
CameronRutherford, line 29: This doesn't actually seem necessary to resolve the issue. It seems like it is sufficient to just have updated Ginkgo module. Not sure if we want to keep this

scripts/marianasVariables.sh
```diff
@@ -1,68 +1,80 @@
...
+# cuda@11.4%gcc@10.2.0~allow-unsupported-compilers~dev arch=linux-centos7-zen2
+module load cuda-11.4-gcc-10.2.0-ewurpsv
+# camp@0.2.3%gcc@10.2.0+cuda~ipo+openmp~rocm~tests build_type=RelWithDebInfo cuda_arch=60,70,75,80 arch=linux-centos7-zen2
+module load camp-0.2.3-gcc-10.2.0-36lcy72
+# openblas@0.3.20%gcc@10.2.0~bignuma~consistent_fpcsr~ilp64+locking+pic+shared patches=9f12903 symbol_suffix=none threads=none arch=linux-centos7-zen2
+module load openblas-0.3.20-gcc-10.2.0-x6v3mwm
+# coinhsl@2019.05.21%gcc@10.2.0+blas arch=linux-centos7-zen2
+module load coinhsl-2019.05.21-gcc-10.2.0-gkzkws6
+# ginkgo@1.5.0.glu_experimental%gcc@10.2.0+cuda~develtools~full_optimizations~hwloc~ipo~oneapi+openmp~rocm+shared build_type=Release cuda_arch=60,70,75,80 arch=linux-centos7-zen2
+module load ginkgo-1.5.0.glu_experimental-gcc-10.2.0-x73b7k3
```
CameronRutherford, line 45: Example of just the updated Ginkgo module

fritzgoebel:
Thanks @CameronRutherford!

CameronRutherford (approved):
Approving, just pointing out points of confusion

pelesh:
> The only downside is that now all CI platforms/developers that use this config will by default now build with all the architectures, rather than only on Marianas as would be desirable.

This is not a big issue. If you want to distribute HiOp binary, for example, this is exactly how you would build it.

CameronRutherford:
All tests passing in CI

pelesh (approved):

CameronRutherford:
Passing spack builds in GitHub actions!

kswirydo (approved):

Issue #547, opened 2022-09-21
Unable to build with CUDA and without RAJA


CameronRutherford:
https://github.com/LLNL/hiop/blob/develop/CMakeLists.txt#L273

CameronRutherford:
cc @fritzgoebel

cnpetra:
I recently introduced that. I am very much for not having RAJA as dependency for anything cuSparse, but for now we need RAJA vectors in the condensed and normal equations linear systems. 

A quick fix would be to not build condensed and normal equations code when RAJA is not present. All the other HIOP_SPARSE =ON code should work without RAJA. 

Essentially guard
https://github.com/LLNL/hiop/blob/eaf241757999837bd62a3daf4597a52a6a2b7814/src/Optimization/CMakeLists.txt#L20
and 
https://github.com/LLNL/hiop/blob/eaf241757999837bd62a3daf4597a52a6a2b7814/src/Optimization/CMakeLists.txt#L21
with `if(HIOP_USE_RAJA)`

I expect code in the examples also need to be guarded in one or two places.

CameronRutherford:
https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/hiop/package.py

Currently it is valid to build `hiop+ginkgo+cuda` or `hiop+ginkgo+rocm` in the spack package. I think this is what @fritzgoebel was building recently, and so we might want to tweak that CMake logic if that configuration is possible.

Currently `hiop+cuda+cusolver` is also valid, but it seems that we would be missing raja in that case.

cnpetra:
closed by PR#549

PR #549, opened 2022-09-22
Enable CUDA build without RAJA

fritzgoebel:
This PR makes a build with shared libraries and CUDA possible without RAJA (see issue #547).
There might be more to do here, @cnpetra and @CameronRutherford please have a look.


pelesh:
This looks good to me and tests are passing. I am not sure how well are features affected by these changes covered by existing tests in HiOp, so I think @cnpetra needs to verify everything works as it should.

cnpetra (approved):

cnpetra:
To be completely consistent, I think we also need to guard the `if` in the NLP sparse example 2 [here](https://github.com/LLNL/hiop/blob/9319539d203831e609b2a9d8811baf94b3ca346e/src/Drivers/Sparse/NlpSparseEx2Driver.cpp#L295). The PR passes all CIs but I think HiOp switches to default sparse KKT formulation (without a warning).

PR #555, opened 2022-10-04
moves pnnl_cleanup to .pre stage and removes files based on age

jdmacam:
Updates the pnnl_cleanup job in .gitlab-ci.yml to:
- run during the .pre stage
- remove files based on age (over 6 hours old) to cleanup ci account space

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -191,14 +191,14 @@ build_on_incline:
     - if: '$CI_PROJECT_ROOT_NAMESPACE == "exasgd"'
 
 pnnl_cleanup:
-  stage: test
-  dependencies:
-    - build_on_incline
+  stage: .pre
   script:
+  # clears directory of files more than 6 hours/360 minutes old
   - |
     set -xv
-    export WORKDIR="$HOME/gitlab/$CI_JOB_ID/"
-    rm -rf "$WORKDIR"
+    export WORKDIR="$HOME/gitlab/"
+    find $WORKDIR -type d -mmin +360 -prune -print -exec rm -rf {} \; || true
+    ls -hal $WORKDIR
   <<: *pnnl_tags_definition
   rules:
     - if: '$CI_PROJECT_ROOT_NAMESPACE == "exasgd"'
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -191,14 +191,14 @@ build_on_incline:
     - if: '$CI_PROJECT_ROOT_NAMESPACE == "exasgd"'
 
 pnnl_cleanup:
-  stage: test
-  dependencies:
-    - build_on_incline
+  stage: .pre
```
CameronRutherford, line 194: https://stackoverflow.com/questions/56324070/with-gitlab-ci-how-to-disable-repository-clone-for-one-job - let's remove the cloning step for this pipeline.

CameronRutherford:
Since this is touching GitLab CI, could we also do some configuration of the git strategy for other pipelines?

I think we can disable cloning in the following:
- test_on_lassen
- test_on_quartz
- test_on_compute_node
- pnnl_cleanup

@nychiang for some reason the LLNL pipelines are also failing. I'm not as familiar with your CI setup (and don't have access), so feel free to clarify anything.

CameronRutherford:
I don't know if this should just be addressed in a different PR, but it would also be good to re-factor to move away from YAML anchors, we well as separating each platform's CI pipelines into separate files for readability.

@jdmacam if you want to leave that for another PR, please create an issue to track the refactoring (along with viewing Ascent/Crusher pipeline status) and we can merge this sooner rather than later.

nychiang:
@CameronRutherford I resubmit the pipeline and everything looks good to me. Probably our platform was too busy when you launched the pipeline.

jdmacam:
> Since this is touching GitLab CI, could we also do some configuration of the git strategy for other pipelines?
> 
> I think we can disable cloning in the following:
> 
> * test_on_lassen
> * test_on_quartz
> * test_on_compute_node
> * pnnl_cleanup

@CameronRutherford I've added the variable to those pipelines

CameronRutherford (approved):
LGTM and all pipelines pass. I think this can be merged.

You should notice tests passing a little bit faster as we will no longer have to clone the git repo before running tests where unecessary.

PR #557, opened 2022-10-11
Condensed/normaleqn linear system without CUDA/RAJA

nychiang:
Condensed/normaleqn linear system should work on CPU, using MA57 as the backend linear solver.

In the current `develop` branch, PR #549 changed the CI behaviors when neither RAJA nor CUDA is defined, i.e., hiop will switch `Condensed/normaleqn` to `xycyd` and solves the problem without any warning. In addition, #549 compiles the cpp files for condensed/normaleqn system only if raja is defined (see [here](https://github.com/LLNL/hiop/pull/549/files#diff-e0cb3c3e440dda836086390d5e72e9426680bd33f91225e6e7f64b749692cad8R43-R46)).

Since condensed/normaleqn system should work without raja, I correct the above issues in this PR. 
This PR also guarantees the feature @fritzgoebel introduced in #549, i.e., enable cuda build without raja. 
I believe now we can compile hiop with cuda and ginkgo, but without raja.

### DIFF
--- a/src/Optimization/CMakeLists.txt
+++ b/src/Optimization/CMakeLists.txt
@@ -17,6 +17,8 @@ set(hiopOptimization_SRC
 
 set(hiopOptimization_SPARSE_SRC
   hiopKKTLinSysSparse.cpp
+  hiopKKTLinSysSparseCondensed.cpp
+  hiopKKTLinSysSparseNormalEqn.cpp
   )
 
 set(hiopOptimization_INTERFACE_HEADERS
@@ -39,10 +41,6 @@ set(hiopOptimization_INTERFACE_HEADERS
   hiopPDPerturbation.hpp
   hiopResidual.hpp
   )
-
-if(HIOP_USE_RAJA)
-  list(APPEND hiopOptimization_SPARSE_SRC hiopKKTLinSysSparseCondensed.cpp hiopKKTLinSysSparseNormalEqn.cpp)
-endif()
 
 if(HIOP_SPARSE)
   list(APPEND hiopOptimization_SRC ${hiopOptimization_SPARSE_SRC})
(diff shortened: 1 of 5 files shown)


cnpetra:
I think this is kosher. In the PR diff, I do not see how/if MA57 and/or xdycyd gets picked up when RAJA/HIOP_SPARSE is not present in the build... but I guess that logic is already in place

src/Optimization/CMakeLists.txt
```diff
@@ -40,10 +42,6 @@ set(hiopOptimization_INTERFACE_HEADERS
   hiopResidual.hpp
   )
 
-if(HIOP_USE_RAJA)
-  list(APPEND hiopOptimization_SPARSE_SRC hiopKKTLinSysSparseCondensed.cpp hiopKKTLinSysSparseNormalEqn.cpp)
-endif()
```
fritzgoebel, line 45: This change will lead to `error: 'hiopVectorRajaPar' does not name a type`, e.g. in line 175 of src/Optimization/hiopKKTLinSysSparseCondensed.cpp when building with Ginkgo and CUDA but without RAJA.

cnpetra: Thank you! I was having a similar concern re: building w/o RAJA. Nai-Yuan and I will meet today and will address this.

cnpetra (approved):

src/Optimization/CMakeLists.txt
cnpetra: I think the issue has been fixed, it would be great if you could confirm

fritzgoebel (approved):
LGTM, I can still build without RAJA but with Ginkgo and Cuda enabled.

PR #558, opened 2022-10-17
Fix mds raja example for raja without GPU

nychiang:
MDS Ex1 RAJA failed if raja is build with OpenMP.
This PR fixes this issue.


cnpetra (approved):

nychiang:
> just to confirm, MDS uses cpu/lapack with OpenMP RAJA backend?

`HIOP_USE_MAGMA` is set to `OFF` when GPU is not available (see [here](https://github.com/LLNL/hiop/blob/develop/CMakeLists.txt#L89)). I just confirmed (via gdb) that hiop uses cpu/lapack with OpenMP RAJA backend.

Issue #556, opened 2022-10-06
Submodule not downloaded


SanPen:
Hi, 

I just downloaded the 0.7.0 release on ubuntu 20.04 and after installing missing dependencies (would be cool to list all those for *nix systems) I get this:

```
CMake Error at cmake/HiOpCheckGitSubmodules.cmake:22 (message):
  It seems required submodule was not downloaded! Please update submodules
  manually with the following command and try again.

  	$ git submodule update --init --recursive

Call Stack (most recent call first):
  CMakeLists.txt:202 (include)
```

I could indeed clone the repo and follow the instructions, but what would be the point on having a release?
Hence, please, publish a complete release.

Best regards,
Santiago

cnpetra:
thank you for reporting, we'll look into it

pelesh:
Santiago, could you please share full output of your cmake execution? Also, what happens when you try to update git submodules manually as the error message suggests?

CameronRutherford:
This is intended behaviour, but I can see how this is unclear.

CMake error is suggesting that you need to initialise submodule manually per @pelesh's suggestion.

cmake/HiOpCheckGitSubmodules.cmake - our CMake config should run this for you automatically, so I'm surprised you are running into this error. Perhaps `find_package(Git)` is failing, and hence your CMake isn't cloning things.

With respect to maintaining a list of required dependencies for *nix systems, we maintain a spack package for HiOp https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/hiop/package.py

Two notes from this
- We should update our GitHub docs to reflect submodule support
- We should update our Spack package with 0.7.0
- We should update ExaGO/HiOp docs to include something like `installing_with_spack`

I'm in the middle of refactoring the ExaGO spack configuration to use a submodule, and so this should be much easier when that's working and fully documented.

cnpetra:
the same error also reported here: https://github.com/xsdk-project/xsdk-issues/issues/187#event-7594423222

I think it makes sense to not error out when the check for `git submodule` fails. Just give a warning that the eigen needs to be installed manually given that `git submodule` is not available.

cnpetra:
Santiago (@SanPen), could you please take look at the new release (https://github.com/LLNL/hiop/releases/tag/v0.7.1) and let us know if it works for you. It should address all the issues you reported.

PR #561, opened 2022-10-19
git submodule fixes

cnpetra:
When the release archives are used to build HiOp, git submodule fails since the sources are not under a git repo. Failure can happen when git is simply not available on the machine. 

It appears that we need to reintroduce HIOP_EIGEN_DIR to allow user specify location of eigen.

### DIFF
--- a/cmake/HiOpCheckGitSubmodules.cmake
+++ b/cmake/HiOpCheckGitSubmodules.cmake
@@ -1,30 +1,43 @@
 find_package(Git QUIET)
-if(GIT_FOUND AND EXISTS "${PROJECT_SOURCE_DIR}/.git")
-  # Update submodules as needed
-  message(STATUS "Submodule update")
-  execute_process(
-    COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive
-    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
-    RESULT_VARIABLE GIT_SUBMOD_RESULT
-  )
-  if(NOT GIT_SUBMOD_RESULT EQUAL "0")
-    message(
-      FATAL_ERROR
-        "git submodule update --init --recursive failed with ${GIT_SUBMOD_RESULT}, please checkout submodules manually"
+if(GIT_FOUND)
+  if(EXISTS "${PROJECT_SOURCE_DIR}/.git")
+    # Update submodules as needed
+    message(STATUS "Submodule update")
+    execute_process(
+      COMMAND ${GIT_EXECUTABLE} submodule update --init --recursive
+      WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+      RESULT_VARIABLE GIT_SUBMOD_RESULT
     )
-  endif()
-endif()
+    if(NOT GIT_SUBMOD_RESULT EQUAL "0")
+      message(
+        FATAL_ERROR
+        "'git submodule update --init --recursive' failed with ${GIT_SUBMOD_RESULT}. "
+        "Please checkout submodules manually with the folowing command and run cmake again:\n"
+        "\t$ git submodule update --init --recursive\n"
+        )
+    else()
+      # For each submodule, we check a file in that submodule we expect to find to
+      # make sure the submodules were downloaded correctly.
+      foreach(CHK_FILE "eigen/Eigen/Core")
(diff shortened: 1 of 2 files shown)


cnpetra:
The current implementation is a draft/discussion started, especially this part: https://github.com/LLNL/hiop/pull/561/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR202

Looks like a better way is to factor out code that checks that eigen is present out of the HiopCheckGitModules.cmake and reuse that for when user provides the location of eigen. Comments & suggestions will be appreciated.

cmake/HiOpCheckGitSubmodules.cmake
```diff
@@ -1,30 +1,43 @@
...
     )
-  endif()
-endif()
+    if(NOT GIT_SUBMOD_RESULT EQUAL "0")
+      message(
+        FATAL_ERROR
+        "'git submodule update --init --recursive' failed with ${GIT_SUBMOD_RESULT}. "
+        "Please checkout submodules manually with the folowing command and run cmake again:\n"
+        "\t$ git submodule update --init --recursive\n"
+        )
```
pelesh, line 17: Is the "fatal error" the best way to go about this? Perhaps just disabling Eigen would suffice. One can still build other components of HiOp. My understanding is that fatal error would prevent building HiOp altogether.

cnpetra: > Is the "fatal error" the best way to go about this? Perhaps just disabling Eigen would suffice. One can still build other components of HiOp. My understanding is that fatal error would prevent building HiOp altogether.

we can go with a warning, but that is really an error on the user ends because `.git` exists but `git submodule` fails

cnpetra:
reviewers, PR was tested with hiop source from release archive and hiop source under git

pelesh:
@cnpetra, how essential Eigen dependency will be for HiOp in the future? If you anticipate you would use eigen for most of HiOp configurations going forward, perhaps the easiest would be to require user to always clone HiOp recursively (see e.g. RAJA and BLT).

Another option could be to use Spack to get Eigen as we do for other dependencies. Eigen package is in Spack upstream. 

I am not necessarily favoring these options, but might be worthwhile considering them at this time.

cnpetra:
> @cnpetra, how essential Eigen dependency will be for HiOp in the future? If you anticipate you would use eigen for most of HiOp configurations going forward, perhaps the easiest would be to require user to always clone HiOp recursively (see e.g. RAJA and BLT).
> 
> Another option could be to use Spack to get Eigen as we do for other dependencies. Eigen package is in Spack upstream.
> 
> I am not necessarily favoring these options, but might be worthwhile considering them at this time.

My preference is to have eigen (and any other dependencies) available without git and spack.

pelesh:
> 
> My preference is to have eigen (and any other dependencies) available without git and spack.

If Eigen does not have a special dependency status, then perhaps we should not provide an "extra service" to the user through git submodules. Instead, we could use the same CMake find approach as for other HiOp dependencies, e.g. UMFPACK.

nychiang (approved):
lgtm

CameronRutherford (approved):

CameronRutherford:
I think we can just merge this to resolve #556.

I don't see any reason why Eigen should be managed as a submodule instead of through Spack like our other dependencies. Perhaps we should plan to transition this to the Spack package with the next release of HiOp?

pelesh (approved):
I suggest using same CMake find approach for Eigen as we use for other dependencies. In the meantime, we can merge this to address immediate issues.

cnpetra:
> I think we can just merge this to resolve #556.
> 
> I don't see any reason why Eigen should be managed as a submodule instead of through Spack like our other dependencies. Perhaps we should plan to transition this to the Spack package with the next release of HiOp?

by managing through spack you mean to have -DHIOP_USE_EIGEN and -DHIOP_EIGEN_DIR ?

CameronRutherford:
> > I think we can just merge this to resolve #556.
> > I don't see any reason why Eigen should be managed as a submodule instead of through Spack like our other dependencies. Perhaps we should plan to transition this to the Spack package with the next release of HiOp?
> 
> by managing through spack you mean to have -DHIOP_USE_EIGEN and -DHIOP_EIGEN_DIR ?

Yes. I am working through whether we can do the same for other submodules in ExaGO. It would be nice to have spack be the only submodule in both projects. cc @pelesh

PR #562, opened 2022-10-20
Disable openmp when raja is not used

nychiang:
fix https://github.com/xsdk-project/xsdk-issues/issues/181

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -181,7 +181,7 @@ if(HIOP_DEVELOPER_MODE)
   install(TARGETS hiop_options hiop_warnings EXPORT hiop-targets)
 endif()
 
-if(NOT (HIOP_USE_GPU AND HIOP_USE_HIP))
+if(HIOP_USE_RAJA AND (NOT (HIOP_USE_GPU AND HIOP_USE_HIP)))
   find_package(OpenMP)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -181,7 +181,7 @@ if(HIOP_DEVELOPER_MODE)
   install(TARGETS hiop_options hiop_warnings EXPORT hiop-targets)
 endif()
 
-if(NOT (HIOP_USE_GPU AND HIOP_USE_HIP))
+if(HIOP_USE_RAJA AND (NOT (HIOP_USE_GPU AND HIOP_USE_HIP)))
```
CameronRutherford, line 184: Agreed that this closes the xSDK issue, however, we should update in future to fix this based on our discussion.

CameronRutherford (approved):

CameronRutherford:
It would be great if we could tag another minor release to have xSDK builds passing on mac

cnpetra:
> It would be great if we could tag another minor release to have xSDK builds passing on mac

will prepare tag 0.7.1 sometime before Monday.

CMakeLists.txt
nychiang: I agree! I saw you have already created an issue for it (#563)

PR #564, opened 2022-10-21
Pre stage pipeline fix

ryandanehy:
Adds a -mindepth 1 tag to solve the deleting parent directory pipeline cleanup issue.


CameronRutherford (approved):
@cnpetra this should resolve the false negatives that we have been seeing. Please merge, although it seems as though there are still other issues with race conditions present in some scenarios in CI.

PR #569, opened 2022-11-09
Fix `hiopVectorRajaPar::copyToStartingAt_w_pattern`

nychiang:
Fix `hiopVectorRajaPar::copyToStartingAt_w_pattern`, for running hiop sparse on device.
Add corresponding unit test.

The current implementation looks ugly, but it works.
Do we have a better way to implement the following function?
image

### DIFF
--- a/src/LinAlg/hiopVectorRajaPar.cpp
+++ b/src/LinAlg/hiopVectorRajaPar.cpp
@@ -101,6 +101,7 @@ hiopVectorRajaPar::hiopVectorRajaPar(
   index_type* col_part /* = NULL */,
   MPI_Comm comm /* = MPI_COMM_NULL */)
   : hiopVector(),
+    idx_cumsum_{nullptr},
     mem_space_(mem_space),
     comm_(comm)
 {
@@ -149,7 +150,8 @@ hiopVectorRajaPar::hiopVectorRajaPar(
 }
 
 hiopVectorRajaPar::hiopVectorRajaPar(const hiopVectorRajaPar& v)
-  : hiopVector()
+  : hiopVector(),
+   idx_cumsum_{nullptr}
 {
   n_local_ = v.n_local_;
   n_ = v.n_;
@@ -190,6 +192,7 @@ hiopVectorRajaPar::~hiopVectorRajaPar()
   devalloc.deallocate(data_dev_);
   data_dev_  = nullptr;
   data_host_ = nullptr;
+  delete idx_cumsum_;
 }
 
 hiopVector* hiopVectorRajaPar::alloc_clone() const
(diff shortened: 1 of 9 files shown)


cnpetra:
I cannot think of anything else than the inclusive scan. An optimization would be to store the cumsum indexes and reuse them when the method is called again (it is called twice per vector per iteration).

nychiang:
@cnpetra another review please. 👍

src/LinAlg/hiopVectorRajaPar.cpp
```diff
@@ -563,34 +564,58 @@ void hiopVectorRajaPar::copyToStarting(hiopVector& vec, int start_index_in_dest)
...
+    RAJA_LAMBDA(RAJA::Index_type i)
     {
-      assert(id[i] == zero || id[i] == one);
-      if(id[i] == one){
-        vd[start_index_in_dest+find_nnz] = dd[i];
-        find_nnz++;
+      if(nnz_in_row[i] != nnz_in_row[i-1]){
+        assert(nnz_in_row[i] == nnz_in_row[i-1] + 1);
+        index_type idx_dest = nnz_in_row[i-1] + start_index_in_dest;
+        vd[idx_dest] = dd[i-1];
```
cnpetra, line 615: I would add an `assert` to check `idx_dest` is less than `vec.n_local_`

src/LinAlg/hiopVectorRajaPar.hpp
```diff
@@ -153,7 +153,10 @@ class hiopVectorRajaPar : public hiopVector
   virtual void copyToStarting(int start_index_in_src, hiopVector& v) const;
   /* Copy 'this' to v starting at start_index in 'v'. */
   virtual void copyToStarting(hiopVector& v, int start_index_in_dest) const;
-  virtual void copyToStartingAt_w_pattern(hiopVector& v, int start_index_in_dest, const hiopVector& ix) const;
+  virtual void copyToStartingAt_w_pattern(hiopVector& v,
+                                          int start_index_in_dest,
+                                          const hiopVector& ix,
+                                          hiopVectorInt* idx_map = nullptr) const;
```
cnpetra, line 159: I think having `idx_map` as a member in the vector class would offer better encapsulation. A good name would be `idx_cumsum_`

cnpetra (changes requested):

src/LinAlg/hiopVectorRajaPar.hpp
nychiang: the reason I didn't implemented it like the way you mentioned, is that `idx_cumsum_` may need to be updated in many functions. For example, in `copyTo` and `copyFrom`. See the latest commit for example. (note that I haven't reset `idx_map` in all the functions yet)

nychiang:
Use mutable variable.
@cnpetra

cnpetra (approved):

PR #570, opened 2022-11-16
Calculate the inverse objective scale correctly.

tepperly:
In a test problem, the iteration log shows the correct objective function, but the solution callback has the wrong value. Changing the value of `inv_obj_scale` fixes the problem. I am not sure if all the other effects are correct, but I suspect they are.


nychiang:
@tepperly Thanks for reporting this. It seems the dual values are computed correctly, while only the objective returns wrong values. I change the name of the corresponding variable. since the previous name is confusing. Please let me know how it works on your end. Cheers!

nychiang (approved):

nychiang:
@cnpetra seems all the CI stop working if a PR is filed from a folk repository.

PR #572, opened 2022-11-25
Development of `hiopVectorCuda`

nychiang:
implementation of  "hiopVectorCuda"
Maybe this file/class should be named as "hiopVectorCudaPar" since it also supports MPI (verified by unit tests).

Right now I manually set "mem_space" to "cuda" to run the unit tests. More works are required to address this issue after we merge #543  (memory backends and execution policies)


tests/testVector.cpp
```diff
@@ -312,7 +325,8 @@ int runTests(const char* mem_space, MPI_Comm comm)
     fail += test.vectorIsfinite(*v);
   }
 
-  fail += test.vector_is_equal(*x, *y, rank);
+  // TODO: remove
+  //fail += test.vector_is_equal(*x, *y, rank);
```
CameronRutherford, line 329: Should we just be removing this code altogether?

src/Drivers/CMakeLists.txt
```diff
@@ -1,8 +1,8 @@
 # Create examples and tests
-add_subdirectory(Dense)
+#add_subdirectory(Dense)
 if(HIOP_SPARSE)
   add_subdirectory(Sparse)
 endif()
-add_subdirectory(MDS)
-add_subdirectory(PriDec)
+#add_subdirectory(MDS)
+#add_subdirectory(PriDec)
```
CameronRutherford, line 7: Should we include these directories again in the future?

src/LinAlg/hiopLinAlgFactory.cpp
```diff
@@ -93,6 +98,14 @@ hiopVector* LinearAlgebraFactory::create_vector(const std::string& mem_space,
   const std::string mem_space_upper = toupper(mem_space);
   if(mem_space_upper == "DEFAULT") {
     return new hiopVectorPar(glob_n, col_part, comm);
+  } else if(mem_space_upper == "CUDA") {
+    #ifdef HIOP_USE_CUDA
+      return new hiopVectorCuda(glob_n, col_part, comm);
+    #else
+      assert(false && "requested memory space not available because Hiop was not"
+                      "built with CUDA support");
+      return new hiopVectorPar(glob_n, col_part, comm);
```
CameronRutherford, line 107: Similarly below, if you assert false, doesn't HiOp exit? Perhaps I am just misremembering how assert works, but if this is intended to warn gracefully, maybe we should reconsider.

src/LinAlg/hiopLinAlgFactory.cpp
```diff
@@ -117,6 +130,14 @@ hiopVectorInt* LinearAlgebraFactory::create_vector_int(const std::string& mem_sp
   const std::string mem_space_upper = toupper(mem_space);
   if(mem_space_upper == "DEFAULT") {
     return new hiopVectorIntSeq(size);
+  } else if(mem_space_upper == "CUDA") {
+    #ifdef HIOP_USE_CUDA
+      return new hiopVectorIntCuda(size, mem_space_upper);
+    #else
+      assert(false && "requested memory space not available because Hiop was not"
+                      "built with CUDA support");
```
CameronRutherford, line 138: Same comment about assert here, but also in other sections of code in this file

src/LinAlg/VectorCudaKernels.cu
```diff
@@ -0,0 +1,1267 @@
...
+/** @brief Checks if each component in `this` matches nonzero pattern of `select`.  */
+void component_match_pattern_kernel(int n_local, int* yd, const double* xd, const double* id)
+{
+  // compute linear damping term
+  int block_size=256;
+  int num_blocks = (n_local+block_size-1)/block_size;
+  set_match_pattern_cu<<<num_blocks,block_size>>>(n_local, yd, xd, id);
+}
+
+/** @brief Adjusts duals. */
```
CameronRutherford, line 860: I only skimmed this file, but I assume these comments should be in the header instead?

src/LinAlg/hiopVectorCuda.cpp
```diff
@@ -0,0 +1,1118 @@
...
+
+void hiopVectorCuda::copyFrom(const double* v_local_data)
+{
+  if(v_local_data) {
+    cudaError_t cuerr = cudaMemcpy(data_dev_, v_local_data, (n_local_)*sizeof(double), cudaMemcpyDeviceToDevice);
+    assert(cuerr == cudaSuccess);
+  }
+}
+
+/// @brief Copy from vec the elements specified by the indices in index_in_src
```
CameronRutherford, line 189: I also only skimmed this file, and same message here about comments.

CameronRutherford (changes requested):
Some relatively minor feedback

src/Drivers/CMakeLists.txt
nychiang: yes. My bad. I am running out of storage and hence I tried to remove some executable files.

src/LinAlg/hiopLinAlgFactory.cpp
nychiang: you are correct. The return function doesn't do anything here. I follow Asher's way to implement this part of code (see other lines in this file). I guess he added return function here to avoid compilation warning.

src/LinAlg/hiopVectorIntCuda.hpp
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * @brief Set the vector entries to be a linear space of starting at i0 containing evenly 
+   * incremented integers up to i0+(n-1)di, when n is the length of this vector
+   *
+   * @pre The elements of the linear space should not overflow the index_type type
+   *  
+   * @param i0 the starting element in the linear space (entry 0 in vector)
+   * @param di the increment for subsequent entries in the vector
+   *
+   */
```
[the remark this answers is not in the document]
nychiang: I mainly copy all the comments from other files, since no new functions have been introduced in this PR.
We may need to open a new PR to correct the comments in `hiopVectorInt.hpp`, `hiopVectorIntSeq.hpp`, `hiopVectorIntRaja.hpp` and this file. It seems identical comments should be added into these four files. 

Should we just add detailed comments in the abstract class, and use brief comments in the derived classes? @cnpetra

src/LinAlg/hiopVectorIntCuda.cpp
```diff
@@ -0,0 +1,145 @@
...
+/**
+ * @brief Set the vector entries to be a linear space of starting at i0 containing evenly 
+ * incremented integers up to i0+(n-1)di, when n is the length of this vector
+ *
+ * @pre The elements of the linear space should not overflow the index_type type
+ *  
+ * @param i0 the starting element in the linear space (entry 0 in vector)
+ * @param di the increment for subsequent entries in the vector
+ *
+ */
```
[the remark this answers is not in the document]
nychiang: Detailed comments should be in the source code, but @pre and @post conditions should be in the header file.

tests/LinAlg/vectorTestsCuda.hpp
```diff
@@ -0,0 +1,86 @@
...
+#pragma once
+
+#include "vectorTests.hpp"
+
+namespace hiop { namespace tests {
+
+/**
+ * @brief Utilities for testing hiopVectorCuda class
+ *
+ * @todo In addition to set and get element ass set and get buffer methods.
```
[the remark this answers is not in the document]
pelesh: Set and get element methods were intended to set or get a single element in the vector. If you want to set a pattern in a vector you need to do that in a buffer and that copy that buffer into the vector data.

This may look cumbersome, but the idea is to have these implementation-specific methods, which will enable creating implementation agnostic unit tests. That allows us to test different vector kernel implementations with the same unit test, ensuring that no matter the implementation, the mathematical result is the same.

src/LinAlg/hiopLinAlgFactory.cpp
pelesh: I think throwing exceptions and not catching them would be equivalent to assertions used in HiOp, but would be cleaner and more flexible solution going forward.

src/LinAlg/hiopVectorCuda.cpp
```diff
@@ -0,0 +1,1118 @@
...
+}
+
+/// Implementation of AXPY kernel
+void hiopVectorCuda::axpy(double alpha, const hiopVector& xvec)
+{
+  const hiopVectorCuda& x = dynamic_cast<const hiopVectorCuda&>(xvec);
+  int one = 1;
+  cublasStatus_t ret_cublas = cublasDaxpy(handle_cublas_, n_local_, &alpha, x.data_dev_, one, data_dev_, one);
+  assert(ret_cublas == CUBLAS_STATUS_SUCCESS);
+}
```
pelesh, line 572: I think if you use abstract interface method to access local vector data, you can avoid dynamic casting. Perhaps if local data accessor is a protected method in the base class that would be a good solution here?
```suggestion
void hiopVectorCuda::axpy(double alpha, const hiopVector& xvec)
{
  int one = 1;
  cublasStatus_t ret_cublas = cublasDaxpy(handle_cublas_, n_local_, &alpha, x.data(), one, data_dev_, one);
  assert(ret_cublas == CUBLAS_STATUS_SUCCESS);
}
```

src/LinAlg/hiopVectorCuda.cpp
```diff
@@ -0,0 +1,1118 @@
...
+
+/** Return the dot product of this hiopVector with v */
+double hiopVectorCuda::dotProductWith( const hiopVector& v ) const
+{
+  const hiopVectorCuda& vx = dynamic_cast<const hiopVectorCuda&>(v);
+  int one = 1;
+  double retval; 
+  cublasStatus_t ret_cublas = cublasDdot(handle_cublas_, n_local_, vx.data_dev_, one, data_dev_, one, &retval);
+  assert(ret_cublas == CUBLAS_STATUS_SUCCESS);
```
pelesh, line 675: If you already use Thrust, why use cuBLAS here?

src/LinAlg/hiopVectorCuda.cpp
pelesh: Same at several other places in the code.

src/LinAlg/hiopVectorCuda.cpp
```diff
@@ -0,0 +1,1118 @@
...
+size_type hiopVectorCuda::numOfElemsAbsLessThan(const double &val) const
+{
+  return hiop::cuda::num_of_elem_absless_than_kernel(n_local_, data_dev_, val);
+}
+
+void hiopVectorCuda::set_array_from_to(hiopInterfaceBase::NonlinearityType* arr, 
+                                          const int start, 
+                                          const int end, 
+                                          const hiopInterfaceBase::NonlinearityType* arr_src,
+                                          const int start_src) const
```
pelesh, line 1076: Indentation seems to be off here.

src/LinAlg/hiopVectorCuda.hpp
```diff
@@ -0,0 +1,342 @@
...
+  /// @brief check if `this` vector is identical to `vec`
+  virtual bool is_equal(const hiopVector& vec) const;
+
+  /* functions for this class */
+  inline MPI_Comm get_mpi_comm() const { return comm_; }
+
+private:
+  MPI_Comm comm_;
+  double* data_host_;
+  double* data_dev_;
```
pelesh, line 327: Given the implementation, consider using more intuitive variable names:
```suggestion
  double* data_host_mirror_;
  double* data_;
```
Your data is processed on device. You copy it to the host mirror for special purposes (e.g. for printing during debugging).

src/LinAlg/hiopVectorIntCuda.cpp
```diff
@@ -0,0 +1,145 @@
...
+// trade name, trademark, manufacturer or otherwise does not necessarily constitute or 
+// imply its endorsement, recommendation, or favoring by the United States Government or 
+// Lawrence Livermore National Security, LLC. The views and opinions of authors expressed 
+// herein do not necessarily state or reflect those of the United States Government or 
+// Lawrence Livermore National Security, LLC, and shall not be used for advertising or 
+// product endorsement purposes.
+
+/**
+ * @file hiopVectorIntCuda.cpp
+ *
```
pelesh, line 50: This file should probably have extension `*.cu`, to emphasize it can be compiled with CUDA only.

cnpetra (approved):
address the other reviewer's comments and will merge. Re: `data_host_`, we will revisit when we address the `'copyTo/FromDev` methods from the abstract vector interface.

pelesh (commented):
There are minor issues that @CameronRutherford and I pointed out that can be easily addressed. As soon as these are done, this PR can be merged.

In a more general context, I have mixed feelings about this PR. It essentially duplicates the capability that already exists in RAJA modules and therefore eliminates RAJA/Umpire as HiOp dependencies. On the other hand, it adds a couple of thousands lines of code more, only for CUDA capability. You would have to replicate the same for HIP and oneAPI, only to replace the RAJA dependency. I don't think this path is sustainable, but this is just my opinion. In any case, we have a real-life experiment to asses the value of portability libraries :).

src/LinAlg/VectorCudaKernels.cu
```diff
@@ -0,0 +1,1267 @@
...
+  // wrap raw pointer with a device_ptr 
+  thrust::multiplies<double> mult_op;
+  thrust::device_ptr<double> dev_v1 = thrust::device_pointer_cast(d1);
+  thrust::device_ptr<double> dev_v2 = thrust::device_pointer_cast(d2);
+  
+  thrust::transform(thrust::device,
+                    dev_v1, dev_v1+n,
+                    dev_v2, dev_v1,
+                    mult_op);
+}
```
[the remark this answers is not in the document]
nychiang: I firstly put all the thrust related code in a .cpp file and used C++ compiler for it. However, all works fine except for `thrust::transform_reduce", which gave me some compilation error. Tried to use nvcc to compile the same source code and then everything works fine. Therefore now I put everything in a .cu file.

src/LinAlg/hiopVector.hpp
```diff
@@ -275,16 +275,15 @@ class hiopVector
...
+                                                   const hiopVector& ix) const = 0;
   /// @brief Entries corresponding to zeros in ix are set to zero
   virtual void selectPattern(const hiopVector& ix) = 0;
   /// @brief checks whether entries in this matches pattern in ix
   virtual bool matchesPattern(const hiopVector& ix) = 0;
 
   /// @brief dual adjustment -> see hiopIterate::adjustDuals_primalLogHessian
-  virtual void adjustDuals_plh(const hiopVector& x, const hiopVector& ix,
-			       const double& mu, const double& kappa)=0;
+  virtual void adjustDuals_plh(const hiopVector& x, const hiopVector& ix, const double& mu, const double& kappa) = 0;
```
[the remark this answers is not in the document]
cnpetra: I do not understand why this method should not be here, but if it requires changes in abstract interface, please open an issue to track the issue and to follow up with another pr.

src/LinAlg/hiopVectorCuda.cpp
nychiang: My strategy is to use Thrust if no cuBlas function is available. If neither is available, then implement our own kernel functions.

src/LinAlg/hiopVectorIntCuda.hpp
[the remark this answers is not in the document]
nychiang: Created issue #576

src/LinAlg/hiopVectorCuda.hpp
```diff
@@ -0,0 +1,342 @@
...
+  virtual size_type numOfElemsAbsLessThan(const double &val) const;  
+
+  /// @brief set int array 'arr', starting at `start` and ending at `end`, to the values in `arr_src` from 'start_src`
+  /// TODO: add unit test
+  virtual void set_array_from_to(hiopInterfaceBase::NonlinearityType* arr, 
+                                 const int start, 
+                                 const int end, 
+                                 const hiopInterfaceBase::NonlinearityType* arr_src,
+                                 const int start_src) const;
+  virtual void set_array_from_to(hiopInterfaceBase::NonlinearityType* arr,
```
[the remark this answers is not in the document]
nychiang: Created issue https://github.com/LLNL/hiop/issues/576

src/LinAlg/hiopLinAlgFactory.cpp
nychiang: created issue #577

src/LinAlg/hiopVector.hpp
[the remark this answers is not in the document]
nychiang: I guess Slaven means this is not a general function for vector kernel, but a function for IPM algorithm.

src/LinAlg/hiopLinAlgFactory.cpp
nychiang: created issue #577

Issue #578, opened 2022-12-16
Re-enable test for condensed linear algebra when raja not present


cnpetra:
see around line 44 in [src/Drivers/Sparse/CMakeLists.txt](https://github.com/LLNL/hiop/pull/543/files/daad986535d5f8013518088b81f6c95e4e893f46#diff-da3068fcb5279d02749c139becc9f0c9c3c5fb509d9db33ae7af28c6fab8a0a2)

Issue #575, opened 2022-12-07
Execution backends options


cnpetra:
Will need to decide before the next release how to merge the `mem_space` and `compute_mode` options with the ones for execution backends options for memory backend and execution policies.

PR #579, opened 2022-12-19
Build: Cuda without raja 

cnpetra:
CUDA KKT and Matrix classes now use the newly added cuda vector. This removes dependencies of `HIOP_USE_CUDA` on `HIOP_USE_RAJA`. 

Close https://github.com/LLNL/hiop/issues/578


src/LinAlg/hiopMatrixSparseCsrCuda.cpp
```diff
@@ -301,11 +298,8 @@ void hiopMatrixSparseCSRCUDA::addDiagonal(const double& alpha, const hiopVector&
 {
   assert(nrows_ == D.get_size());
   assert(nrows_ == ncols_ && "Matrix must be square");
-#ifdef HIOP_USE_RAJA
-  assert(dynamic_cast<const hiopVectorCuda*>(&D) && "input vector must be RAJA-CUDA");
-#else
-  assert("input vector must be RAJA-CUDA, but HiOp was not built with RAJA");
-#endif
+  assert(dynamic_cast<const hiopVectorCuda*>(&D) && "input vector must be CUDA");
```
CameronRutherford, line 301: I tried to figure this out from other LinAlg objects, but couldn't get that far - should we place any of these asserts behind `#ifdef DEEPCHECKS`, or are they necessary to have during all HiOp execution?

src/LinAlg/hiopVectorCuda.cpp
```diff
@@ -192,6 +194,17 @@ void hiopVectorCuda::copy_from_w_pattern(const hiopVector& vv, const hiopVector&
   componentMult(select);
 }
 
+
+void hiopVectorCuda::copy_from_vec_par(const hiopVectorPar& src)
+{
+  assert(n_local_ == src.get_size());
```
CameronRutherford, line 200: Same for assertions in this file

src/Optimization/hiopKKTLinSysSparseCondensed.cpp
```diff
@@ -181,24 +170,24 @@ bool hiopKKTLinSysCondensedSparse::build_kkt_matrix(const hiopPDPerturbation& pd
...
   //temporary code, see above note
   {
-    if(mem_space_internal == "DEVICE") {
+    if(mem_space_internal == "CUDA") {
 #ifdef HIOP_USE_CUDA
-      auto Hd_raja = dynamic_cast<hiopVectorCuda*>(Hd_copy_);
+      auto Hd_cuda = dynamic_cast<hiopVectorCuda*>(Hd_copy_);
       auto Hd_par =  dynamic_cast<hiopVectorPar*>(Hd_);
-      assert(Hd_raja && "incorrect type for vector class");
+      assert(Hd_cuda && "incorrect type for vector class");
```
CameronRutherford, line 177: Similar for these assertions.

CameronRutherford (approved):
Just one stylistic comment about assertions in `LinAlg` and `Optimization` code.

Apart from that this looks great!

src/LinAlg/hiopMatrixSparseCsrCuda.cpp
cnpetra: they catch a supposedly "no-go" case caused by a potential change in the code by a developer (use of improper HiOp vector type), so it should pop up under debug builds. It is not user-related nor potentially some abnormal numerics, which sometimes go under deepchecks.

cnpetra:
> LGTM. It seems condensed system and normal equation depends on 'CUDA' vector, doesn't it? May be we should add some comments about this in the user guide and option file (hiopOptions.cpp).

https://github.com/LLNL/hiop/issues/575 is supposed to take care of it

PR #582, opened 2023-01-10
reorganization of device namespace 

cnpetra:
using `cuda` and `hip` namespaces instead of `device` in math kernels. Some file names were changed to ensure consistence in naming.


nychiang (approved):
LGTM.

PR #583, opened 2023-01-11
RAJA Vector int with ExecSpace

cnpetra:
Instrumentation of the int vector RAJA implementation with the newly added Exec spaces machinery


cnpetra:
passes tests on AMD/HIP as well

src/LinAlg/hiopVectorIntRajaCuda.cpp
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2017, Lawrence Livermore National Security, LLC.
```
nychiang, line 1: I saw you use 2022 in some other files. Do we need to use the new date for all the files?

nychiang (approved):
All the changes are clear to me. Cheers!

PR #584, opened 2023-01-12
Instrumentation of host vectors with execution spaces

cnpetra:
VectorPar and VectorIntSeq are now instrumented with `MemBackendCpp` and sequential execution policy for memory allocations and transfers.

### DIFF
--- a/src/LinAlg/hiopVectorPar.cpp
+++ b/src/LinAlg/hiopVectorPar.cpp
@@ -46,6 +46,7 @@
 // product endorsement purposes.
 
 #include "hiopVectorPar.hpp"
+#include "MemBackendCppImpl.hpp"
 #include "hiopVectorIntSeq.hpp"
 #include "hiopCppStdUtils.hpp"
 #include "MathHostKernels.hpp"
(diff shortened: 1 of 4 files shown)


src/LinAlg/hiopVectorPar.cpp
```diff
@@ -120,11 +121,15 @@ hiopVector* hiopVectorPar::new_copy () const
 
 void hiopVectorPar::setToZero()
 {
-  for(int i=0; i<n_local_; i++) data_[i]=0.0;
+  for(int i=0; i<n_local_; i++) {
+    data_[i]=0.0;
+  }
```
nychiang, line 126: I am fine with the code, but style guide implies that we should add spaces before and after '=' and '<'. \
There are many places where we do not add spaces before and after `=`, especially in the logical conditions for the loops. Should we update the style guide?

nychiang (commented):
LGTM. Just a quick question about style guide.

src/LinAlg/hiopVectorPar.cpp
cnpetra: I don't remember (and I can't find in the style guidelines) to be specific about those spaces. It can go either way afaik, but I tend to not use them.

Issue #581, opened 2023-01-09
removal of `Vector::copyToDev` and `Vector::copyFromDev`


cnpetra:
Exclusively used to get host mirror pointers for processing on the CPU, these two methods should be removed. Instead, whenever cpu processing is needed (in processing bounds and returning optimization variables when `callback_mem_space` differs from `mem_space`), one should create a host HiOp vector and copy from / to original vector before / after the host processing. Each vector class would have a copy_to/from_vector_par. This may also obliterate the need for `Vector::local_data_host()`. Another benefit would be that host mirror arrays will not be needed for CUDA and HIP vector implementations (potentially from RAJA vectors as well). 

This also applies to the matrices classes, where the host mirror is more pervasive.

cnpetra:
closed by pr#587

PR #587, opened 2023-01-25
remove copy from/to device methods in vector classes

cnpetra:
- Addresses issue #581 
- (pre)processing bounds and constraints in `hiopNlpFormulation` (currently is done on host) is now exclusively done via `hiopVector` interface, which facilitates the porting of this processing to device (this will be done by a future PR)
- instruments cuda vectors with execution spaces
- provides implementation of a couple of `copy` methods for device-host transfers needed by the cuda vectors


nychiang (commented):
LGTM. Just some minor issues.

PR #590, opened 2023-02-03
Implementation of `hiopVectorHip`

nychiang:
Add support of `hiopVector` with Hip backends.

### DIFF
--- a/src/ExecBackends/ExecSpace.hpp
+++ b/src/ExecBackends/ExecSpace.hpp
@@ -474,7 +474,7 @@ template<class MEMDEST, class EXEPOLDEST, class MEMSRC, class EXEPOLSRC, typenam
 struct TransferImpl
 {
   inline static bool do_it(T* p_dest,
-                           const ExecSpace<MEMDEST, EXEPOLDEST>& hwb_dest,
+                           ExecSpace<MEMDEST, EXEPOLDEST>& hwb_dest,
                            const T* p_src,
                            const ExecSpace<MEMSRC, EXEPOLSRC>& hwb_src,
                            const I& n)
--- a/src/ExecBackends/ExecSpace.hpp
+++ b/src/ExecBackends/ExecSpace.hpp
@@ -85,6 +85,10 @@ struct ExecSpaceInfo
       mem_backend_ = "CUDA";
       mem_backend_host_ = "STDCPP";
       exec_backend_ = "CUDA";
+    } else if(mem_space_ == "HIP") {
+      mem_backend_ = "HIP";
+      mem_backend_host_ = "STDCPP";
+      exec_backend_ = "HIP";
     } else {
       assert(mem_space_ == "DEVICE" || mem_space_ == "UM" || mem_space_ == "HOST");
       mem_backend_ = "UMPIRE";
--- a/src/ExecBackends/ExecSpace.hpp
+++ b/src/ExecBackends/ExecSpace.hpp
@@ -404,6 +404,33 @@ public:
     return TransferImpl<MEMBACKEND, EXECPOLICIES, MEMBACKEND, EXECPOLICIES, T, I>::do_it(p_dest, *this, p_src, *this, n);
   }
 
+  /**
+   * Copy `n` elements of the array `p_src` to the `p_dest` array.
+   * 
+   * @pre `p_src` and `p_dest` should be allocated so that they can hold at least 
+   * `n` elements.
+   * @pre `p_dest` should be managed by the memory backend of `this`.
+   * @pre `p_src` should be managed by the memory backend of `ms`.
+   */
+  template<class MEMSRC, class EXEPOLSRC, typename T, typename I>
+  inline bool copy(T* p_dest, const T* p_src, const I& n, const ExecSpace<MEMSRC,EXEPOLSRC>& ms) const
+  {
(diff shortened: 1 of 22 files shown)


cnpetra (changes requested):
impressive turnaround for this PR! It looks great to me, just left a couple of minor comments. please also look into why LLNL CI fails

src/LinAlg/hiopVectorIntHip.cpp
```diff
@@ -63,52 +63,32 @@
...
-  hipFree(buf_);
-}
-
-void hiopVectorIntHip::copy_from_dev()
-{
-  if (buf_ != buf_host_) {
-    hipError_t cuerr = hipMemcpy(buf_host_, buf_, (sz_)*sizeof(index_type), hipMemcpyDeviceToHost);
-    assert(cuerr == hipSuccess);
+  buf_ = exec_space_.template alloc_array<index_type>(sz);
+  if(exec_space_.mem_backend().is_device()) {
```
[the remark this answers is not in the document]
nychiang: I'm afraid you reviewed an old version. I fixed many things including this one in the next commits.
you'd better review the new changes again, since I need to change function signatures in `ExecSpace.hpp`.

src/LinAlg/hiopVectorCuda.cpp
```diff
@@ -287,11 +278,8 @@ void hiopVectorCuda::startingAtCopyFromStartingAt(int start_idx_dest,
...
 void hiopVectorCuda::copyTo(double* dest) const
 {
-  cudaError_t cuerr = cudaMemcpy(dest,
-                                 data_,
-                                 (n_local_)*sizeof(double),
-                                 cudaMemcpyDeviceToDevice);
-  assert(cuerr == cudaSuccess);
+  auto b = exec_space_.copy(dest, data_, n_local_);
+  assert(b);
 }
```
nychiang, line 283: This const method requires me to change the signature of function `do_it` [(see here)](https://github.com/LLNL/hiop/blob/45909b3c1dc4ae669b4833327e10d0e1f161315e/src/ExecBackends/ExecSpace.hpp#L477). I think this change is safe for now, since `hwb_dest` has never been changed in the code.

cnpetra: I understand. Since that `copyTo` method works will raw pointers, it will be removed soon. Please remove the `const` in `TransferImpl` and use a workaround in this method, e.g. const cast as in (https://github.com/LLNL/hiop/blob/d047558d69065c2d220d76d169188f7231826b58/src/LinAlg/hiopVectorRajaImpl.hpp#L844)

src/ExecBackends/ExecSpace.hpp
```diff
@@ -400,6 +404,33 @@ class ExecSpace
...
+  /**
+   * Copy `n` elements of the array `p_src` to the `p_dest` array.
+   * 
+   * @pre `p_src` and `p_dest` should be allocated so that they can hold at least 
+   * `n` elements.
+   * @pre `p_dest` should be managed by the memory backend of `this`.
+   * @pre `p_src` should be managed by the memory backend of `ms`.
+   */
+  template<class MEMSRC, class EXEPOLSRC, typename T, typename I>
+  inline bool copy(T* p_dest, const T* p_src, const I& n, const ExecSpace<MEMSRC,EXEPOLSRC>& ms) const
```
cnpetra, line 416: I think there is some confusion here. None of the `copy` should be `const`. Original methods should work.

src/LinAlg/hiopVectorCuda.cpp
```diff
@@ -1017,31 +1007,32 @@ hiopVector* hiopVectorCuda::new_copy () const
 /// @brief copy data from host mirror to device
 void hiopVectorCuda::copyToDev()
 {
-  cudaError_t cuerr = cudaMemcpy(data_, data_host_mirror_, (n_local_)*sizeof(double), cudaMemcpyHostToDevice);
-  assert(cuerr == cudaSuccess);
+  assert(exec_space_.mem_backend().is_device() && "should have data_dev_==data_host_");
```
cnpetra, line 1010: does this assert makes sense here? `is_device` is always true in this class.

src/LinAlg/hiopVectorHip.cpp
```diff
@@ -0,0 +1,1097 @@
...
+{
+  hiopVector* v = new hiopVectorHip(*this); assert(v);
+  v->copyFrom(*this);
+  return v;
+}
+
+/// @brief copy data from host mirror to device
+void hiopVectorHip::copyToDev()
+{
+  assert(exec_space_.mem_backend().is_device() && "should have data_dev_==data_host_");
```
cnpetra, line 1014: same as above. assert not needed

cnpetra (changes requested):

src/ExecBackends/ExecSpace.hpp
```diff
@@ -400,6 +404,19 @@ class ExecSpace
...
 
+  /**
+   * Copy `n` elements of the array `p_src` to the `p_dest` array.
+   * 
+   * @pre `p_src` and `p_dest` should be allocated so that they can hold at least 
+   * `n` elements.
+   * @pre Both `p_dest` and `p_src` should be managed by the memory backend of `this`.
+   */  
+  template<typename T, typename I>
+  inline bool copy(T* p_dest, const T* p_src, const I& n) const
```
cnpetra, line 415: please remove this method as well

cnpetra (changes requested):

cnpetra (approved):

PR #602, opened 2023-03-03
Update Image for Spack Action

### LINKED ISSUE #601: Spack CI failures
One of the Spack-based CIs started to fail recently. Namely, `hiop@develop~mpi+raja~shared~kron~sparse`

This happens for all the PRs/branches. It reports 
```
/opt/spack/opt/spack/linux-ubuntu18.04-x86_64/gcc-7.5.0/cmake-3.23.3-
          u7spcudumu4d6olw2rjdhugl6g6z6ndq/bin/cmake: error while loading share
          d libraries: libxml2.so.2: cannot open shared object file: No such fi
          le or directory
```

CameronRutherford:
Closes #601


CameronRutherford:
Seems like we aren't getting any free builds from e4s cache in GitHub actions builds. I have contacted E4S team about those failing builds. I will continue to try and get some re-use from Spack public build caches, although it is unclear what arch they support.

CameronRutherford:
Seems like E4S cache isn't really needed in place of spack build cache, so we might get away with using spack develop.

CameronRutherford:
https://github.com/spack/spack/pull/35751 - we should really run some variation of spack test at the end of these installs to make sure that HiOp is functional. Might require re-working how we have spack-test configured in hiop's package.py.

CameronRutherford:
@nychiang I think we can make an issue out of the need to run `spack test`, but this should be ready to merge.

I think we should use this PR to debug permissions in the call.

nychiang (approved):
LGTM. Thanks Cameron for the quick response! Cheers!

PR #613, opened 2023-04-10
Develop.pcr

rothpc:
This pull request addresses the problem of not completing non-blocking point-to-point communication operations.  It is hypothesized that the lack of completing requests causes some communications libraries to run out of resources for large, long-running runs.  The modifications in this pull request should ensure that any non-blocking operation that is posted will have a matching call to complete it (MPI_Wait or MPI_Test).

The modifications represented by this pull request are intentionally minimal and do not change the fundamental communication approach or implementation.

### DIFF
--- a/src/Optimization/hiopAlgPrimalDecomp.cpp
+++ b/src/Optimization/hiopAlgPrimalDecomp.cpp
@@ -88,20 +88,16 @@ namespace hiop
       int mpi_test_flag; MPI_Status mpi_status;
       int ierr = MPI_Test(&request_, &mpi_test_flag, &mpi_status);
       assert(MPI_SUCCESS == ierr);
-#ifndef NDEBUG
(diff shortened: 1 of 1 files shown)


src/Optimization/hiopAlgPrimalDecomp.cpp
```diff
@@ -88,32 +88,45 @@ namespace hiop
       int mpi_test_flag; MPI_Status mpi_status;
       int ierr = MPI_Test(&request_, &mpi_test_flag, &mpi_status);
       assert(MPI_SUCCESS == ierr);
+      if (mpi_test_flag) {
+        request_ = MPI_REQUEST_NULL;
```
cnpetra, line 92: good idea to set it to null request and (sanity) check it later...

cnpetra (approved):
lgtm

PR #621, opened 2023-05-02
use compound vec in bicg IR

### LINKED ISSUE #630: More changes are required in `hiopCompoundVector`
1. Add comments and function descriptions in `hiopCompoundVector`
2. Rename `hiopCompoundVector` to `hiopVectorCompoundPD`

nychiang:
1. Development of compound vector.
2. use compound vector in bicg IR. 
3. remove unnecessary temporary vectors

CLOSE #630

### DIFF
--- a/src/Optimization/hiopKKTLinSys.cpp
+++ b/src/Optimization/hiopKKTLinSys.cpp
@@ -1655,9 +1655,12 @@ hiopMatVecKKTFullOpr::hiopMatVecKKTFullOpr(hiopKKTLinSys* kkt,
 
 bool hiopMatVecKKTFullOpr::split_vec_to_build_it(const hiopVector& x)
 {
+#ifdef HIOP_DEEPCHECKS
   size_type nx = dx_->get_size();
   size_type neq = dyc_->get_size();
   size_type nineq = dyd_->get_size();
+  assert(x.get_size() == nx+nineq+neq+nineq+nx+nx+nineq+nineq+nx+nx+nineq+nineq);
+#endif
 
   const auto& dir_ir = dynamic_cast<const hiopCompoundVector&>(x);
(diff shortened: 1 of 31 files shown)


src/Optimization/hiopKKTLinSys.cpp
```diff
@@ -2034,20 +2008,22 @@ bool hiopPrecondKKTOpr::combine_dir_to_build_vec(hiopVector& vec)
...
+  dir_ir.getVector(2).copyFrom(*dyc_);
+  dir_ir.getVector(3).copyFrom(*dyd_);
+  dir_ir.getVector(4).copyFrom(*dsxl_);
+  dir_ir.getVector(5).copyFrom(*dsxu_);
+  dir_ir.getVector(6).copyFrom(*dsdl_);
+  dir_ir.getVector(7).copyFrom(*dsdu_);
+  dir_ir.getVector(8).copyFrom(*dzl_);
+  dir_ir.getVector(9).copyFrom(*dzu_);
+  dir_ir.getVector(10).copyFrom(*dvl_);
+  dir_ir.getVector(11).copyFrom(*dvu_);
```
nychiang, line 2025: the order to build the compound vector is `hard-coded`. Should we use unordered_map to avoid this? @cnpetra

nychiang: It's still a draft PR. If we agree to this implementation, I will clean the code and add some comments.

cnpetra: > the order to build the compound vector is `hard-coded`. Should we use unordered_map to avoid this? @cnpetra

I was thinking to use a `PrimalDualVector` class instead of the so-called compound. It may not covered all the use cases though...

cnpetra (approved):

PR #622, opened 2023-05-03
fix memory leaks in inertia-free alg and condensed linsys

nychiang:
Fix some memory leaks detected by valgrind


nychiang:
@cnpetra For regularizations, I print their `infnorm` instead of the `min` and `max` values, since we don't have function `hiopVector::max()`. Please have a look of this change.

nychiang:
@cnpetra @CameronRutherford PNNL CI failed due to out of space. Will rerun this test later.

CameronRutherford:
> @cnpetra @CameronRutherford PNNL CI failed due to out of space. Will rerun this test later.

cc @ryandanehy - we were working on CI in HyKKT, and since all the projects within ExaSGD share the same PNNL account space, HyKKT pipelines ended up running everything out of space. 

Should be fine now...

cnpetra (approved):

PR #627, opened 2023-05-16
Use integer preprocessor macros for version information

tepperly:
I recommend this patch to change the `HIOP_VERSION_MAJOR`, `HIOP_VERSION_MINOR`, and `HIOP_VERSION_PATCH` from string to integer values. As string values, this comparison does not seem to do what I hope it would:
```c++
#if defined(LIDO_USE_IPOPT) || (defined(LIDO_USE_HIOP) && ((HIOP_VERSION_MAJOR > 0) || (HIOP_VERSION_MAJOR == 0) && ((HIOP_VERSION_MINOR > 7) || ((HIOP_VERSION_MINOR == 7) && (HIOP_VERSION_PATCH >= 1)))))
```
Based on what I've read, changing it to do things like `HIOP_VERSION_MAJOR > "0"` won't work either.

You don't really lose anything because you can change an integer value into a string using `# LIDO_VERSION_MAJOR` to change `LIDO_VERSION_MAJOR` into a string.


cnpetra:
I am not opposed to using integer values and lgtm. 

It does not trigger any of the CI for some strange reason. Any idea why?

cnpetra (approved):

PR #634, opened 2023-06-02
Refactor CUSOLVER-LU module and interface

### LINKED ISSUE #629: hiopVectorCuda.hpp should not include hiop_cusolver_defs.hpp
HiOp CUDA vector header file `hiopVectorCuda.hpp` includes `hiop_cusolver_defs.hpp`, which contains definitions of experimental and unsupported GLU solver. These definitions are needed only for cuSOLVER LU interface. It is probably better idea to just:
```c++
#include <cublas_v2.h>
```

pelesh:
Refactor linear solver interface to what we referred to as CUSOLVER-LU. The solver itself is now implemented as a standalone library and renamed ReSolve as it will now have HIP backend, as well. Its cuSOLVER dependency will be optional going forward.

This PR has been verified using standard HiOp tests, as well as with ACOPF examples in ExaGO using 10k-, 25k- and 80k-bus grid models.

This PR implements changes that are needed before the interface that allows receiving data on host or device can be implemented (and temporary class `hiopLinSolverSymSparseReSolveGPU` can be removed).

Also, fixes #629.


pelesh:
I changed names for inner IR options to be generic for any IR implementation. IR now print warning if it fails to converge, but it does not stop the execution.

pelesh:
I reverted back to flat, user-set tolerance for "inner" iterative refinement. I believe all comments are addressed with this latest commit and the PR is ready for merge.

kswirydo (approved):

nychiang (approved):

PR #638, opened 2023-06-23
Issue 637 fix

### LINKED ISSUE #637: update the code after removing deepcheck in the dense linear algebra
The PR about add bicg into quasi-Newton solver removes a deep check.
However, this introduce new computation overhead, e.g., allocate vectors a and b in this function [(timesvec)](https://github.com/LLNL/hiop/blob/develop/src/Optimization/hiopHessianLowRank.cpp#L948-L952C44)

In addition, we expect #626 will happen without deep check.

nychiang:
Fix issue #637 

CLOSE #637

### DIFF
--- a/src/Optimization/hiopHessianLowRank.cpp
+++ b/src/Optimization/hiopHessianLowRank.cpp
@@ -141,6 +141,9 @@ hiopHessianLowRank::hiopHessianLowRank(hiopNlpDenseConstraints* nlp_, int max_me
   _Vmat = V->alloc_clone();
 #endif
 
+  yk = dynamic_cast<hiopVectorPar*>(nlp->alloc_primal_vec());
+  sk = dynamic_cast<hiopVectorPar*>(nlp->alloc_primal_vec());
+
 }  
 
 hiopHessianLowRank::~hiopHessianLowRank()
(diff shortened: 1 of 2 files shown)


cnpetra (approved):

src/Optimization/hiopHessianLowRank.cpp
```diff
@@ -696,20 +706,23 @@ void hiopHessianLowRank::solveWithV(hiopMatrixDense& rhs)
   
   hiopMatrixDense& sol = rhs; //matrix of solutions
   /// TODO: get rid of these uses of specific hiopVector implementation
```
cnpetra, line 708: remove comment since has been addressed

src/Optimization/hiopHessianLowRank.cpp
```diff
@@ -328,14 +340,14 @@ bool hiopHessianLowRank::update(const hiopIterate& it_curr, const hiopVector& gr
     }
 
     //save this stuff for next update
-    _it_prev->copyFrom(it_curr);  _grad_f_prev->copyFrom(grad_f_curr); 
+    _it_prev->copyFrom(it_curr);  _grad_f_prev->copyFrom(grad_f_curr_);
```
cnpetra, line 343: can you please clean up the code a bit. Enforce coding style guidelines, remove variables starting with "_", use `nullptr`. I can fill an issue with these and we can address these later, after we merge this, if this PR urgent.

cnpetra (approved):
looks good, we should clean up this code at some point

src/Optimization/hiopHessianLowRank.cpp
nychiang: Well, it seems too many lines need to be updated.
I just opened a new issue and I will address all the code style in a new PR, so we can see the changes in code style easily in that PR.

PR #642, opened 2023-06-30
apply code style to hiopHessianLowRank

nychiang:
Re-edit `hiopHessianLowRank.hpp` and 'hiopHessianLowRank.cpp`, according to the latest code style guidelines in `README_developers.md`

### DIFF
--- a/src/Optimization/hiopHessianLowRank.cpp
+++ b/src/Optimization/hiopHessianLowRank.cpp
@@ -46,6 +46,13 @@
 // Lawrence Livermore National Security, LLC, and shall not be used for advertising or 
 // product endorsement purposes.
 
+/**
+ * @file hiopHessianLowRank.cpp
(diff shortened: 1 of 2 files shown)


nychiang:
@cnpetra Further clean up can be applied to `hiopHessianLowRank.cpp`. For example, some code have never been used, [see ex1](https://github.com/LLNL/hiop/blob/1cff5a636bb676df4b562689d3a8ade55ca2d5b0/src/Optimization/hiopHessianLowRank.cpp#L2099C9-L2099C9) and [ex2](https://github.com/LLNL/hiop/blob/1cff5a636bb676df4b562689d3a8ade55ca2d5b0/src/Optimization/hiopHessianLowRank.cpp#L2099C9-L2099C9)
You may want to have a look of them, and decide if we should keep them or not.

PR #649, opened 2023-08-03
Add more options to control NLP scaling

### LINKED ISSUE #648: Introduce more user parameters to control NLP scaling
A feature request for more user parameters to control NLP scaling.
These requires the team to add the following two user parameters:
1. nlp_scaling_obj_target_gradient  to control the objective function gradient size.
2. nlp_scaling_constr_target_gradient  to control the constraint gradient size.

nychiang:
In this PR, the following user parameters are introduced:

`scaling_max_obj_grad`: If a positive value is given, the objective of user's NLP will be scaled so that the inf-norm of its gradient is equal to the given value. This value overwrites the value given by `scaling_max_grad`. Default value is 0

`scaling_max_con_grad`: If a positive value is given, the constraints of user's NLP will be scaled so that the inf-norm of its gradient is equal to the given value. This value overwrites the value given by `scaling_max_grad`. Default value is 0

`scaling_min_grad`: If a positive value is given, it is used as the lower bound for the scaling factors. This option has a priority, i.e., the final scaling factor computed must greater or equal to this value, even thought it may violate the values given in `scaling_max_grad`, `scaling_max_obj_grad` and `scaling_max_con_grad`. Default value is 1e-8

CLOSE #648

### DIFF
--- a/doc/src/sections/solver_options.tex
+++ b/doc/src/sections/solver_options.tex
@@ -404,6 +404,15 @@ This option takes double values in $[0, 10^{20}]$ and has a default value $10^{-
 \noindent \textbf{scaling\_max\_grad}: the user's NLP will be rescaled if the inf-norm of the gradient at the starting point is larger than the value of this option. Double values in $[10^{-20}, 10^{20}]$. Default value: $100$.
 \medskip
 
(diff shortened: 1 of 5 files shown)


cnpetra:
we should test this manually to ensure that it does not mess up the duals (and of course brings the derivatives norms at the target).

nychiang:
I did some tests with `NlpSparseEx1`, where one can specify a scaling factor in the command line. 
I played with the scaling factor and 3 options introduced in this PR. I can see that all the problem converges, despite different number of iterations. The worst case is that I got some warnings about big residual from the compressed linear system.

nychiang:
@cnpetra new commit has updated the option descriptions, according to the .tex file

cnpetra (approved):

PR #650, opened 2023-08-11
GPU linear solver interface

pelesh:
This PR creates linear solver interface where ReSolve linear solver can receive data on both CPU and GPU, based on user selection.


pelesh:
CC @cameronrutherford @ryandanehy

kswirydo (approved):
I run it on Ascent with exago, it works as intended.

src/LinAlg/ReSolve/RefactorizationSolver.cpp
```diff
@@ -58,7 +58,6 @@
 #include "RefactorizationSolver.hpp"
 
 #include "hiop_blasdefs.hpp"
```
nychiang, line 60: @pelesh Guess you will remove this line later, once ReSolve is published as a standalone package. Maybe you'd like to address this issue in a later PR.

nychiang (approved):
LGTM. Just one minor comment.

src/LinAlg/ReSolve/RefactorizationSolver.cpp
pelesh: Thanks for catching this. I don't think this include is needed at all.

cnpetra (approved):
Glad this worked out as expected with the cuda backend.

PR #660, opened 2023-09-07
release 1.0

cnpetra:


nychiang:
Everything looks good to me. Just not sure if we want to keep having all the .tex and .bib files in the official release.

nychiang (approved):
Everything looks good to me. Just not sure if we want to keep having all the .tex and .bib files in the official release.

cnpetra:
it's been like that before. any reason to change?

PR #653, opened 2023-08-23
Make compile -Wall/-Werror proof

tepperly:
NY_202309:
This PR replaces PR #646 
CLOSE #646

PR #661, opened 2023-09-12
Stay with C++14

nychiang:
Use C++ standard 14, instead of 17.


nychiang:
@cnpetra @tepperly it seems C++14 works fine with all the changes we added for C++17. #653

cnpetra:
ready for review?

nychiang:
@cnpetra yes.

Issue #670, opened 2023-10-23
`hiop+cuda ^cuda@12` failes to build


cameronrutherford:
https://github.com/spack/spack/issues/40678 - thanks @eugeneswalker for reporting this.

Please refer to spack GitHub issue for full reporting of build failure. From what I can understand, this is the first time I have seen this error.

I speculated that this is code @kswirydo and/or @pelesh know about, but git-blame shows @cnpetra added the lines throwing errors 1+ year ago.

Two questions:
- Anyone have a different hypothesis why this error is happening (apart from `^cuda@12`)?
- Has anyone built with `cuda@12`?

pelesh:
I think @cnpetra is the right person to address this issue. The error comes from CSR matrix class and is related to CUDA-SDK function for CSC->CSR matrix conversion as far as I can see.

pelesh:
> * Anyone have a different hypothesis why this error is happening (apart from `^cuda@12`)?
> * Has anyone built with `cuda@12`?

https://forums.developer.nvidia.com/t/cusparse-csr2csc-agl2-is-missing-at-cuda-12-1/257431

cameronrutherford:
> > * Anyone have a different hypothesis why this error is happening (apart from `^cuda@12`)?
> > * Has anyone built with `cuda@12`?
> 
> https://forums.developer.nvidia.com/t/cusparse-csr2csc-agl2-is-missing-at-cuda-12-1/257431

Thanks @pelesh for finding the docs, and I have edited the issue title to reflect this. Not sure if we can just switch this to the default method without causing other issues - @cnpetra

pelesh:
> > > * Anyone have a different hypothesis why this error is happening (apart from `^cuda@12`)?
> > > * Has anyone built with `cuda@12`?
> > 
> > 
> > https://forums.developer.nvidia.com/t/cusparse-csr2csc-agl2-is-missing-at-cuda-12-1/257431
> 
> Thanks @pelesh for finding the docs, and I have edited the issue title to reflect this. Not sure if we can just switch this to the default method without causing other issues - @cnpetra

As far as I remember one algorithm is fast but inaccurate and the other one is slower but more precise. Changing the algorithm would require a bit of testing to make sure everything still works correctly. This is for @cnpetra to decide.

nychiang:
@cnpetra @cameronrutherford I think this issue has been fixed by PR #671. Please confirm and we can close this issue.

cameronrutherford:
[...]

PR #671, opened 2023-10-24
removed deprecated ALG2 for cusparseCsr2cscEx2

cnpetra:
ALG2 deprecated in CUDA12: switching to ALG1

closes issue #670


cameronrutherford (approved):
LGTM but perhaps there are algorithmic implications of this...

cnpetra:
> LGTM but perhaps there are algorithmic implications of this...

I assume ALG2 was removed because ALG1 (which is the only option) offers better performance. Not sure what you mean by algorithmic implications, everything will work fine with these changes.

PR #681, opened 2024-02-05
Moving limits options 

cnpetra:
Adds two new options to enforce upper limits on the update of the `x` primal variables during the line-search. In particular `moving_lim_abs` and `moving_lim_rel` options are added.


kswartz92:
At first glance this look like exactly what we need, thanks @cnpetra!

nychiang (approved):
LGTM

PR #688, opened 2024-06-05
add new function for warmstart

nychiang:
Not sure if we still need this method.

Remove pipeline tests on PNNL platforms.

@cnpetra


cnpetra:
it is good to have it. A small remark: a more informative name for the new method would be `iterate_internal_callback' or ``iterate_full_callback' '

nychiang:
@cameronrutherford I commented out the PNNL CI for now. We will use it once the problem is fixed.

cnpetra (approved):

PR #689, opened 2024-06-06
Update spack_build.yml to remove E4S cache and use newer OS

cameronrutherford:
Attempts to fix spack pipelines


cameronrutherford:
@nychiang seems like the builds are getting along. We can just merge this into develop instead if you want.

PR #692, opened 2024-08-26
fixed a LNNL typo left over

cnpetra:


nychiang:
@cameronrutherford 
seems to me gcc 7.5 is not supported anymore
[see here](https://github.com/LLNL/hiop/actions/runs/10605563003/job/29394599640?pr=693#step:4:175)

cameronrutherford:
> @cameronrutherford seems to me gcc 7.5 is not supported anymore [see here](https://github.com/LLNL/hiop/actions/runs/10605563003/job/29394599640?pr=693#step:4:175)

You find a new spack bug every day... https://github.com/spack/spack/issues/43919#issuecomment-2318321043

I think this is an issue in using the public buildcache, and things getting out of sync somehow. The only thing that I can consider is to just not use the spack build cache, but that would significantly slow down pipelines. Hopefully there is a quick resolution to the problem and we don't have to change much.

cameronrutherford:
Okay, looked into it a little bit more https://github.com/spack/spack/issues/43919#issuecomment-2318337166

Let me create an issue to support a HiOp GHCR build cache, and we can go from there. I think that should fix the problem long-term, and would bring some added benefit to HiOp as a package as well. People will be able to install a working version of HiOp in just a few minutes, and pipelines will be running faster as well...

EDIT: Turns out there was already an issue!

PR #695, opened 2024-09-03
Add working spack build with build cache.

### LINKED ISSUE #601: Spack CI failures

cameronrutherford:
Closes #601


cnpetra:
FYI, one of the spack ubuntu builds fails.

nychiang:
@cameronrutherford CURL cannot be found from the path [(see here)](https://github.com/LLNL/hiop/actions/runs/10690678532/job/29635606533?pr=695#step:10:172), but I see spack has specified to install curl.

cameronrutherford:
> @cameronrutherford CURL cannot be found from the path [(see here)](https://github.com/LLNL/hiop/actions/runs/10690678532/job/29635606533?pr=695#step:10:172), but I see spack has specified to install curl.

Yes... not really sure what's going on but I haven't had time to try and debug. I'm thinking if we let spack build curl this will resolve as the bug is with using external curl. 

Just need to exclude it the same way we exclude finding Python..

cnpetra:
it loads curl and openssl, but now fails with 
``CMake Error: The source directory "/__w/hiop/hiop" does not appear  to contain CMakeLists.txt. ``

cameronrutherford:
https://github.com/spack/spack/pull/46378 - I thought this would fix the issue. I think it's a reasonable fix to the spack package, but didn't resolve the issue.

What's frustrating here is that the error log isn't printing it's full output. I can't even see the full CMake line. I went down this path because there was an extra `''` in the CMake line, but now I can't even tell what command is being invoked.

GitHub actions further complicates the problem because it's all running inside a VM, so unless I spend money on Azure it's hard to reproduce.

I will take a look again later with a fresh perspective and see what I can figure out. Confusing why we are only seeing this sort of issue with HiOp, not ExaGO or other packages...

cnpetra:
This time fails with 
``CMake Error: The source directory "/home/runner/work/hiop/hiop" does not appear to contain CMakeLists.txt.``
Looks like it is a path concatenation issue.

If it is too cumbersome to make this work, then I suggest to disable this build. All the functionality is tested by the LLNL CI.

cameronrutherford:
> This time fails with `CMake Error: The source directory "/home/runner/work/hiop/hiop" does not appear to contain CMakeLists.txt.` Looks like it is a path concatenation issue.
> 
> If it is too cumbersome to make this work, then I suggest to disable this build. All the functionality is tested by the LLNL CI.

I can't believe I missed it, but we simply weren't cloning the repo in the spack pipelines. The pipelines are really flying here and complete in ~5 minutes

cameronrutherford:
MPI CPU builds are running, passing tests and being pushed to a buildcahe for each branch in your repo. All in under 5 minutes if we don't have to re-build dependencies!

.github/workflows/spack_build.yml
```diff
@@ -0,0 +1,190 @@
...
+          # Once we move submodule deps into spack, we can do some more builds
+          # Also need to change build script to use spack from base image
+          submodules: true
+
+      - name: Clone Spack
+        run: |
+          # TODO: Move back to spack@develop once patch in in develop
+          git clone https://github.com/cameronrutherford/spack.git
+          pushd spack
+          git checkout 838f2b6708a28cae7b1c1de1f1e1f29a1bf2ee0d
```
cameronrutherford:
```suggestion
          git clone https://github.com/spack/spack.git
```

I am pretty sure I never needed to make my spack change...

.github/workflows/spack_build.yml
```diff
@@ -0,0 +1,185 @@
...
+      - name: Check for existing base images
+        run: |
+          set -e
+          CONTAINER_TAG=${{ env.BASE_VERSION }}
+          OCI_URL="${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.BASE_VERSION }}"
+          echo Checking for CONTAINER_TAG $CONTAINER_TAG
+          skopeo inspect \
+              docker://$OCI_URL \
+              --raw \
+              --creds "${{ env.USERNAME }}:${{ secrets.GITHUB_TOKEN }}" \
```
nychiang, line 43: where do we provide the token? 
I'd like to learn more about how you create the user `hiop-bot` and the dependencies in the container, but we can do it offline.

nychiang (approved):
LGTM. It turns out I need to update my knowledge from Cameron.

.github/workflows/spack_build.yml
```diff
@@ -0,0 +1,185 @@
...
+  TMPDIR: /opt/spack-cache
+  # Our repo name contains upper case characters, so we can't use ${{ github.repository }}
+  IMAGE_NAME: hiop
+  USERNAME: hiop-bot
+  BASE_VERSION: ubuntu-24.04-fortran
+
+jobs:
+  base_image_build:
+    runs-on: ubuntu-24.04
+    permissions:
```
cameronrutherford, line 21: @nychiang tokens are automatically created by pipelines with permissions based on this YAML config.

.github/workflows/spack_build.yml
```diff
@@ -0,0 +1,185 @@
...
+env:
+  SPACK_COLOR: always
+  REGISTRY: ghcr.io/llnl
+  SPACK_CACHE: /opt/spack-cache
+  tempdir: /opt/spack-cache
+  TMP: /opt/spack-cache
+  TMPDIR: /opt/spack-cache
+  # Our repo name contains upper case characters, so we can't use ${{ github.repository }}
+  IMAGE_NAME: hiop
+  USERNAME: hiop-bot
```
cameronrutherford, line 15: @nychiang name is arbitrary. You can pick a different name 😁

.github/workflows/spack_build.yml
```diff
@@ -0,0 +1,185 @@
...
+          RUN apt-get update && \
+              apt-get install -y --no-install-recommends \
+                software-properties-common \
+                gpg-agent \
+                openssh-client \
+                openssh-server \
+              && rm -rf /var/lib/apt/lists/*
+          RUN add-apt-repository ppa:ubuntu-toolchain-r/test && \
+              apt-get install -y --no-install-recommends \
+                gfortran \
```
cameronrutherford, line 62: @nychiang gfortran was needed to launch MPI based applications when I was doing this for ExaGO.

I admittedly didn't pull/test the HiOp images, but this is covered in the tutorial I followed https://github.com/spack/spack-tutorial/blob/main/tutorial_binary_cache.rst#creating-runnable-container-images

cameronrutherford: Figuring out the dependencies was admittedly trial and error.

For ExaGo, I was running MPI enabled OPFLOW to get our `--version` output, and logically added to the base image from the error messages I got.

cameronrutherford (commented):
Hopefully this clarifies your question @nychiang

PR #698, opened 2024-11-06
Update spack_build.yml to fix cmake pulling llvm

cameronrutherford:
Fix failing spack pipelines

### DIFF
--- a/.github/workflows/spack_build.yml
+++ b/.github/workflows/spack_build.yml
@@ -99,9 +99,9 @@ jobs:
     strategy:
       matrix:
         spack_spec:
-          - hiop@develop+mpi~raja~shared~kron~sparse ^openmpi 
-          - hiop@develop~mpi~raja~shared~kron~sparse
-          - hiop@develop~mpi+raja~shared~kron~sparse
+          - hiop@develop+mpi~raja~shared~kron~sparse ^openblas ^openmpi 
+          - hiop@develop~mpi~raja~shared~kron~sparse ^openblas
+          - hiop@develop~mpi+raja~shared~kron~sparse ^openblas
 
             # We will need coinhsl for this, but what are the rules for using
             # a coinhsl tarball?
@@ -146,8 +146,6 @@ jobs:
               all:
                 require: "%gcc"
               # https://spack.readthedocs.io/en/latest/packages_yaml.html#setting-requirements-on-virtual-specs
-              blas:
-                require: "openblas"
               cmake:
                 require: "~qtgui~ncurses"
           EOF
--- a/.github/workflows/spack_build.yml
+++ b/.github/workflows/spack_build.yml
@@ -99,9 +99,9 @@ jobs:
     strategy:
       matrix:
         spack_spec:
-          - hiop@develop+mpi~raja~shared~kron~sparse ^openblas ^openmpi 
-          - hiop@develop~mpi~raja~shared~kron~sparse ^openblas
-          - hiop@develop~mpi+raja~shared~kron~sparse ^openblas
+          - hiop@develop+mpi~raja~shared~kron~sparse ^openmpi 
+          - hiop@develop~mpi~raja~shared~kron~sparse
+          - hiop@develop~mpi+raja~shared~kron~sparse
 
             # We will need coinhsl for this, but what are the rules for using
             # a coinhsl tarball?
@@ -129,9 +129,12 @@ jobs:
           cat << EOF > spack.yaml
           spack:
             specs:
-            - ${{ matrix.spack_spec }} ^libevent~openssl target=x86_64_v2
+            - ${{ matrix.spack_spec }} target=x86_64_v2
+            - openblas
+            - libevent~openssl
             concretizer:
-              reuse: dependencies
+              reuse: false
+              unify: true
             config:
               source_cache: $SPACK_CACHE/source_cache
               misc_cache: $SPACK_CACHE/misc_cache
--- a/.github/workflows/spack_build.yml
+++ b/.github/workflows/spack_build.yml
@@ -133,7 +133,7 @@ jobs:
             - openblas
             - libevent~openssl
             concretizer:
-              reuse: false
+              reuse: true
               unify: true
             config:
               source_cache: $SPACK_CACHE/source_cache
--- a/.github/workflows/spack_build.yml
+++ b/.github/workflows/spack_build.yml
@@ -129,7 +129,7 @@ jobs:
           cat << EOF > spack.yaml
           spack:
             specs:
-            - ${{ matrix.spack_spec }} ^openblas ^cmake~qtgui~ncurses ^libevent~openssl target=x86_64_v2
+            - ${{ matrix.spack_spec }} ^libevent~openssl target=x86_64_v2
             concretizer:
               reuse: dependencies
             config:
@@ -141,10 +141,15 @@ jobs:
                 padded_length: False
             mirrors:
               local-buildcache: oci://${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
-            # spack: https://binaries.spack.io/develop
+              spack: https://binaries.spack.io/develop
             packages:
               all:
                 require: "%gcc"
+              # https://spack.readthedocs.io/en/latest/packages_yaml.html#setting-requirements-on-virtual-specs
+              blas:
+                require: "openblas"
+              cmake:
+                require: "~qtgui~ncurses"
           EOF
 
       - name: Configure GHCR mirror
--- a/.github/workflows/spack_build.yml
+++ b/.github/workflows/spack_build.yml
@@ -99,9 +99,9 @@ jobs:
     strategy:
       matrix:
         spack_spec:
-          - hiop@develop+mpi~raja~shared~kron~sparse ^openblas ^openmpi ^libevent~openssl
-          - hiop@develop~mpi~raja~shared~kron~sparse ^openblas ^libevent~openssl
-          - hiop@develop~mpi+raja~shared~kron~sparse ^openblas ^libevent~openssl
+          - hiop@develop+mpi~raja~shared~kron~sparse ^openmpi 
+          - hiop@develop~mpi~raja~shared~kron~sparse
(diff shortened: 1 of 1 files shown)


cameronrutherford:
@nychiang found this potential fix by looking at why llvm was building and disabling relevant package features

.github/workflows/spack_build.yml
```diff
@@ -129,7 +129,7 @@ jobs:
           cat << EOF > spack.yaml
           spack:
             specs:
-            - ${{ matrix.spack_spec }} target=x86_64_v2
+            - ${{ matrix.spack_spec }} ^openblas ^cmake~qtgui~ncurses ^libevent~openssl target=x86_64_v2
```
cameronrutherford: @nychiang seems like my specification of CMake variant worked for the HiOp package, but not for low level packages like openblas.

Do you want me to try and fix this, or can you try and take it from here?

.github/workflows/spack_build.yml
```diff
@@ -141,10 +141,15 @@ jobs:
                 padded_length: False
             mirrors:
               local-buildcache: oci://${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
-            # spack: https://binaries.spack.io/develop
+              spack: https://binaries.spack.io/develop
```
cameronrutherford, line 150: Maybe this is a bad idea

cameronrutherford:
@nychiang now it seems like my openblas specification didn't work properly, so we will have to do a better job enforcing openblas / revert a change I made.

Do you want me to keep debugging this? Are you following what I'm doing/how I'm fixing these issues?

.github/workflows/spack_build.yml
nychiang: I can see what you did change here, and it seems like solving the problem.
However, now it introduce more problems, complaining that some packages are not installed:
 python-venv@1.0/bp54t5w: PackageNotInstalledError: package not installed
These packages have been pre-builded, haven't them?  @cameronrutherford

.github/workflows/spack_build.yml
```diff
@@ -99,9 +99,9 @@ jobs:
     strategy:
       matrix:
         spack_spec:
-          - hiop@develop+mpi~raja~shared~kron~sparse ^openblas ^openmpi ^libevent~openssl
-          - hiop@develop~mpi~raja~shared~kron~sparse ^openblas ^libevent~openssl
-          - hiop@develop~mpi+raja~shared~kron~sparse ^openblas ^libevent~openssl
+          - hiop@develop+mpi~raja~shared~kron~sparse ^openmpi 
+          - hiop@develop~mpi~raja~shared~kron~sparse
```
cameronrutherford, line 103: @nychiang this build is failing because for some reason it's trying to build `libflame`, and not `openblas`.

Looks at the concretization comparison between builds, and you'll see that a 3 hour build would make no sense for the smallest build of the batch here.

cameronrutherford (commented):
I am working on getting the final build to use the right openblas...

.github/workflows/spack_build.yml
cameronrutherford: Why would that spec possibly need python for this build?

.github/workflows/spack_build.yml
```diff
@@ -146,8 +146,6 @@ jobs:
               all:
                 require: "%gcc"
               # https://spack.readthedocs.io/en/latest/packages_yaml.html#setting-requirements-on-virtual-specs
-              blas:
-                require: "openblas"
```
cameronrutherford: I don't understand why this didn't work. It worked for cmake.

I would consider reporting this as a spack bug, but it's near impossible to debug GitHub actions runners sometimes since they run in a special VM

cameronrutherford: This didn't work either... Trying more extreme measures.

.github/workflows/spack_build.yml
```diff
@@ -129,9 +129,12 @@ jobs:
...
           spack:
             specs:
-            - ${{ matrix.spack_spec }} ^libevent~openssl target=x86_64_v2
+            - ${{ matrix.spack_spec }} target=x86_64_v2
+            - openblas
+            - libevent~openssl
             concretizer:
-              reuse: dependencies
+              reuse: false
+              unify: true
```
cameronrutherford: Hopefully this gets openblas to always be built...

.github/workflows/spack_build.yml
nychiang: python is required via
hiop -> amdlibflame -> aocl-utils -> doxygen -> graphviz -> gtkplus -> meson -> python-xxx
why do we need amdlibflame?  Concretization shows that it installs a lot of packages.
@cameronrutherford

.github/workflows/spack_build.yml
cameronrutherford: I got what I wanted... I meant to just have openblas be a part of the concretization, but it appears as though it's always building from source now, and not pulling from the cache. I think reverting to `^openblas` in the spec might fix this... I also am trying to better remember where `libevent` gets pulled in, but I digress.

cameronrutherford:
Guess it's working...

nychiang:
> Guess it's working...

A mystery for me. Thanks a lot!

.github/workflows/spack_build.yml
```diff
@@ -141,10 +144,13 @@ jobs:
...
             mirrors:
               local-buildcache: oci://${{ env.REGISTRY }}/${{ env.IMAGE_NAME }}
-            # spack: https://binaries.spack.io/develop
+              spack: https://binaries.spack.io/develop
             packages:
               all:
                 require: "%gcc"
+              # https://spack.readthedocs.io/en/latest/packages_yaml.html#setting-requirements-on-virtual-specs
+              cmake:
+                require: "~qtgui~ncurses"
```
nychiang: @cameronrutherford how did you figure out these parameters?

.github/workflows/spack_build.yml
```diff
@@ -130,8 +130,11 @@ jobs:
           spack:
             specs:
             - ${{ matrix.spack_spec }} target=x86_64_v2
+            - openblas
+            - libevent~openssl
             concretizer:
-              reuse: dependencies
+              reuse: true
+              unify: true
```
nychiang: @cameronrutherford how did you figure out these?

nychiang (approved):
LGTM. Just some questions.

.github/workflows/spack_build.yml
cameronrutherford: So this was part of my initial debugging.

As we are using `spack@develop`, we are always getting the latest and "greatest" changes. In this case we inherited some changes to CMake that seemed to make it build these options when it wasn't previously.

This forces any build of CMake to have these options disabled.

See [this job's](https://github.com/LLNL/hiop/actions/runs/11710960641/job/32618445922) concretization step for an example of what I mean. Compare that to a later passing pipeline [here](https://github.com/LLNL/hiop/actions/runs/11730809533/job/32679500556). Notice how the updated config results in a much smaller build.

Sure we could probably get the more complicated builds to work too, but a minimal build is faster, and less error prone - we shouldn't build things we don't need.

cameronrutherford: What pointed me to this was noticing the delta in the concretizations. I looked through the logs, noticed that we were having builds fail on a dependency we don't normally build, and then figured out what package was to blame. Then I worked with that package's spack build config to figure out what to enable / disable.

https://packages.spack.io is a great way to get this information quickly.

.github/workflows/spack_build.yml
cameronrutherford: reuse true is just a sane default. It makes sure you always pull a version of a package from a build cache when possible. There are different options as detailed [here](https://spack.readthedocs.io/en/latest/build_settings.html).

I was having trouble with the blas provider specification, so I decided to put it in the core spack spec. Then, by specifying unify to be true, I was able to ensure the version of hiop was also built with the version of blas that I wanted.

.github/workflows/spack_build.yml
```diff
@@ -99,13 +99,16 @@ jobs:
...
+          - hiop@develop+mpi~raja~shared~kron~sparse ^openmpi 
+          - hiop@develop~mpi~raja~shared~kron~sparse
+          - hiop@develop~mpi+raja~shared~kron~sparse
+          # We will need coinhsl for this, but what are the rules for using
+          # a coinhsl tarball?
+          # - hiop@develop~mpi~raja~shared~kron+sparse
+        blas_provider:
+          - openblas
+        compiler:
+          - gcc
```
cameronrutherford, line 111: Feel free to add other variants of your current build, or add compiler / blas specifications as wanted.

Would be easy enough to extend this to build / test with different MPI providers as well.

.github/workflows/spack_build.yml
```diff
@@ -141,10 +146,18 @@ jobs:
...
               all:
-                require: "%gcc"
+                require: "%${{ matrix.compiler }}"
+              # https://spack.readthedocs.io/en/latest/packages_yaml.html#setting-requirements-on-virtual-specs
+              cmake:
+                # We don't need CMake GUI features for minimal container builds
+                require: "~qtgui~ncurses"
+              libevent:
+                # Building OpenSSL was causing errors
+                require: "~openssl"
```
cameronrutherford, line 160: I was finding spack's `^` syntax in the spec to not always specify variants as I wanted. This is a more sustainable way to always enforce certain options on desired pacakges.

cameronrutherford:
Some final notes:

- Disable the spack public mirror as you have issues. Normally issues are reported and fixed relatively soon after you notice them, but feel free to ping the spack slack / make an issue
- Checking for external packages and concretizing are still the longest parts of the jobs - this is possible to speed up, but the tradeoff is that you wouldn't be getting rolling spack updates
- Feel free to extend these builds to other platforms / GPU runners / MPIs / compilers / OSs as you find convenient
- Logging is currently quite verbose. You can minimize / maximize it as needed
- The GHCR registry you have configured might need to be cleared out once every so often. Re-generating it from scratch isn't so bad, just will have initial builds take a while

PR #703, opened 2025-01-24
use C++17

nychiang:
Set C++17 as the standard in CMake files


cnpetra (approved):

cnpetra:
@nychiang can you please confirm `CMAKE_CXX_STANDARD` is configurable from command line and spack?

CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
 cmake_minimum_required (VERSION 3.18)
 
-set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++17 -lstdc++fs ")
+
```
pelesh: I am not sure if this is portable. It will probably not work with MSVC, for example.

Kitware [suggests](https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#requiring-language-standards) adding language standards to targets such as
```cmake
target_compile_features(HIOP PRIVATE cxx_std_17)
```
My understanding is that the built-in target `cxx_std_17` is supposed to add correct flags for the compiler of your selection.

nychiang: @pelesh  We are worrying about if this PR will affect ExaGO, as it uses `c++14`. [see here](https://github.com/pnnl/ExaGO/blob/12d296801161327a788d74dc1ef241e79ae09496/CMakeLists.txt#L11C1-L11C27)

pelesh: I doubt it is going to be an issue, because ExaGO only links to HiOp.

nychiang: after checking others' code, e.g., RAJA, MFEM, etc..., I don't think I need to specify `-std=c++17` or `target_compile_features`. Just pushed another commit for review.

PR #706, opened 2025-02-01
Use the latest RAJA and UMPIRE

nychiang:
Update HiOp to use the latest versions for RAJA and UMPIRE.

Note that `HIOP_USE_RESOLVE` is switched off since it fails ALL the cuda-related tests, i.e., even the `testVectors` fails.  For example, function `vectorSetToZero` fails with the error message 
```
terminate called after throwing an instance of 'thrust::system::system_error'
  what():  parallel_for failed: cudaErrorInvalidDeviceFunction: invalid device function
```

However, when I switch off `HIOP_USE_RESOLVE`, nothing wrong except for two unit tests under `NlpSparseRajaEx2`.
It is expected as these two unit tests need to use `ReSolve` [(see here)](https://github.com/LLNL/hiop/blob/develop/src/Drivers/Sparse/CMakeLists.txt#L86C1-L89C8)

Therefore this PR also corrects the Cmake file. Now these two examples are created only if `ReSolve` is given.

@pelesh Have you ever seen this kind of error? Does `ReSolve` require some specific cusolver packages?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -236,6 +236,8 @@ if(HIOP_USE_GPU)
     enable_language(CUDA)
     check_language(CUDA)
 
+    set(CMAKE_CUDA_HOST_COMPILER g++)
+
     if(NOT DEFINED CMAKE_CUDA_STANDARD)
       set(CMAKE_CUDA_STANDARD 14)
       set(CMAKE_CUDA_STANDARD_REQUIRED ON)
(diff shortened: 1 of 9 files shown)


pelesh:
> @pelesh Have you ever seen this kind of error? Does `ReSolve` require some specific cusolver packages?

I haven't seen this error before and Re::Solve (HiOp or standalone version) is well tested with CUDA 11.7. 

Off the top of my head, this error is common when one tries to run GPU code out of queue. It is also caused when you don't build library with matching compute capability (e.g. build with 70 and run on Pascal). Does your scheduler allocates a matching device? I'll try to reproduce this on my end.

nychiang:
> > @pelesh Have you ever seen this kind of error? Does `ReSolve` require some specific cusolver packages?

I was in an interactive session on a debugging node. In the same session, I can: 1)compile and run HiOp without Re::Solve; 2) then failed to run it with Re::Solve; 3) and then run it without Re::Solve. Maybe we can setup a quick teams meeting in which I can show you the problem.

cnpetra (approved):

PR #707, opened 2025-02-06
Cmake option HIOP_TEST_WITH_SLURM

thartland:
Including an option `HIOP_TEST_WITH_SLURM` in the Cmake file in order that the tests can be run using `srun` instead of `mpirun` on machines running slurm schedulers.


cnpetra:
this was long overdue, thanks! have you tested it on any of the LC machines?

thartland:
Yes, I tested on ruby last night. All make tests passed having built with the option.

pelesh:
Alternatively,you could just let user specify the command to run tests rather than offering a few options. @cameronrutherford did something like that for [ExaGO](https://github.com/pnnl/ExaGO/blob/develop/CMakeLists.txt#L388). The idea back then was to implement similar in HiOp.

cnpetra (approved):

PR #708, opened 2025-02-10
Update README.md

cnpetra:
removed some content that is not applicable anymore; added Tucker as an author


pelesh:
While you are at it, you may want to remove stanza on custom CUDA installation from the README file. The workaround described there pertains to older versions of CMake and it is probably misleading for anyone with a reasonably recent CMake. See [here](https://cmake.org/cmake/help/latest/module/FindCUDAToolkit.html) for more recent CMake target names for CUDA libraries.

thartland (approved):
Looks good. Updates to bib reference info, contributors and copyright.

nychiang (approved):

thartland (approved):

PR #709, opened 2025-02-12
improved log output for NLP scaling

cnpetra:
Logs (for verbosity>=4) info on the internal NLP scaling such as which scaling method and some scaling scalars.


nychiang (approved):
Just curious about what brings you here to modify this log output. :-)

cnpetra:
> Just curious about what brings you here to modify this log output. :-)

ill-scaled problem :)

PR #711, opened 2025-03-11
Add Bayesian Optimization into HiOp

nychiang:
This PR adds a GP-based Bayesian Optimization (BO) solver into HiOp.

- [x] Implementation of a abstract GP class and an inherited class
- [x] Implementation of a BO algorithm
- [x] Implementation of an example


src/Drivers/hiopbbpy/BODriver3.py
cnpetra, line 1: name should be `BoDriver.py` ?

src/Drivers/hiopbbpy/lp_problem.py
```diff
@@ -0,0 +1,25 @@
+"""
+Implementation of the LP problem class
```
cnpetra: need better description. I would change the name  to `LpNormProblem`. The name of the file should be `LpNormProblem.py`

cnpetra (changes requested):
see the comments. I think we should merge this without updating the user manual.

src/Drivers/hiopbbpy/lp_problem.py
thartland: The file name/class names have been changed.

thartland (approved):
I think this is a great first time and don't have any big concerns that prevent me from approving this but do have a few questions. Are the smtProblems omitted in order to reduce the amount of code in hiopbbpy? Should we change the name from hiopbbpy to hioppy? While hiopbbpy is intended for use in black box BO optimization. If we eventually make hiop algorithms available in the hiopbbpy interface then it seems that a user could just use of the hiop algorithms in any python application.

cnpetra (approved):

PR #714, opened 2025-03-17
spack v1.0 support: `%foo +bar` -> `+bar %foo`

haampie:
Spack v1.0 will parse `pkg %gcc +foo` as "pkg with direct dependency gcc
with variant foo enabled" instead of "pkg with variant foo enabled and compiler
gcc". Reorder as `pkg +foo %gcc` to be compatible with Spack v0.x and
v1.0.

### DIFF
--- a/scripts/platforms/marianas/spack.yaml
+++ b/scripts/platforms/marianas/spack.yaml
@@ -1,6 +1,6 @@
 spack:
   specs:
-  - hiop%gcc@10.2.0@develop+cuda+deepchecking+sparse+kron+cusolver_lu+ginkgo+raja cuda_arch=60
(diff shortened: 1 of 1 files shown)


cnpetra (approved):

nychiang (approved):

cnpetra:
does not pass one of the test, we need to look into it

nychiang:
I changed the base to a new branch, and we are fine to merge this PR to that new branch.
Later I will file a PR to merge that branch to develop. Sorry for the duplicated work, but some CI jobs need to be activated from the LLNL domain, i.e., not from an external folk.

@haampie  It would also be great if you could file this PR from the LLNL domain, without forking the repository.

Issue #712, opened 2025-03-14
Update hiopBBPy


cnpetra:
- [x] new acquisition fcn: EI 
- [x] bounds - make sure the code works; have a test problem with bounds
- [x] more example problems: Branin and others (including all used by SMT)
- [x] CI for python code
- [x] add SMT as dependency (`pyproject.toml`)

cnpetra:
closed by PR#715

PR #715, opened 2025-03-19
Add EI and Branin example in

nychiang:
This is the first PR to address some issues listed in #712
More examples and CI will be added in another PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["setuptools", "wheel", "numpy"]
+requires = ["setuptools", "wheel", "numpy", "smt", "cyipopt"]
 build-backend = "setuptools.build_meta"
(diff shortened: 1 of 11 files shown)


thartland:
I added a `*-1` in the definition of the EI acquistion function. You made the correct changes in going from maximizing `f` to minimizing `-f` but since we minimize the acquisiton function it needs an additional scaling, see for example in smt [here](https://github.com/SMTorg/smt/blob/master/smt/applications/ego.py#L382).

thartland (commented):
I think the issue we discussed earlier has been resolved. Go ahead and test. Let me know when this is ready for review.

thartland (approved):
This looks good to me. I previously tested the code prior to it being ready for review. Thanks!

cnpetra (approved):
lgtm. I think we need tests and CI asap

Issue #721, opened 2025-04-08
Asynchronous task parallelization


cnpetra:
Asynchronous task parallelization of (true) objective function using existing schedulers (Python thread, process, and MPI executors) and possibly Tucker's (simplistic) task scheduler.

thartland:
The work in progress branch [hiopbbpy-v0.05-dev](https://github.com/LLNL/hiop/tree/hiopbbpy-v0.05-dev) aims to address these issues.

thartland:
hiopbbpy v0.05 dev ([here](https://github.com/LLNL/hiop/pull/724)) was rescoped to only cover batched Bayesian optimization routines. hiopbbpy v0.06  ([here](https://github.com/LLNL/hiop/tree/hiopbbpy-v0.06dev)) aims to address this issue.

cnpetra:
I think further work has been done in addition to batched BO (such as multi-start and initial sampling) and this issue is resolved, @thartland please confirm

thartland:
I confirm this issue has been resolved.

Issue #720, opened 2025-04-08
Batched BO


cnpetra:
Batched algorithm for BO, ideally one algorithm that works for both EI and LCB acquisition functions.

- [x] identify one or two popular approaches
- [ ] implement using shared-memory parallelism (Python thread pool executor) --> deferred to Issue #721

thartland:
The work in progress branch [hiopbbpy-v0.05-dev](https://github.com/LLNL/hiop/tree/hiopbbpy-v0.05-dev) aims to address these issues.

cnpetra:
closed by pr #724

PR #722, opened 2025-04-08
Add support to use ipopt to solve constrained problem

nychiang:
- [ ] Support constrained optimization problems --- linear constraint only!
- [ ] Compute 1st order information for EI and LCB acquisition functions.
- [ ] Pass 1st order information to scipy optimization solver.
- [ ] Add support to IPOPT.
- [ ] Provide new examples.
- [ ] Redesign some classes 

Right now CI fails as expected, since ipopt is not installed in the container.
I cannot find preinstalled ipopt from the github maketplace, and hence I'd like to swith to conda and install cyipopt and ipopt under conda.


src/Drivers/hiopbbpy/BODriverEX.py
thartland, line 1: Is the default solver in ipopt a quasi-Newton method when Hessian information is not provided? If the default option is quasi-Newton then the infastructure seems ready for problems with nonlinear constraints. If the default option is not quasi-Newton then we should make it our default option in `optproblem.py`.

src/hiopbbpy/opt/optproblem.py
thartland, line 1: Since here, the constraints are either `c = 0` or `c >= 0`, it might be worth noting that here (or somewhere else) in the comments.

src/Drivers/hiopbbpy/BODriverEX.py
nychiang: I had checked it, and ipopt uses quasi-newton since Hessian is not provided. 
If I set option `hessian_approximation` to `exact`, ipopt will throw errors and terminates.
yes, supporting nonlinear constraints is not hard, but we also need to change the way scipy loads the problem.

thartland: Well good. It seems we don't want to use full-Newton methods at the moment. I guess we can talk about the problem of linear vs nonlinear constraints at a later time.

thartland (approved):
This looks good to me.

src/hiopbbpy/opt/boalgorithm.py
```diff
@@ -159,29 +194,41 @@ def optimize(self, prob:Problem):
...
+      #else:
+      #    print(f"Optimal at BO iteration: {self.idx_opt-n_init_sample+1} ")
           
       print(f"Optimal point: {self.x_opt.flatten()}, Optimal value: {self.y_opt}")
       print()
 
-
-
 # Find the minimum of the input objective `fun`, using the minimize function from SciPy. 
 def pyminimize(fun, x0, method, bounds, constraints, options):
```
cnpetra: I would remove `py` prefix and rename the function to `minimizer` or similar.

Also update the comment on the previous line to reflect the addition of Ipopt

cnpetra (approved):
lgtm

one other comment would be related to name `IpoptProbFromScipy` . I think just `IpoptProb` would do it for now.

finally, `user_constraints` are attached to BOAlgorithm. Wouldn't make more sense to be attached to the `problem` instead?

src/hiopbbpy/opt/boalgorithm.py
nychiang: @cnpetra The name `IpoptProb` is used for the `cyipopt` module, see [here](https://github.com/LLNL/hiop/blob/1731af766639525cfe9b0c0f4076a8fd176edf22/src/hiopbbpy/opt/boalgorithm.py#L12C32-L12C41). Therefore I need to find another name for our class.
That module and ours are used later at[ L226-L227](https://github.com/LLNL/hiop/blob/1731af766639525cfe9b0c0f4076a8fd176edf22/src/hiopbbpy/opt/boalgorithm.py#L226-L227)

nychiang: How about using `NlpProb` for `cyipopt`, and change ours to `IpoptProb`?

nychiang:
CI is fixed!

src/hiopbbpy/opt/boalgorithm.py
nychiang: See my latest commit. I put cyipopt into our class `IpoptProb`.

nychiang:
@cnpetra @thartland  I just push a commit, in which:
1) I hide cyipopt in our class `IpoptProb`
2) check Ipopt return value and msg --- currently there is a mismatch between the return value and return msg. It might caused by the version of Ipopt, and I just commented in a relevant issue in cyipopt.
3) terminate hiop if Ipopt failes to find one solution from all the initial points.

thartland:
@nychiang 

> @cnpetra @thartland I just push a commit, in which:
> 
>     1. I hide cyipopt in our class `IpoptProb`
> 
>     2. check Ipopt return value and msg --- currently there is a mismatch between the return value and return msg. It might caused by the version of Ipopt, and I just commented in a relevant issue in cyipopt.
> 
>     3. terminate hiop if Ipopt failes to find one solution from all the initial points.

I do not see the same mismatch you are seeing. I the result of `info.get('status', -1)` is `0` and the `status_msg` indicates successful convergence of the interior-point solver.

thartland:
It is reasonable to terminate the BO loop if the optimizer for the acquisition function fails to find a single local minimizer from the multiple starting points.

nychiang:
> I do not see the same mismatch you are seeing. I the result of `info.get('status', -1)` is `0` and the `status_msg` indicates successful convergence of the interior-point solver.

I figured this out. I should not use -1 as the default value, as it indicates [maximum iterations exceed](https://github.com/coin-or/Ipopt/blob/8400b4342d456743f2de6c22ddd3f88deb30d40b/src/Interfaces/IpReturnCodes_inc.h#L23). I will change it to -999 in the upcoming commit.

src/hiopbbpy/opt/boalgorithm.py
nychiang: constraints have been moved to the problem class.

thartland:
This looks great! I have some preliminary work started but once this is merged I'll start a new merge request to update hiopbbpy according to issues #720 and #721.

PR #724, opened 2025-04-15
Hiopbbpy v0.05 dev

thartland:
This PR adds capabilities to hiopbbpy for batched Bayesian optimization.


thartland:
I am including some plots here which constitutes the tests of the added capabilities. I modified the BODriverEX.py to one without constraints besides the bound constraints. I ran 10 BO iterations using with a batch size of 2 and using the various strategies. I repeated experiments 10 times and compare against mean behavior of EI with a batch size of 1. The error bars show the variance in the repeated experiments. The horizontal axis is the number of batches, so there are twice as many function evaluations with `q=2` then with `q=1` (see plot legend).

Note: the number of function evaluations done in the BO loop will be equal to $q \cdot \text{number of BO batches}$.

CLmin
KB
KBLB
KBRand
KBUB

src/Drivers/hiopbbpy/BODriverCI.py
```diff
@@ -103,7 +103,7 @@
          max_obj[prob_type][acq_type] = max(max_obj[prob_type][acq_type], y_opt[prob_type][acq_type][n_repeat])
          min_obj[prob_type][acq_type] = min(min_obj[prob_type][acq_type], y_opt[prob_type][acq_type][n_repeat])
 
-#if(num_repeat >= 1000 ):
+#if(num_repeat == 1000 ):
 #   np.save("yopt_20iter_1000run.npy", y_opt)
```
nychiang: I think we can remove these two lines. (we can do it in the next release, i.e., v0.0.6)

src/Drivers/hiopbbpy/BODriverEX.py
```diff
@@ -57,9 +57,9 @@ def con_jac_ineq(x):
 for prob_type in prob_type_l:
    print()
    if prob_type == "LpNorm":
-      problem = LpNormProblem(nx, xlimits, constraints=None)
+      problem = LpNormProblem(nx, xlimits)
    else:
-      problem = BraninProblem(constraints=None)
+      problem = BraninProblem()
    problem.set_constraints(user_constraint)
```
nychiang, line 63: I have two ways to add the constraints:
1. in the constructor, setting `constraints=user_constraint`
or
2. using function `set_constraints`.
I should leave a comment in this example.

src/hiopbbpy/opt/boalgorithm.py
```diff
@@ -175,31 +195,39 @@ def optimize(self):
...
-          y_new = self.prob.evaluate(np.atleast_2d(x_new))
+             # Get a virtual point
+             y_virtual = self._get_virtual_point(np.atleast_2d(x_new))
 
-          # Update training set
-          x_train = np.vstack([x_train, x_new])
-          y_train = np.vstack([y_train, y_new])
+             # Update training set with the virtual point
+             x_train         = np.vstack([x_train,         x_new    ])
+             y_train_virtual = np.vstack([y_train_virtual, y_virtual])
```
nychiang: if `batch_size == 1`, we don't need to compute the virtual point, do we?
No matter which `batch_type` is used, the virtual point only works with batch_size>=2, right?

```
x_new = self._find_best_point(x_train, y_train)
if self.batch_size > 1:
    y_train_virtual = y_train.copy()
    for batch in range(self.batch_size):
         # Get a virtual point
         y_virtual = self._get_virtual_point(np.atleast_2d(x_new))
         # Update training set with the virtual point
         x_train         = np.vstack([x_train,         x_new    ])
         y_train_virtual = np.vstack([y_train_virtual, y_virtual])
             
         x_new = self._find_best_point(x_train, y_train_virtual)
```

src/hiopbbpy/problems/problem.py
```diff
@@ -5,19 +5,19 @@
             Nai-Yuan Chiang <chiang7@llnl.gov>
 """
 import numpy as np
+import collections.abc
```
nychiang, line 8: is `collections` a new dependency?

nychiang (changes requested):
some minor comments and questions.

src/hiopbbpy/problems/problem.py
thartland: `collections` is a built in module that is not pip installable. See [here](https://stackoverflow.com/questions/45738414/how-to-install-collections-in-python-3-6).

src/Drivers/hiopbbpy/BODriverEX.py
thartland: Yes, I understand that there are two methods to set constraints. However, passing `None` as the constraints to `BraninProblem` is equivalent to instantiating a `BraninProblem` with no arguments since the default `constraints` argument of `BraninProblem` is `None.` Or at least it was until I recently changed the default to `[ ]`.

src/Drivers/hiopbbpy/BODriverCI.py
thartland: These lines have been removed.

src/hiopbbpy/opt/boalgorithm.py
thartland: This has now been address. The code was fine as is but it is not strictly necessary to call `_get_virtual_point` at the final point in the batch.

cnpetra (approved):
looks great!

thartland:
@nychiang -- I believe I have addressed all of the changes that you requested. Let me know if there is anything left for me to update.

nychiang (approved):

PR #726, opened 2025-04-25
HiOp-v0.0.7 --- support of Nonlinear constraints.

nychiang:
- add support to solver `scipy` optimization solver `trust-constr`
- add support to vectorized constraints, for solver `IPOPT` and `trust-constr`

Right now hiopbb support three optimization solvers:
1. `SLSQP` from `scipy`. Constraints for `SLSQP` are defined as a list of dictionaries. Vectorized constraint is not allowed, i.e., the implementation of a constraint function can only return a scalar, and the corresponding Jacobian implementation returns a vector.
3. `trust-constr` from `scipy`. Constraints for `SLSQP` are defined as a single object or a list of objects of `NonlinearConstraint` specifying constraints to the optimization problem. Vectorized constraint is allowed, i.e., the implementation of a constraint function can return a vector, and the corresponding Jacobian implementation returns a matrix.
4. `IPOPT`. Constraints can be defined in either way, i.e., 1) a list of dictionaries where each dictionary defines one constraint; or 2) a dictionary that include two function pointers, `cons` and `jac`, where both functions support vectorized implementation. 

[See scipy's user guid for more details about the constraints used in SLSQP and trust-constr.](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html)


src/hiopbbpy/opt/optproblem.py
```diff
@@ -16,38 +16,48 @@
...
-                self.cu.append(0.0)
-            elif con['type'] == 'ineq':
-                self.cl.append(0.0)
-                self.cu.append(np.inf)
-            else:
-                raise ValueError(f"Unknown constraint type: {con['type']}")
+        if isinstance(self.cons, list):
+            # constraints is provided as a list of dict, supported by SLSQP and Ipopt
+            for con in self.cons:
+                if con['type'] == 'eq':
```
thartland, line 40: What if the `con` dictionary does not have key `type`?

src/hiopbbpy/opt/optproblem.py
```diff
@@ -16,38 +16,48 @@
...
+                    self.cl.append(0.0)
+                    self.cu.append(0.0)
+                elif con['type'] == 'ineq':
+                    self.cl.append(0.0)
+                    self.cu.append(np.inf)
+                else:
+                    raise ValueError(f"Unknown constraint type: {con['type']}")
```
nychiang, line 47: @thartland  see here

thartland: That guards against the case where the constraint dictionary has key `type` but the value associated to the key is not `eq` or `ineq`. I was suggesting that we do something if the constraint dictionary is provided but it does not have key `type` e.g., if the user erroneously used something like
```
 user_constraint_list = [{ 'fun': con_eq,   'jac': con_jac_eq},
                   {'fun': con_ineq, 'jac': con_jac_ineq}]
```
I suggest adding to see if the key `type` is in the dictionary. `if 'type' not in con:`...

src/hiopbbpy/opt/optproblem.py
nychiang: I added a function to check it. Note that this is also checked by SLSQP.

src/hiopbbpy/opt/optproblem.py
nychiang: yes, I just added this check by a helper function, in a new commit.

src/hiopbbpy/opt/optproblem.py
thartland: Great!

thartland (approved):
Good addition

nychiang:
@thartland Is your 0.0.6 ready to PR merge? Or should I rename it as 0.0.6 and yours be 0.0.7?

thartland:
@nychiang -- my 0.0.6. is waiting on a meeting to coordinate how to proceed. You can change the naming of mine and yours as this PR will be ready for my 0.0.6.

Issue #727, opened 2025-04-25
Compilation issues in LiDO codevelop build


tepperly:
I recently upgraded my Linux system to Fedora Core 41, and the new compiler is clang 19.1.7. During the LiDO co-develop build, it's giving errors about variable length arrays being a clang C++ extension. For example:
```sh
samu: job failed: /usr/bin/clang++  -I/usr/lib64/gfortran/modules/mpich -I/home/userid/current/lido/lido-2.0/build/clang-codevelop/hiop -I/home/userid/current/lido/lido-2.0/hiop/src/Interface -I/home/userid/current/lido/lido-2.0/hiop/src/ExecBackends -I/home/userid/current/lido/lido-2.0/hiop/src/Optimization -I/home/userid/current/lido/lido-2.0/hiop/src/LinAlg -I/home/userid/current/lido/lido-2.0/hiop/src/Utils -I/home/userid/current/lido/lido-2.0/hiop/src/_Internals -isystem /usr/include/mpich-x86_64 -fPIC -Wall -Werror -std=c++14 -fPIC -MD -MT hiop/src/Utils/CMakeFiles/hiopUtils.dir/MathKernelsHost.cpp.o -MF hiop/src/Utils/CMakeFiles/hiopUtils.dir/MathKernelsHost.cpp.o.d -o hiop/src/Utils/CMakeFiles/hiopUtils.dir/MathKernelsHost.cpp.o -c /home/userid/current/lido/lido-2.0/hiop/src/Utils/MathKernelsHost.cpp
In file included from /home/userid/current/lido/lido-2.0/hiop/src/Utils/MathKernelsHost.cpp:62:
/home/userid/current/lido/lido-2.0/hiop/src/Utils/hiopCppStdUtils.hpp:80:12: error: variable length arrays in C++ are a Clang extension [-Werror,-Wvla-cxx-extension]
   80 |     T temp[n];
      |            ^
/home/userid/current/lido/lido-2.0/hiop/src/Utils/hiopCppStdUtils.hpp:80:12: note: function parameter 'n' with unknown value cannot be used in a constant expression
/home/userid/current/lido/lido-2.0/hiop/src/Utils/hiopCppStdUtils.hpp:78:77: note: declared here
   78 |   template<class T> inline void reorder(T *arr, std::vector<int> index, int n)
      |                                                                             ^
1 error generated.
[...]

PR #728, opened 2025-04-25
Address hiop issue #727

tepperly:
This removes a use of [variable length arrays](https://en.wikipedia.org/wiki/Variable-length_array) in a hiop header file. VLA's are part of C, but clang 19 is reporting them as a clang C++ extension as the feature isn't in the standard. This PR also improves the method signature for `reorder()` to avoid copying the `index` argument.

hiop has other uses of VLA's, but clang only emits warnings rather than errors for those. Hence, I haven't made any changes to address those potential issues.


tepperly:
There are other instances of variable length arrays in hiop (e.g., line 493 of hiopMatrixDenseRowMajor.cpp), but it seems that it's only giving me an error for the ones in header files.

While I was editing this file, I was also tempted to make this change, but it's not related to issue #727. 
```diff
$ git diff
diff --git a/src/Utils/hiopCppStdUtils.hpp b/src/Utils/hiopCppStdUtils.hpp
index 0ae1748..479dda9 100644
--- a/src/Utils/hiopCppStdUtils.hpp
+++ b/src/Utils/hiopCppStdUtils.hpp
@@ -12,7 +12,7 @@
 #include <numeric>
 #include <cassert>
 #include <math.h>
-#include <chrono>
+#include <random>
 
 namespace hiop
 {
@@ -85,7 +85,7 @@ inline void reorder(T* arr, const std::vector<int> &index, int n)
 static inline unsigned long generate_seed()
 {
 #ifdef NDEBUG
-  return std::chrono::system_clock::now().time_since_epoch().count();
+  return std::random_device{}();
 #else
   return 0;
 #endif
```
On Linux systems, `std::random_device` which draws from potentially true random sources such as hardware random number generators in the CPU, /dev/random, or /dev/urandom. It should have more entropy that the system clock.

tepperly:
As far as I can tell, the build failures may be related to the fact that I'm not part of the development team. Perhaps someone on the hiop team can check. I can't tell from the output what I need to do to fix it.

cnpetra:
@tepperly just gave you write access so now you should be able to create branches. It would help if you could move your fork to a branch since it would trigger additional, LC CIs and the ones that appear to have failed or didn't run.

Feel free to update the branch to use `std::random_device`.

src/Utils/hiopCppStdUtils.hpp
```diff
@@ -74,18 +74,12 @@ static inline void toupper(std::string& str_in) { std::transform(str_in.begin(),
...
 
   // arr[i] should be present at index[i] index
-  for(int i = 0; i < n; i++) temp[i] = arr[index[i]];
-
-  // Copy temp[] to arr[]
-  for(int i = 0; i < n; i++) {
-    arr[i] = temp[i];
-    //       index[i] = i;
-  }
+  for(int i = 0; i < n; ++i) arr[i] = temp[index[i]];
```
cnpetra, line 82: fyi, lately we are enforcing braces (K&R style) for one-liners in selection or loop blocks.

cnpetra: only applies to new or edited code :)

tepperly: We use clang-format to forcefully make the style uniform in all of LiDO. I think it's terrific.

Issue #683, opened 2024-03-08
Derivative checker


cnpetra:
Ideally having a method/function that does the numerical finite differencing checks for general functions so that can be reused in different contexts (optimization solver, nonlinear systems solver, complementarity solver, etc.)

- [x] first-order obj gradient and constraints Jacobian
- [ ] second-order

PR #731, opened 2025-05-12
derivative checker 

cnpetra:
Tests user-provided derivatives using finite differences. Currently only first-order derivatives are checked. Hessians will be finished in a follow up PR. 

Other issues addressed 
 - clarified relationship between `relax_bounds` and `elastic_mode`
 - added a disabler of the NLP transformations in NLP problem formulations to allow finite differencing directly on user functions
 - bugs and some code cleaning related to handling of fixed variables
 - added getters and setters for vector classes.

Note: Hessian checks will not be supported for MDS problems. 

Partially addresses issue #683

### DIFF
--- a/src/LinAlg/hiopVectorRaja.hpp
+++ b/src/LinAlg/hiopVectorRaja.hpp
@@ -329,7 +329,6 @@ public:
    */
   virtual double get_elem(const index_type& idx_global) const;
 
-  
   const ExecSpace<MEMBACKEND, EXECPOLICYRAJA>& exec_space() const { return exec_space_; }
   
 private:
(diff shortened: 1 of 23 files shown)


src/LinAlg/hiopVectorRaja.hpp
```diff
@@ -318,7 +340,7 @@ class hiopVectorRaja : public hiopVector
   // and transfers within and from `exec_space_host_` work with EXECPOLICYHOST (currently all such
   // combinations work).
   using EXECPOLICYHOST = hiop::ExecPolicySeq;
-  ExecSpace<MEMBACKENDHOST, EXECPOLICYHOST> exec_space_host_;
+  mutable ExecSpace<MEMBACKENDHOST, EXECPOLICYHOST> exec_space_host_;
```
nychiang, line 343: When we need to change `exec_space_host_`?

nychiang (approved):
LGTM.

src/LinAlg/hiopVectorRaja.hpp
cnpetra: device to host transfer in `get_elem` (on const objects)

PR #732, opened 2025-05-19
fix a bug in calling different samplers

nychiang:
- fix a bug in calling different samplers
- update some parameters


cnpetra (approved):

thartland (approved):
Looks good to me.

PR #733, opened 2025-06-13
Adds evaluation manager to utils

weslleyspereira:


thartland (approved):
Great addition! I have tested it on another application code. In a separate PR I will add the MPIEvaluator wrapper of the EvaluationManager and utilize it for various tasks in a BO loop.

thartland:
@nychiang -- the driver script BODriverEx now uses the MPIEvaluator wrapper of Weslley's EvaluationManager, see [here](https://github.com/weslleyspereira/hiop/blob/add-parallel-eval-manager/src/Drivers/hiopbbpy/BODriverEX.py#L84). See [here](https://github.com/weslleyspereira/hiop/blob/add-parallel-eval-manager/src/hiopbbpy/utils/util.py#L67-L69) for instructions on how to run the code with multiple mpi processes.

src/hiopbbpy/opt/boalgorithm.py
```diff
@@ -187,65 +187,65 @@ def set_options(self, solver_options):
...
-
-      print(f"\n\nOptimal at BO iteration: {self.idx_opt+1} ")
-      #if self.idx_opt < n_init_sample:
-      #    print(f"Optimal at initial sample: {self.idx_opt+1}")
-      #else:
-      #    print(f"Optimal at BO iteration: {self.idx_opt-n_init_sample+1} ")
-          
-      print(f"Optimal point: {self.x_opt.flatten()}, Optimal value: {self.y_opt}")
-      print()
+        x_train = self.xtrain
```
cnpetra: Is there any reason there are 4 whitespaces used for alignement in this function?

thartland: No, there is not a reason. I'll change it back shortly to be consistent.

thartland: @cnpetra -- actually hiopbbpy has alignment inconsistencies, see [here](https://github.com/LLNL/hiop/blob/develop/src/hiopbbpy/opt/boalgorithm.py#L185-L190). I saw this while reviewing this PR and made the change so it would be consistent.

cnpetra: should be exactly two, see  [here](https://github.com/LLNL/hiop/blob/develop/README_developers.md#indent-with-spaces)

cnpetra (approved):

PR #735, opened 2025-07-11
fix array size

nychiang:
fix one issue about mismatched array dimension caused by introducing evaluator manager (PR 733)


cnpetra:
ready to merge (if passes tests) ?

nychiang:
> ready to merge (if passes tests) ?
Yes.

PR #734, opened 2025-07-02
Parallel Evaluator Upgrade

thartland:
This PR is a work in progress but aims to add evaluators to take advantage of other embarrassingly parallel portions of a typical BO algorithm -- initial sampling and instantiating a local optimizer on a batch of random initial points.

PR #736, opened 2025-07-15
use LLNL-CI for hiopbbpy

nychiang:
Use LLNL gitlab-CI to run pipeline tests for hiopbbpy

Update the script for existing LLNL-CI pipelines. Now the workflow is 
Screenshot 2025-07-15 at 9 42 16 AM

### DIFF
--- a/.gitlab/llnl-ci.yml
+++ b/.gitlab/llnl-ci.yml
@@ -149,19 +149,19 @@ test_on_dane:
   extends:
     - .dane
     - .llnl_script_test
-  needs: ['build_on_dane']
+  needs: ['build_on_dane_hiopbbpy']
 
 build_on_dane_hiopbbpy:
   extends:
     - .dane
     - .llnl_build_python
-  needs: ['config_on_dane']
+  needs: ['build_on_dane']
 
 test_on_dane_hiopbbpy:
   extends:
     - .dane
     - .llnl_test_hiopbbpy
-  needs: ['build_on_dane_hiopbbpy']
+  needs: ['test_on_dane']
 
 clean_on_dane:
   extends:
--- a/.gitlab/llnl-ci.yml
+++ b/.gitlab/llnl-ci.yml
@@ -112,7 +112,8 @@ variables:
       echo "CUSTOM_CI_BUILDS_DIR = ${CUSTOM_CI_BUILDS_DIR}"
       set -xv
       cd "$CUSTOM_CI_BUILDS_DIR"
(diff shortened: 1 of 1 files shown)


nychiang:
@thartland I’ve just moved the pipelines to LLNL GitLab. Everything is working well overall, but I’m currently unable to run pipeline jobs in parallel on Dane — one of the tasks intermittently fails due to user authentication issues. As a temporary workaround, I’ve configured the jobs to run sequentially.

I’ll update the setup to enable parallel execution once I hear back from LC support. In the meantime, feel free to add your tests to the pipeline.

thartland:
@nychiang great! Perhaps, this PR should build on PR #734 as right now it might not make sense to add the additional test. I have a sequential test of MPIEvaluator (when run without MPI) in said PR, see [here](https://github.com/LLNL/hiop/blob/hiopbbpy-v0.08-dev/.github/workflows/test_hiopbbpy.yml#L65-L70). Also, given that there are some things to work out with LC the timing might make sense to get PR #734 in and then build on that.

nychiang:
I fixed the credential problem with the help of LC. Here’s what the workflow looks like now:
Screenshot 2025-07-15 at 9 42 16 AM

nychiang:
@thartland 
I suggest keeping this PR as a standalone one since it rearranges the workflow for the existing Gitlab CI, and your sequential test works fine on Github workflow. I can see we may need another PR to add parallel python jobs into LLNL-Gitlab CI. Particularly, I got the following error when I tried to run the serial code `BODriverEX.py` in a parallel way, and one AI-suggested solution shows that we may need to change the source code of hiopbbpy:

Screenshot 2025-07-15 at 10 05 39 AM

Screenshot 2025-07-15 at 10 06 44 AM

nychiang:
@thartland In other words, additional issues may come up when we add pipeline tests for parallel Python tasks.

thartland:
@nychiang -- I see. We can use ThreadPoolExecutor as the executor. I updated MPIEvaluatorCI.py in the PR that I mentioned previously in this discussion thread so that it runs with ThreadPoolExecutor.

There are some `from mpi4py.futures import MPIPoolExecutor` calls in the source code but only does so if the call to `is_running_with_mpi()` call returns true. We can figure out at a later time. 

I agree these PRs can be made separately and then we can do more fine grained CI later.

thartland (approved):
Look good!

cnpetra (approved):

PR #739, opened 2025-08-02
Add logger for hiopbbpy

### LINKED ISSUE #717: HiOpBB logging
consider instrument the code with https://docs.python.org/3/library/logging.html 

need an "verbosity" option that controls the output

nychiang:
implement a logger using the code from https://docs.python.org/3/library/logging.html

CLOSE #717

### DIFF
--- a/src/hiopbbpy/opt/boalgorithm.py
+++ b/src/hiopbbpy/opt/boalgorithm.py
@@ -267,6 +267,7 @@ class minimizer_wrapper:
   # Find the minimum of the input objective `fun`, using the minimize function from SciPy. 
   def minimizer_callback(self, x0s):
     output = []
+    msg = ""
     for x0 in x0s:
       if self.method == "SLSQP":
         if 'grad' in self.fun:
--- a/src/hiopbbpy/opt/boalgorithm.py
+++ b/src/hiopbbpy/opt/boalgorithm.py
@@ -132,7 +132,7 @@ class BOAlgorithm(BOAlgorithmBase):
     self.logger.info(f"Internal optimization solver: {opt_solver}")
     self.logger.info(f"Internal optimization solver options: {self.solver_options}")
     self.logger.info(f"Initial training set: {xtrain.shape[0]} samples, {xtrain.shape[1]} dimensions")
-    self.logger.info(f"Bounds: {self.bounds}")
+    self.logger.debug(f"Bounds on optimization variable: {self.bounds}")
     self.logger.info(f"Logger level: {logger_level}")
 
   # Method to train the GP model
@@ -195,7 +195,7 @@ class BOAlgorithm(BOAlgorithmBase):
     self.logger.scalars(
         f"  Acquisition values: min = {y_min:.4e}, mean = {y_mean:.4e}, max = {y_max:.4e}"
     )
-    self.logger.debug(f"Chosen candidate: {best_xopt}")
+    self.logger.debug(f"Estimated optimal point x: {best_xopt}")
 
     return best_xopt
   
@@ -256,6 +256,11 @@ class BOAlgorithm(BOAlgorithmBase):
 
           # Update training set with the virtual point
           y_train_virtual = np.vstack([y_train_virtual, y_virtual])
+
+        mean_val = self.gpsurrogate.mean(np.array([x_new])).item()
+        var_val = self.gpsurrogate.variance(np.array([x_new])).item()
+        self.logger.scalars(f"  Acqusition mean at new sample x: {mean_val}")
+        self.logger.scalars(f"  Acqusition var at new sample x: {var_val}")
       
       y_new = self.obj_evaluator.run(self.prob.evaluate, x_train[-self.batch_size:])
       y_new = np.array(y_new)
@@ -264,7 +269,7 @@ class BOAlgorithm(BOAlgorithmBase):
       self.logger.scalars(f"Training set size is now {x_train.shape[0]}")
       self.logger.iterations(f"Current best objective: {np.min(y_train):.4e} "
                              f"(previous best: {prev_best:.4e})")
-      self.logger.scalars(f"Improvement: {prev_best - np.min(y_train):.4e}")
+      self.logger.scalars(f"Objective function improvement: {prev_best - np.min(y_train):.4e}")
 
       # Save the new sample points and objective evaluations
       for j in range(1, self.batch_size+1):
--- a/src/hiopbbpy/opt/boalgorithm.py
+++ b/src/hiopbbpy/opt/boalgorithm.py
@@ -82,20 +82,19 @@ class BOAlgorithm(BOAlgorithmBase):
 
     logger_level = options.get('log_level', "INFO")
     self.logger.setlevel(logger_level)
-    self.logger.info(f"Problem name: {prob.name}")
 
     self.bo_maxiter = options.get('bo_maxiter', self.bo_maxiter)
-    assert self.bo_maxiter > 0, f"Invalid bo_maxiter: {self.bo_maxiter }"
-    self.logger.info(f"bo_maxiter: {self.bo_maxiter}")
+    assert self.bo_maxiter > 0, f"Invalid bo_maxiter: {self.bo_maxiter}"
+
+    self.n_start = options.get('n_start', self.n_start)
+    assert self.n_start > 0, f"Invalid n_start: {self.n_start}"
 
     acquisition_type = options.get('acquisition_type', "LCB")
     assert acquisition_type in ["LCB", "EI"], f"Invalid acquisition_type: {acquisition_type}"
-    self.logger.info(f"acquisition_type: {acquisition_type}")
 
     batch_size = options.get('batch_size', 1)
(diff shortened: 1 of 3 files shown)


nychiang:
here is one example of the log:

hiopbbpy Problem name: LpNormProblem
hiopbbpy Max BO iter: 5
hiopbbpy Optimizing acquisition (LCB) with 10 random initial points
hiopbbpy Batch type: KB
hiopbbpy Batch size: 2
hiopbbpy Internal optimization solver: IPOPT
hiopbbpy Internal optimization solver options: {'max_iter': 200, 'print_level': 1}
hiopbbpy Initial training set: 5 samples, 2 dimensions
hiopbbpy Bounds: [(-5, 5), (-5, 5)]
hiopbbpy Logger level: info
hiopbbpy *****************************
hiopbbpy Iteration 1/5
hiopbbpy In batch 1/2
hiopbbpy Start finding the best sampling point:
hiopbbpy   Acquisition optimization finished with 10 successes, 0 failures
hiopbbpy   Acquisition values: min = -4.0166e-01, mean = -4.0166e-01, max = -4.0166e-01
hiopbbpy In batch 2/2
hiopbbpy Start finding the best sampling point:
hiopbbpy   Acquisition optimization finished with 10 successes, 0 failures
hiopbbpy   Acquisition values: min = 4.1435e-01, mean = 1.7229e+00, max = 2.5907e+00
hiopbbpy Training set size is now 7
hiopbbpy Current best objective: 1.8002e+00 (previous best: 2.3277e+00)
hiopbbpy Improvement: 5.2747e-01
hiopbbpy *****************************
...
hiopbbpy *****************************
hiopbbpy Iteration 5/5
hiopbbpy In batch 1/2
hiopbbpy Start finding the best sampling point:
hiopbbpy   Acquisition optimization finished with 10 successes, 0 failures
hiopbbpy   Acquisition values: min = 1.7660e+00, mean = 1.7971e+00, max = 1.8910e+00
hiopbbpy In batch 2/2
hiopbbpy Start finding the best sampling point:
hiopbbpy   Acquisition optimization finished with 10 successes, 0 failures
hiopbbpy   Acquisition values: min = 1.7675e+00, mean = 1.7763e+00, max = 1.8014e+00
hiopbbpy Training set size is now 15
hiopbbpy Current best objective: 1.7875e+00 (previous best: 1.7875e+00)
hiopbbpy Improvement: 0.0000e+00
hiopbbpy ===================================
hiopbbpy Bayesian Optimization completed
hiopbbpy Total evaluations: 10
hiopbbpy Optimal at BO iteration: 4 
hiopbbpy Best value: [1.78751014]
hiopbbpy ===================================

src/hiopbbpy/opt/boalgorithm.py
```diff
@@ -111,17 +115,36 @@ def __init__(self, prob:Problem, gpsurrogate:GaussianProcess, xtrain, ytrain,
...
+    self.logger.info(f"Problem name: {prob.name}")
+    self.logger.info(f"Max BO iter: {self.bo_maxiter}")
+    self.logger.info(f"Optimizing acquisition ({self.acquisition_type}) "
+                     f"with {self.n_start} random initial points")
+    self.logger.info(f"Batch type: {self.batch_type}")
+    self.logger.info(f"Batch size: {batch_size}")
+    self.logger.info(f"Internal optimization solver: {opt_solver}")
+    self.logger.info(f"Internal optimization solver options: {self.solver_options}")
+    self.logger.info(f"Initial training set: {xtrain.shape[0]} samples, {xtrain.shape[1]} dimensions")
+    self.logger.info(f"Bounds: {self.bounds}")
```
thartland: @nychiang Could we say something more descriptive than `Bounds: `? Something like: `Bounds on optimization variable:`

src/hiopbbpy/opt/boalgorithm.py
```diff
@@ -132,27 +155,48 @@ def _find_best_point(self, x_train, y_train, x0 = None):
...
+    best_xopt = x_all[np.argmin(y_all)]
+    y_min, y_max, y_mean = np.min(y_all), np.max(y_all), np.mean(y_all)
+
+    self.logger.scalars(
+        f"  Acquisition optimization finished with {len(y_all)} successes, {n_failures} failures"
+    )
+    self.logger.scalars(
+        f"  Acquisition values: min = {y_min:.4e}, mean = {y_mean:.4e}, max = {y_max:.4e}"
+    )
+    self.logger.debug(f"Chosen candidate: {best_xopt}")
```
thartland: How about `Estimated optimal point x` instead of  `Chosen candidate`?

src/hiopbbpy/opt/boalgorithm.py
```diff
@@ -190,16 +234,19 @@ def optimize(self):
...
-      print(f"Iteration {i+1}/{self.bo_maxiter}")
+      self.logger.critical(f"*****************************")
+      self.logger.critical(f"Iteration {i+1}/{self.bo_maxiter}")
+
+      prev_best = np.min(y_train)
 
       y_train_virtual = y_train.copy() # old training + batch_size num of virtual points
       for j in range(self.batch_size):
         # Get a new sample point
+        self.logger.scalars(f"In batch {j+1}/{self.batch_size}")
```
thartland, line 245: This is not the batch number but rather the `j`th point of a given batch. `In batch {j+1}/{self.batch_size}` is misleading.

src/hiopbbpy/opt/boalgorithm.py
```diff
@@ -214,32 +261,43 @@ def optimize(self):
...
     self.y_opt = self.y_hist[self.idx_opt]
     self.setTrainingData(x_train, y_train)
 
-    print(f"\n\nOptimal at BO iteration: {self.idx_opt+1} ")
-    print(f"Optimal point: {self.x_opt.flatten()}, Optimal value: {self.y_opt}")
-    print()
+    self.logger.critical("===================================")
+    self.logger.critical("Bayesian Optimization completed")
+    self.logger.critical(f"Total evaluations: {len(self.y_hist)}")
+    self.logger.critical(f"Optimal at BO iteration: {self.idx_opt//self.batch_size+1} ")
```
thartland, line 302: When the batch_size is $1$ I think it makes more sense to say "Optimal at BO iteration: {self.idx_opt}/{self.bo_maxiter}". When the batch_size is greater than $1$ I think we need to be more careful as a batch of points are added per BO iteration and it's more complicated. E.g, if batch size is $5$ then the first $5$ points in `self.y_hist` are obtained in the first BO iteration and so self.idx_opt being $2$ doesn't mean that the optimal point was obtained in the second BO iteration.

src/hiopbbpy/opt/boalgorithm.py
```diff
@@ -214,32 +261,43 @@ def optimize(self):
       y_new = np.array(y_new)
       y_train = np.vstack([y_train, y_new])
       
+      self.logger.scalars(f"Training set size is now {x_train.shape[0]}")
+      self.logger.iterations(f"Current best objective: {np.min(y_train):.4e} "
+                             f"(previous best: {prev_best:.4e})")
+      self.logger.scalars(f"Improvement: {prev_best - np.min(y_train):.4e}")
```
thartland: I suggest something more descriptive than `Improvement: `, maybe `Objective function improvement:` or something similar.

thartland (changes requested):
See inline comments below for requests of more descriptive logging messages. Overall great addition!

src/hiopbbpy/opt/boalgorithm.py
nychiang: I updated it to the option you suggested. Additionally, I’m using logger.debug instead of logger.info, since printing high-dimensional variables is not recommended by Cosmin.

src/hiopbbpy/opt/boalgorithm.py
nychiang: I don't get this point. `j in range(self.batch_size)` --> should j mean the batch number? in my example above, this line prints the lines of
hiopbbpy In batch 1/2
hiopbbpy In batch 2/2

src/hiopbbpy/opt/boalgorithm.py
nychiang: I don't get this.
if batch_size = 5, and idx_opt = 2, it means the optimal point was obtained in the 1st BO iteration, doesn't it?
The equation used here is a floor division, and hence {self.idx_opt//self.batch_size+1}  = 2//5+1 = 0+1 = 1.

src/hiopbbpy/opt/boalgorithm.py
thartland: Perfect. Thanks for clearing up my confusion.

src/hiopbbpy/opt/boalgorithm.py
thartland: Thanks for clearing up my confusion. Thanks.

thartland (approved):
Looks great. Thanks @nychiang

cnpetra (approved):

PR #744, opened 2025-08-29
fix logger 

nychiang:
previously, if the best observation y is occurred at the initial points, the logger output wrong information, as it only reads/writes the values from the BO iterations.
I have corrected this issue.

In this PR, I also set sb=yes for ipopt.

### DIFF
--- a/src/Drivers/hiopbbpy/BODriverEX.py
+++ b/src/Drivers/hiopbbpy/BODriverEX.py
@@ -87,7 +87,7 @@ if __name__ == "__main__":
       problem = LpNormProblem(nx, xlimits)
     else:
       problem = BraninProblem()
-    problem.set_constraints(user_constraint_list) # for solver 'trust-constr' and IPOPT, use user_constraint_dict; for olver 'SLSQP' and IPOPT, user_constraint_list
+    problem.set_constraints(user_constraint_list) # for solver 'trust-constr' and IPOPT, use user_constraint_dict; for solver 'SLSQP' and IPOPT, user_constraint_list
    
     for acq_type in acq_type_l:
       ### initial training set
--- a/src/Drivers/hiopbbpy/BODriverEX.py
+++ b/src/Drivers/hiopbbpy/BODriverEX.py
@@ -48,7 +48,8 @@ def con_ineq(x):
 def con_jac_ineq(x):
   return  np.array([1.0, -1.0])
 
-# 'SLSQP' requires constraints defined in a list of dict
+# 'SLSQP' requires constraints defined in a list of dict.
+# IPOPT can suport this format, too
 user_constraint_list = [{'type': 'eq',   'fun': con_eq,   'jac': con_jac_eq},
                    {'type': 'ineq', 'fun': con_ineq, 'jac': con_jac_ineq}]
 
@@ -72,7 +73,7 @@ def cons_jac_vec(x):
 cl = -np.inf * np.ones(3)
 cu = np.zeros(3)
 
-# 'trust-constr' method supports vector-valued constraints
+# 'trust-constr' and IPOPT support vector-valued constraints
 user_constraint_dict = {'cons': cons_vec, 'jac': cons_jac_vec, 'cl': cl, 'cu': cu}
 
 
@@ -86,7 +87,7 @@ if __name__ == "__main__":
       problem = LpNormProblem(nx, xlimits)
     else:
       problem = BraninProblem()
-    problem.set_constraints(user_constraint_dict) #use user_constraint_dict or user_constraint_list
+    problem.set_constraints(user_constraint_list) # for solver 'trust-constr' and IPOPT, use user_constraint_dict; for olver 'SLSQP' and IPOPT, user_constraint_list
    
     for acq_type in acq_type_l:
       ### initial training set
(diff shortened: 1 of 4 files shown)


thartland:
Are we sure that the initial samples used to train the GP for the objective will satisfy the constraints of the problem? Because if so, then we need to check if the initial training points actually satisfy the constraints and exclude them as a potential best point if they do not satisfy the constraints.

nychiang:
> Are we sure that the initial samples used to train the GP for the objective will satisfy the constraints of the problem? Because if so, then we need to check if the initial training points actually satisfy the constraints and exclude them as a potential best point if they do not satisfy the constraints.

Great catch! The initial points satisfy only the boundary constraints but not the common NLP constraints. I will switch back to reporting only the best observation from the iterates to guarantee feasibility. My question now is: do we need to enforce feasibility for all the initial training samples? I did this for the wind farm problem, but in general it may be difficult to ensure in practice. If feasibility is not required for the initial samples, could the GP surrogate be misled by these infeasible points?

thartland:
> > Are we sure that the initial samples used to train the GP for the objective will satisfy the constraints of the problem? Because if so, then we need to check if the initial training points actually satisfy the constraints and exclude them as a potential best point if they do not satisfy the constraints.
> 
> Great catch! The initial points satisfy only the boundary constraints but not the common NLP constraints. I will switch back to reporting only the best observation from the iterates to guarantee feasibility. My question now is: do we need to enforce feasibility for all the initial training samples? I did this for the wind farm problem, but in general it may be difficult to ensure in practice. If feasibility is not required for the initial samples, could the GP surrogate be misled by these infeasible points?

Perhaps we can check which of the initial training points actually satisfy the NLP constraints and then only include the associated objective values as potential "best" objective values. 

The GP surrogate can be a surrogate of the objective not just on the manifold defined by the NLP constraints, so it seems to include training points that don't satisfy the NLP constraints. I don't immediately see how the BO would get fooled by the GP surrogate trained on points outside the feasible domain defined by the NLP constraints.

src/hiopbbpy/opt/boalgorithm.py
```diff
@@ -227,18 +227,16 @@ def set_options(self, solver_options):
...
     y_train = self.ytrain
-    self.logger.iterations(f"Best objective from {np.size(x_train, 0)} initial samples: {np.min(y_train):.4e} ")
+    self.logger.iterations(f"Best UNCONSTRAINED objective from {np.size(x_train, 0)} initial samples: {np.min(y_train):.4e} ")
 
     self.x_hist = []
     self.y_hist = []
     
-    min_in_init_samples = True
+    prev_best_y = np.inf
     for i in range(self.bo_maxiter):
```
thartland, line 252: In the first iterate `i=0` we can determine which of the training points are feasible, so that we can use them to report a best feasible objective.

cnpetra:
> > > Are we sure that the initial samples used to train the GP for the objective will satisfy the constraints of the problem? Because if so, then we need to check if the initial training points actually satisfy the constraints and exclude them as a potential best point if they do not satisfy the constraints.
> > 
> > 
> > Great catch! The initial points satisfy only the boundary constraints but not the common NLP constraints. I will switch back to reporting only the best observation from the iterates to guarantee feasibility. My question now is: do we need to enforce feasibility for all the initial training samples? I did this for the wind farm problem, but in general it may be difficult to ensure in practice. If feasibility is not required for the initial samples, could the GP surrogate be misled by these infeasible points?
> 
> Perhaps we can check which of the initial training points actually satisfy the NLP constraints and then only include the associated objective values as potential "best" objective values.
> 
> The GP surrogate can be a surrogate of the objective not just on the manifold defined by the NLP constraints, so it seems to include training points that don't satisfy the NLP constraints. I don't immediately see how the BO would get fooled by the GP surrogate trained on points outside the feasible domain defined by the NLP constraints.

I agree with @thartland, ok to include in GP the initial samples that do not satisfy the constraints

src/Drivers/hiopbbpy/BODriverEX.py
```diff
@@ -86,7 +87,7 @@ def cons_jac_vec(x):
       problem = LpNormProblem(nx, xlimits)
     else:
       problem = BraninProblem()
-    problem.set_constraints(user_constraint_dict) #use user_constraint_dict or user_constraint_list
+    problem.set_constraints(user_constraint_list) # for solver 'trust-constr' and IPOPT, use user_constraint_dict; for olver 'SLSQP' and IPOPT, user_constraint_list
```
thartland: Typo `; for olver 'SLSQP' and IPOPT`, `olver` --> `solver`

thartland:
Fixed a typo and having `if_feasible` return an array of True booleans if there are no NLP constraints.

thartland:
This now works with IPOPT when the constraint set is a list and SLSQP when the constraint set is a list.

However, things fail when using either trust-constr, or IPOPT when the constraint set is a dict. There is the error

```
Traceback (most recent call last):
  File "/home/tucker/software/ai4sci/hiop_logger/src/Drivers/hiopbbpy/BODriverEX.py", line 117, in <module>
    bo.optimize()
  File "/home/tucker/anaconda3/envs/hiopbb2/lib/python3.9/site-packages/hiopbbpy/opt/boalgorithm.py", line 280, in optimize
    feas_new = self.prob.if_feasible(x_train[-self.batch_size:])
  File "/home/tucker/anaconda3/envs/hiopbb2/lib/python3.9/site-packages/hiopbbpy/problems/problem.py", line 123, in if_feasible
    y = self.con_evaluate(x)
  File "/home/tucker/anaconda3/envs/hiopbb2/lib/python3.9/site-packages/hiopbbpy/problems/problem.py", line 87, in con_evaluate
    con_eval = self.constraints['cons'](x)  # shape (n_samples, n_con)
  File "/home/tucker/software/ai4sci/hiop_logger/src/Drivers/hiopbbpy/BODriverEX.py", line 57, in cons_vec
    x1, x2 = x
ValueError: not enough values to unpack (expected 2, got 1)
```
I believe the issue is that the `x` passed into the constraint functions in the course of optimization is the tuple `np.array([x1, x2]) ` but the x passed to it during the feasibility check is an array of tuples.

thartland:
Looks good!

thartland (approved):
Looks good!

cnpetra (approved):

Issue #741, opened 2025-08-12
Preparations for release v1.2.0


cnpetra:
hiopBBpy related updates (add items as needed)
- [x] - short description of solver in README.md and pointers to additional doc: installation, usage, interface, parallel runs (T)
- [x] - user manual update (as per in-person discussion) (T)
- [x] - logging
- [x] - merge PR #710 (C)
- [x] - make Ipopt optional, including the build system (NY)
- [x] change the updated date in pdf. some references are missing in user manual (C)
- [x] revisit user manual (recent additions): bump ver number, capitalizations, etc.
- [x] test non-MPI build, including derivative checker (C)

PR #745, opened 2025-09-03
Set ipopt as an optional solver in hiopbbpy

nychiang:
CyIpopt is treated as an optional dependency.
The hiopbbpy package can function without cyipopt installed.
If a user explicitly invokes Ipopt-specific functionality, the code attempt to import cyipopt and raise a clear runtime error if it is not available.

Address one item in #741


thartland (approved):
Looks good. I tested that things run smoothly on a hiopbbpy build that does not include cyipopt.

thartland:
Things also run fine in a python environment with cyipopt and hiop. Specifically when we make use of ipopt as the acquisition function optimizer.

nychiang:
> Things also run fine in a python environment with cyipopt and hiop. Specifically when we make use of ipopt as the acquisition function optimizer.

Thanks for giving it a try. I also tested this branch with two different venv, with or without cyipopt, and both work fine.

cnpetra:
we should say somewhere in documentation (and readme) that Ipopt can be used as well (and say it requires cyipopt -- maybe a bit on venv)

cnpetra:
> we should say somewhere in documentation (and readme) that Ipopt can be used as well (and say it requires cyipopt -- maybe a bit on venv)

Definitely should update user manual to make it clear Ipopt is optional. Maybe also provide additional info that we support venv.

nychiang:
> > we should say somewhere in documentation (and readme) that Ipopt can be used as well (and say it requires cyipopt -- maybe a bit on venv)
> 
> Definitely should update user manual to make it clear Ipopt is optional. Maybe also provide additional info that we support venv.
@thartland has already written a section on setting up Ipopt and cyipopt, which I slightly revised to emphasize that this is an optional feature. For the Python virtual environment, I kept the instructions minimal since creating a venv is standard practice for Python users and can be done in multiple ways — for example, I prefer Python’s built-in venv module, while Andrew and Tucker prefer conda. Users can also install all required packages directly with the system Python, without creating a venv. For this reason, I don’t think we should mention that our package supports venv — it’s more of a user environment choice than a package feature.

thartland:
> > > we should say somewhere in documentation (and readme) that Ipopt can be used as well (and say it requires cyipopt -- maybe a bit on venv)
> > 
> > 
> > Definitely should update user manual to make it clear Ipopt is optional. Maybe also provide additional info that we support venv.
> > @thartland has already written a section on setting up Ipopt and cyipopt, which I slightly revised to emphasize that this is an optional feature. For the Python virtual environment, I kept the instructions minimal since creating a venv is standard practice for Python users and can be done in multiple ways — for example, I prefer Python’s built-in venv module, while Andrew and Tucker prefer conda. Users can also install all required packages directly with the system Python, without creating a venv. For this reason, I don’t think we should mention that our package supports venv — it’s more of a user environment choice than a package feature.

I agree that venv support is not needed. If a python package, such as HiOpBBpy, is installable then it can be used within venv and conda environments.

cnpetra (approved):

PR #750, opened 2025-10-02
Fix LLNL CI

nychiang:
add CI on matrix


nychiang:
Given the reply from LC "Unfortunately, given how matrix is setup, we cannot support a ci queue on matrix. We do have a CI queue on rzvector which has GPU nodes that you could use.", I think we can close this MR.
In short, currently there is NO CZ GPU machine that supports a CI queue. 
@cnpetra

PR #751, opened 2026-01-06
Add support to unconstrained problems + fix logger

nychiang:
1. Support unconstrained problems. 
2. prevent double logger issue when hiopbbpy module is used in a notebook.


thartland (approved):
Looks good to me.

cnpetra (approved):

PR #753, opened 2026-03-13
match definition name with implementation of logBarrierWeighted_local

chapman39:
Fixes the following error:

```
hiopVectorHip.cpp:682:23: error: out-of-line definition of 'logBarrier_local' does not match any declaration in 'hiop::hiopVectorHip'
```

This appears to happen in hiop version 1.2 and beyond


cnpetra (approved):
Thanks for the fix. Not sure how I missed the signature, but I'm glad it's fixed

chapman39:
@cnpetra thanks! and when you have a chance, i have spack-packages PR with this patch https://github.com/spack/spack-packages/pull/3841

nychiang (commented):
LGTM

PR #756, opened 2026-03-31
Parallel performance profiler

nychiang:
This PR adds instrumentation for performance profiling of parallel evaluations done with EvaluationManager. Parallel performance statistics are printed after each chunck of parallel blocks is finished/retrieved.

Also fixes a name of the input arguments for a SMT function that was causing a CI pipeline to fail.


cnpetra (approved):

thartland (approved):
Looks great.

PR #757, opened 2026-04-10
Use New Evaluation Manager

nychiang:
Adopt Weslley's new EavaluationManager.

### DIFF
--- a/src/hiopbbpy/utils/util.py
+++ b/src/hiopbbpy/utils/util.py
@@ -28,6 +28,10 @@ import numpy as np
 from .evaluation_manager import EvaluationManager
 import logging
 
+import os
+import time
+import uuid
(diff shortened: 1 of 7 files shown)


src/hiopbbpy/utils/util.py
nychiang, line 68: @thartland do we still need to have this wrapper?

thartland: The previous purpose of the MPIEvaluator was to have a light wrapper on an MPI EvaluationManager in order that we can just submit tasks to it. If the new EvaluationManager has the functionality of the MPIEvaluator then I don't see why we need the MPIEvaluator. It would be nice though if there was a base Evaluator class that something like the new EvaluationManager would derive from. I'll think more about this.

PR #760, opened 2026-05-15
flushed output and randseed fixing

cnpetra:
- fixes rand sees for deterministic runs (also uses midpoint as starting point for local upper bound solver)
- flushes output (use with `python -u`)
- fixes an indexing problem for the new eq constraint under opt_mode 3 ([this](https://github.com/llnl/hiop/blob/84c7c7cfa0e797c4c8e9d8e11d0de9fad3872ccc/src/hiopbbpy/opt/bnbalgorithm.py#L542) constraint should run up to `ntrain`) and removes the other redundant constraints. Convex solver performs much better, very few failures.
- some changes to convex solver status handling
- more robust GP fitting (hyper_opt="Cobyla") makes BnB work much better and more predictably over BO iterations

### DIFF
--- a/src/Drivers/hiopbbpy/BnBBoDriverEX.py
+++ b/src/Drivers/hiopbbpy/BnBBoDriverEX.py
@@ -44,8 +44,8 @@ class QuadraticShift(Problem):
 
 class PeriodicObjective(Problem):
   """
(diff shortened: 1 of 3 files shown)


thartland (approved):
Looks great. Will be doing testing now.

PR #764, opened 2026-08-03
Add turbo solver and MuyGPyS surrogate backend

jwang125:
This PR adds a preliminary trust region BO implementation and an interface with MuyGPyS. There are several documented choices that will be modified and updated in the future. For instance, to incorporate MuyGPyS, we only use diagonal correlation matrix approximation to sample. Turbo uses Thompson sampling by default and currently is the only available acquisition function. Issues will be filed separately.


src/hiopbbpy/opt/turbo.py
```diff
@@ -0,0 +1,378 @@
...
+"""
+
+import math
+import numpy as np
+from scipy.stats import qmc
+from ..surrogate_modeling.gp import GaussianProcess
+from ..problems.problem import Problem
+from ..utils.util import Evaluator
+from .boalgorithm import BOAlgorithmBase
+#from hiopbbpy.surrogate_modeling.muygp import muyGP   # uncomment if using muyGP
```
thartland: @jwang125 

Having someone uncomment this if they want to use muyGP in the turbo algorithm seems undesirable.

What in `turbo.py` requires functionality from `muyGP`? 

From a Python environment with hiopbbpy and muyGP installed if I ran `TurboDriverEX.py` with a Gaussian process via muyGP instead of the smtKRG would it not run if this line was still commented out?

src/hiopbbpy/opt/turbo.py
```diff
@@ -0,0 +1,378 @@
...
+
+# Map real-space points into the unit cube defined by xlimits
+def to_unit(x, lb, ub):
+  return (x - lb) / (ub - lb)
+
+# Map unit-cube points back into real space
```
thartland: Is another way we can describe the "original"/"unnormalized"/"real" space. When I originally saw `X_real` in the `posterior_stats` `def` I thought we were dealing with the real part of a complex number.

src/Drivers/hiopbbpy/TurboDriverEX.py
```diff
@@ -0,0 +1,59 @@
...
+
+  problem = LpNormProblem(nx, xlimits)
+
+  ### initial training set (evaluated in parallel)
+  x_train = problem.sample(n_samples)
+  y_train = obj_evaluator.run(problem.evaluate, x_train)
+
+  ### surrogate FACTORY: TuRBO builds new models per region and on restart.
+  ### Swap this single line for `lambda: muyGP(nx, xlimits)` to use the scalable
+  ### GP backend; nothing else in the driver changes.
```
thartland: To run this driver with `muyGP` as the surrogate factory does one need to uncomment the `from hiopbbpy.surrogate_modeling.muygp import muyGP` line in `turbo.py`? If not what is the purpose of keeping `from hiopbbpy.surrogate_modeling.muygp import muyGP` commented out in `turbo.py`?

src/Drivers/hiopbbpy/TurboDriverEX.py
```diff
@@ -0,0 +1,59 @@
+"""
+  Code description:
+     for an example LpNormProblem
+        1) randomly sample training points and evaluate them in parallel
+        2) define a Kriging-based Gaussian Process (smt) as the surrogate
+        3) determine the minimizer via TurboAlgorithm (TuRBO / TuRBO-m)
```
thartland, line 6: Is the difference between `TuRBO` and `TuRBO-m` that one uses one global GP and the other uses one local GP per region?

src/hiopbbpy/opt/turbo.py
```diff
@@ -0,0 +1,378 @@
...
+#   log_level         logging verbosity                          default="INFO"
+
+class TurboAlgorithm(BOAlgorithmBase):
+  def __init__(self, prob:Problem, surrogate_factory, xtrain, ytrain,
+               options = None):
+    super().__init__()
+    options = options or {}
+    assert callable(surrogate_factory), "surrogate_factory must be callable() -> GaussianProcess"
+    # surrogate_factory: callable() -> GaussianProcess. A factory (not a built model)
+    # because TuRBO-m needs a fresh surrogate per region and on restart.1
```
thartland: @jwang125 Is the `1` at the end of this line is a typo?

src/hiopbbpy/surrogate_modeling/muygp.py
```diff
@@ -0,0 +1,103 @@
...
+optimize_scale) since it sets the variance magnitude Thompson sampling uses.
+Optimizing the kernel hyperparameters via MuyGPyS.optimize is the natural
+next step.
+
+"""
+import numpy as np
+from .gp import GaussianProcess
+
+# MuyGPyS >= 0.9 API
+from MuyGPyS.gp import MuyGPS
```
thartland, line 16: Should `MuyGPyS` be included as an optional dependency in the `pyproject.toml` file? More generally we should more up front (perhaps in the user manual) communicate that `MuyGPys` is an optional dependency.

nychiang: Yes, I also think we should include it as an optional dependency and set its minimum version in `setup.py`

src/hiopbbpy/opt/turbo.py
nychiang: how about making muyGP an optional dependency:            
```
try:                                                                                                 
      from hiopbbpy.surrogate_modeling.muygp import muyGP                                              
  except ImportError:                                    
      muyGP = None  # later this will fail with clear error if muyGP is used without install                               
```

src/hiopbbpy/opt/turbo.py
```diff
@@ -0,0 +1,378 @@
...
+    self.logger.setlevel(logger_level)
+
+    self.bo_maxiter = options.get('bo_maxiter', self.bo_maxiter)
+    assert self.bo_maxiter > 0, f"Invalid bo_maxiter: {self.bo_maxiter}"
+
+    batch_size = int(options.get('batch_size', 1))
+    assert batch_size > 0, f"batch_size {batch_size} is not strictly positive"
+    self.batch_size = batch_size
+
+    self.n_trust_regions = options.get('n_trust_regions', 1)
```
nychiang, line 198: keep consistent to the others
```suggestion
    self.n_trust_regions = int(options.get('n_trust_regions', 1))
```

src/hiopbbpy/opt/turbo.py
```diff
@@ -0,0 +1,378 @@
...
+from scipy.stats import qmc
+from ..surrogate_modeling.gp import GaussianProcess
+from ..problems.problem import Problem
+from ..utils.util import Evaluator
+from .boalgorithm import BOAlgorithmBase
+#from hiopbbpy.surrogate_modeling.muygp import muyGP   # uncomment if using muyGP
+
+# Map real-space points into the unit cube defined by xlimits
+def to_unit(x, lb, ub):
+  return (x - lb) / (ub - lb)
```
nychiang, line 29: Can we get a devide-by-zero error here?
should we add a test to check the bounds, e.g.,
```
if np.any(np.abs(ub - lb) < 1e-10):
      raise ValueError("Bounds are too tight")   
```

src/hiopbbpy/opt/turbo.py
```diff
@@ -0,0 +1,378 @@
...
+  mu = np.asarray(surrogate.mean(X_real)).reshape(-1)
+  var = np.asarray(surrogate.variance(X_real)).reshape(-1)
+  sd = np.sqrt(np.clip(var, 0.0, None))
+  return mu, sd
+
+# Select a batch by Thompson sampling: draw batch_size independent posterior
+# realizations f_tilde = mu + sd * z over the pooled candidates, taking the
+# argmin of each (minimization) without replacement. q independent draws give a
+# decorrelated batch, rather than the q nearest neighbors of a single draw. 
+# Important: this is diagonal (marginal) approximation due to integration with MuyGPy.
```
nychiang, line 74: Is this a limitation due to muygpy? I asked AI and it says `Standard TuRBO implementations use full GP covariance`

src/hiopbbpy/opt/turbo.py
```diff
@@ -0,0 +1,378 @@
...
+
+  # Method to (re)train the surrogate(s): one global model, or one per region
+  def _train_surrogates(self):
+    if self.local_gp:
+      self.surrogates = []
+      for r in range(len(self.regions)):
+        gp = self.surrogate_factory()
+        Xr, Yr = self._region_data(r)
+        if Xr.shape[0] < 2:                        # too sparse: fall back to all data
+          Xr, Yr = self.xtrain, self.ytrain
```
nychiang, line 239:
```suggestion
        if Xr.shape[0] < 2:                        # too sparse: fall back to all data
          self.logger.warning(f"Region {r} has <2 points, using global data for GP training") 
          Xr, Yr = self.xtrain, self.ytrain
```

src/hiopbbpy/opt/turbo.py
```diff
@@ -0,0 +1,378 @@
...
+#   obj_evaluator     Evaluator (serial) or MPIEvaluator         default from base
+#   seed              RNG seed for reproducibility               default=0
+#   log_level         logging verbosity                          default="INFO"
+
+class TurboAlgorithm(BOAlgorithmBase):
+  def __init__(self, prob:Problem, surrogate_factory, xtrain, ytrain,
+               options = None):
+    super().__init__()
+    options = options or {}
+    assert callable(surrogate_factory), "surrogate_factory must be callable() -> GaussianProcess"
```
nychiang, line 167:
```suggestion
    assert callable(surrogate_factory), "surrogate_factory must be callable() -> GaussianProcess"
    _test_gp = surrogate_factory()
    if not isinstance(_test_gp, GaussianProcess): 
      raise TypeError(f"surrogate_factory must return GaussianProcess, got {type(_test_gp)}")  
```

nychiang (commented):
LGTM and I just left some minor comments.
however, I think we should also update the README file, or write another documents, to record this new integration.

cnpetra:
Looks good on a quick pass! We should coordinate at one of the following meetings. I think we should also add the new solver in the suite of tests.

src/Drivers/hiopbbpy/TurboDriverEX.py
jwang125: Not exactly. We can have multiple surrogate models or multiple trust regions when we select points. There are two options n_trust_regions and local_gp. The restriction is that when we use local gp, right now we set the number of trust regions to be the same as the number of local GPs.

src/hiopbbpy/opt/turbo.py
jwang125: yes

src/hiopbbpy/surrogate_modeling/muygp.py
jwang125: should I add it in setup.py like how cyipopt is done?

jwang125: should the dependency declaration be added to README.md? project.toml doesn't deal with optional ones yet I think?

src/hiopbbpy/opt/turbo.py
jwang125: Should I enforce the int conversion on all options? right now no options has "int".

src/hiopbbpy/opt/turbo.py
jwang125: No it doesn't need it. Will be removed in updated version.

src/hiopbbpy/opt/turbo.py
jwang125: Will be changed.

src/Drivers/hiopbbpy/TurboDriverEX.py
jwang125: The import will only happen in the driver now.

src/hiopbbpy/opt/turbo.py
jwang125: added this to __init__ for one-time check

src/hiopbbpy/opt/turbo.py
jwang125: Yes. And this is a design choice. But I plan to add smt with full covariance in the future. This is also a scalability issue that may speed up the run.

src/hiopbbpy/opt/turbo.py
jwang125: Added.

src/hiopbbpy/opt/turbo.py
jwang125: added

src/hiopbbpy/surrogate_modeling/muygp.py
thartland: Yes, add it into `setup.py` so there is a means for a user to install `hiopbbpy` with `MuyGPys`.

thartland: I think a short mention of this optional dependency should be included in section 4.1 of the user manual.

src/hiopbbpy/opt/turbo.py
thartland: I do not see NYs suggested line `self.logger.warning(f"Region {r} has <2 points, using global data for GP training") ` as added. Did you commit the changes?

src/hiopbbpy/opt/turbo.py
thartland: I suggest that we don't actually cast `n_trust_regions` to `int`. Rather as in `boalgorithm.py` let's add an `assert` statement, as [here](https://github.com/llnl/hiop/blob/develop/src/hiopbbpy/opt/boalgorithm.py#L97), so something like

`assert isinstance(self.n_trust_regions, int), f"n_trustion_regions {self.n_trust_regions} is not an integer"`

thartland: Why is `batch_size` cast to `int` but not the others? I would do something uniform and my preference is to not cast but have `assert isinstance(..., int)` like statements.

src/hiopbbpy/surrogate_modeling/muygp.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def _nn_indices(result):
+  # NN_Wrapper.get_nns / get_batch_nns return either an index array or an
+  # (indices, distances) tuple depending on version. this function normalizes both to just the indices.
+  return result[0] if isinstance(result, tuple) else result
+
+
+class muyGP(GaussianProcess):
+  def __init__(self, ndim, xlimits, nn_count=None, smoothness=1.5,
```
thartland, line 33: `smoothness` is a little ambiguous, `matern_smoothness_parameter` would be a more descriptive.

src/hiopbbpy/surrogate_modeling/muygp.py
```diff
@@ -0,0 +1,103 @@
+"""
+A GaussianProcess surrogate backed by MuyGPyS: predicts each point from its
+nearest neighbors, avoiding the dense n x n solve. Same train/mean/variance
```
thartland: Somewhere in the file we should indicate that `nn` stands for nearest neighbors. I suggest here as `nearest neighbors (nn), avoiding the...` and below with a comment such as `self.nn_method = nn_method # nearest neighbor method`

src/hiopbbpy/surrogate_modeling/muygp.py
jwang125: I added description to the Sections/hiopbbpy.tex . Should we update the pdf file as well?

src/hiopbbpy/opt/turbo.py
jwang125: ok. modified according to suggestions.

src/hiopbbpy/surrogate_modeling/muygp.py
jwang125: MuyGPyS calls this parameter smoothness. I can make it more clear in the description if you are ok with it.

src/hiopbbpy/surrogate_modeling/muygp.py
jwang125: corrected.

src/hiopbbpy/surrogate_modeling/muygp.py
thartland: Yes, please update the pdf. I believe we want the tex and pdf's to be synced.

src/hiopbbpy/surrogate_modeling/muygp.py
thartland: The parameter name `smoothness` seems fine to be consistent with `MuyGPys` but some comment about what it corresponds to would be great. Thanks!

src/hiopbbpy/opt/turbo.py
thartland:
```suggestion
    # because TuRBO-m needs a fresh surrogate per region and on restart.
```

jwang125:
Aside from adding the new solver in the suite of tests, the comments are all addressed. Please take another look. Let me know if we want the tests to be set up in this PR.

src/hiopbbpy/opt/turbo.py
nychiang: @jwang125 Did you commit the changes?

jwang125: Yes I just did today.

thartland (changes requested):
Overall this looks good to me, the only outstanding item is to update the user manual pdf.

jwang125:
User manual has been updated.

PR #769, opened 2026-08-14
Branch bound async

cnpetra:
Adds asynchronous task-parallel branch-and-bound search.

Tested extensively with convex relaxations `opt-mode` 4 and works well. Significant speed-ups due to parallel execution and asynchronous processing as the search gets deep in the tree. BnB restart code is not tested, will be revisited after merge.

The PR also contains some fixes related to the use of parallel evaluation manager.


thartland (approved):
Looks good. Did some runs and looks great to me. Will merge this in and update based on some of my recent changes that are not committed.

